PR #3, opened 2020-08-25
4515 - Corrections FrameWork

cameron-freshworks:
*Issue #:* /bcgov/entity#4515

*Description of changes:*

* Implements an updated application flow for our Corrections and Alterations filings components.

* A view for Correction and Alteration

* Updated to use Incorporation Number opposed to logic around NR #'s for app initialization.

* Updated some error dialogs.

* Updated the legal api and filing template mixin.

* Implemented some components in summary view to present the data parsed from the filings, only a temporary representation.

*Note: *

* Have not updated the unit tests. This will need a significant rework and will most likely be started again from ground up. Using the old tests as reference.

* A lot of the old / existing components still remain in the repo. Many of which will provide a good foundation for future adaptations. 


src/components/common/Actions.vue
```diff
@@ -31,20 +31,19 @@
...
+<!--      <v-fade-transition hide-on-leave>-->
+<!--        <v-btn id="review-confirm-btn" large color="primary"-->
+<!--          :to="nextRoute"-->
+<!--          v-show="isShowReviewConfirmBtn"-->
+<!--          :disabled="isBusySaving"-->
+<!--        >-->
+<!--          <span>{{ nextButtonLabel }}</span>-->
+<!--          <v-icon>mdi-chevron-right</v-icon>-->
+<!--        </v-btn>-->
+<!--      </v-fade-transition>-->
```
cameron-freshworks, line 43: Wasn't sure of the entire flow if we will need anything beyond a File And Pay. Commented out for now, can remove later.

src/components/common/EntityInfo.vue
```diff
@@ -40,29 +54,32 @@
...
   @Getter getTempId!: GetterIF
   @Getter getApprovedName!: GetterIF
 
   /** The entity title  */
   private entityTitle (): string {
     if (this.isTypeBcomp) {
-      return 'BC Benefit Company Incorporation Application'
-    } else if (this.isTypeCoop) {
-      return 'Incorporate a BC Cooperative Association'
+      return 'Correction - Incorporation Application'
```
cameron-freshworks, line 81: This could use an update once we see the designs for alterations / profile view etc. Then can populate the entity info accordingly. This goes for the Breadcrumb as well.

src/mixins/legal-api-mixin.ts
```diff
@@ -75,6 +75,29 @@ export default class LegalApiMixin extends Vue {
       })
   }
 
+  /**
+   * Fetches a filing.
+   * @returns a promise to return the filing of the specified type, or null if not found
+   */
+  async fetchFiling (businessId: string, filingType: string): Promise<IncorporationFilingIF> {
```
cameron-freshworks, line 82: This function will most likely need to pass an ID or something in the future as well, if in the event that we are pulling on multiple filings of the same type. Ie.. pulling an Address Change for correction, will require more logic.

src/router/routes.ts
```diff
@@ -22,59 +23,27 @@ export const routes = [
...
+      requiresAuth: true,
+      isStaffOnly: false
     }
   },
   {
     // default/fallback route
     // must be last
     path: '*',
-    redirect: '/define-company'
+    redirect: '/correction'
```
cameron-freshworks, line 47: Not sure what the fallback route should be, so Corrections for now.

src/router/router.ts
```diff
@@ -51,6 +56,16 @@ export function getVueRouter () {
     return Boolean(sessionStorage.getItem(SessionStorageKeys.KeyCloakToken))
   }
 
+  /** Returns True if route requires staff role, else False. */
+  function isStaffOnly (route: Route): boolean {
+    return route.matched.some(r => r.meta?.isStaffOnly)
+  }
+
+  /** Returns True if user is staff, else False. */
+  function isStaff (): boolean {
```
severinbeauvais, line 65: If this is the only place this is used then I'm OK with it here. Otherwise move it to getter module.

src/router/routes.ts
severinbeauvais: Sounds reasonable.
What happens if there is no default/fallback route and you go an undefined route?

src/enums/routeNames.ts
```diff
@@ -11,5 +11,7 @@ export enum RouteNames {
     CREATE_SHARE_STRUCTURE = 'create-share-structure',
     DEFINE_COMPANY = 'define-company',
     INCORPORATION_AGREEMENT = 'incorporation-agreement',
```
severinbeauvais, line 13: Don't need these other ones?

src/mixins/legal-api-mixin.ts
```diff
@@ -75,6 +75,29 @@ export default class LegalApiMixin extends Vue {
...
+    const url = `businesses/${businessId}/filings`
+    return axios.get(url)
+      .then(response => {
+        const filings = response?.data?.filings
+        const returnfiling = filings.find(filings => filings.filing.header.name === filingType)
```
severinbeauvais, line 87: `filings` could be null/undefined -- use ?

src/mixins/legal-api-mixin.ts
```diff
@@ -75,6 +75,29 @@ export default class LegalApiMixin extends Vue {
...
+
+        if (!filings || !returnfiling) {
+          throw new Error('Invalid API response')
+        }
+        return returnfiling
+      })
+      .catch((error) => {
+        if (error?.response?.status === NOT_FOUND) {
+          throw error // IA not found
```
severinbeauvais, line 96: Obsolete comment.

src/store/getters/state-getters.ts
```diff
@@ -104,7 +104,7 @@ export const getNameRequestNumber = (state: any): string => {
  * Returns the approved name of a name request.
  */
 export const getApprovedName = (state: any): string => {
-  return state.stateModel.nameRequest.details.approvedName
+  return state.stateModel.nameRequest.legalName
```
severinbeauvais, line 107: I think, at this point, the NR is no longer useful. You should be able to get the Legal Name from the business data instead.
(I'm unsure what the code will look like if we associated a new NR!)

src/utils/auth-helper.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Fetches config from environment and API.
+ * @returns A promise to get & set session storage keys with appropriate values.
+ */
```
severinbeauvais, line 4: Obsolete comment.

src/utils/auth-helper.ts
```diff
@@ -0,0 +1,37 @@
...
+/** Gets Keycloak JWT and parses it. */
```
severinbeauvais, line 5: This comment goes with line 6.

src/utils/auth-helper.ts
```diff
@@ -0,0 +1,37 @@
...
+import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
+
+export function getJWT (): any {
```
severinbeauvais, line 8: This function doesn't need to be exported.

src/router/routes.ts
cameron-freshworks: Good question, i'll find out!

src/router/router.ts
cameron-freshworks: As of right now, this is the only place we are verifying the role directly off of the token. 
Why i did it this way, when we mount the app we have the keycloak token in session but we don't have the store(or values) set when attempting the route. This allows us to verify the user is staff before the store is mounted using the `auth-helper.ts`.

src/enums/routeNames.ts
cameron-freshworks: At this time, i don't believe so. From what i see in the wireframes, we are a single view for an entire 'Correction' or 'Alteration' and so on. So i believe it will be up to us what components we populate in each respective views, dependant on the type of filing we are correcting, altering etc.

cameron-freshworks: Will update this!

src/mixins/legal-api-mixin.ts
cameron-freshworks: ahh, yes will update this!

src/mixins/legal-api-mixin.ts
cameron-freshworks: Your right, this function is supposed to be generic (eventually 😅 ). Thanks! Updated in next commit.

src/store/getters/state-getters.ts
cameron-freshworks: Ahh right, thanks. Will look at this!

src/router/routes.ts
severinbeauvais: I'm guessing you'll get a 404... this is OK with me.
I'm thinking that defaulting to a "profile" (read-only) page would be a safe fallback as well.
Maybe this can be sorted out in the (near) future, not now.

severinbeauvais:
For now, maybe disable the broken unit tests?
IMHO, this ticket's scope is too large: Create the framework, yes. Implement a bunch of functionality as well, no -> subsequent ticket.
So I'm happy with what you've done so far... Let's finish this ticket and see what needs doing next.

src/router/routes.ts
cameron-freshworks: In this case, we mount the app but get stuck on our infinite loading spinner because we can't reach our views. 
The refactor puts the data loading for the respective filings into our views which in turn let the app know its ready or not ready. So without that, infinite loader

cameron-freshworks:
> For now, maybe disable the broken unit tests?
> IMHO, this ticket's scope is too large: Create the framework, yes. Implement a bunch of functionality as well, no -> subsequent ticket.
> So I'm happy with what you've done so far... Let's finish this ticket and see what needs doing next.

Awesome, thanks! Sounds like a good plan 👍 Just wrapping up some clean up based on the feedback!

src/router/routes.ts
severinbeauvais: Makes sense.
I can't decide what's best (or worst) for now, so do whatever and we can rethink this when we know more :)

src/utils/auth-helper.ts
cameron-freshworks: Updated in next commit 👍

src/utils/auth-helper.ts
cameron-freshworks: oops, sloppy copy pasta here! Fixed in next commit

src/utils/auth-helper.ts
cameron-freshworks: Fixed!

cameron-freshworks:
After 1 review and approval, will merge due to very low impact and risk at this point.

PR #11, opened 2020-09-03
Changes as per the new Share strucuture schema

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4553

*Description of changes:*
Changes in share structure

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -131,7 +131,8 @@ export default class FilingTemplateMixin extends Vue {
     this.setOrgPersonList(filing.incorporationApplication.parties)
 
     // Set Share Structure
-    this.setShareClasses(filing.incorporationApplication.shareStructure.shareClasses)
+    this.setShareClasses(filing.incorporationApplication.shareStructure
+      ? filing.incorporationApplication.shareStructure.shareClasses : [])
 
     // Set Incorporation Agreement
     this.setIncorporationAgreementStepData({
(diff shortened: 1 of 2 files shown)


src/mixins/filing-template-mixin.ts
```diff
@@ -131,7 +131,7 @@ export default class FilingTemplateMixin extends Vue {
     this.setOrgPersonList(filing.incorporationApplication.parties)
 
     // Set Share Structure
-    this.setShareClasses(filing.incorporationApplication.shareClasses)
+    this.setShareClasses(filing.incorporationApplication.shareStructure.shareClasses)
```
severinbeauvais, line 134: Should this check whether shareStructure is truthy first, as in
https://github.com/bcgov/bcrs-business-create-ui/blob/a1879ad75aace3a8b4c793aba79516f1c1a94774/src/mixins/filing-template-mixin.ts#L129
?

severinbeauvais (commented):
Thanks for updating our new project 👍

src/mixins/filing-template-mixin.ts
lekshmimallika-aot: Share structure will be truthy in our completed incorporation application. The reason why we have the check in other place is for the temp reg when we start the filing we will only have the nameRequest within it.. share structure wont be present will we save it / file it. It is not the case when it comes to edit UI.

However, there are some applications in prod that are completed and will have the old share structure. But we have an action item to correct the json of those completed filings in prod and should be done before this UI gets deployed to prod

cameron-freshworks (approved):
Excellent, thanks for this!

src/mixins/filing-template-mixin.ts
severinbeauvais: I think we should check anyway, please and thanks!

severinbeauvais (approved):
Thanks for the safety check!
Looks good.

PR #19, opened 2020-09-11
Correct Company Name Component and UI Design Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#4843

*Description of changes:*

* Implements sub component to update Company name.

* Refactored store to save IA Snapshot for comparisons and to preserve some UI elements.

* Implemented 'Undo` functionality and UI indicators where correction was made. 

* Updates UI for all Name Correction components as per invision designs. 
* https://projects.invisionapp.com/d/main/#/console/20440642/429271383/preview

### DIFF
--- a/src/components/Company/CompanyName/CorrectCompanyName.vue
+++ b/src/components/Company/CompanyName/CorrectCompanyName.vue
@@ -1,0 +1,96 @@
+<template>
+  <v-form ref="correctNameForm" v-model="valid" lazy-validation>
+    <v-layout row>
+      <v-flex>
+        <v-text-field
+          v-model="companyName"
+          filled
+          req
+          persistent-hint
+          :rules="companyNameRules"
+          data-test="business-identifier"
+        ></v-text-field>
+      </v-flex>
+    </v-layout>
+  </v-form>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Prop, Watch, Emit, Vue } from 'vue-property-decorator'
+import { Action, Getter, State } from 'vuex-class'
+
+// Interfaces && enums
+import { ActionBindingIF, BusinessInformationIF, NameRequestIF } from '@/interfaces'
+import { CorrectionTypes } from '@/enums'
+
+@Component({})
+export default class CorrectNameRequest extends Vue {
+  /** Form Submission Prop */
+  @Prop({ default: false }) submitId: string
+
+  // Global state
+  @State(state => state.stateModel.nameRequest)
+  readonly nameRequest!: NameRequestIF
+
+  @Action setNameRequest!: ActionBindingIF
+
+  @Getter getApprovedName!: string
+
+  private valid = false
+  private companyName = ''
+
+  // Form Ref
+  $refs: { correctNameForm: HTMLFormElement }
+
+  // Rules
+  private companyNameRules = [
+    v => !!v || ' A company name is required'
+  ]
+
+  mounted (): void {
+    // Set the current company name to the form
+    if (this.getApprovedName) {
+      this.companyName = this.getApprovedName
+    }
+  }
+
+  // Validations
+  private get isFormValid (): boolean {
+    return this.valid
+  }
+
+  private resetForm () {
+    this.$refs.correctNameForm.resetValidation()
+  }
+
+  /** Watch for form submission and emit results. */
+  @Watch('submitId')
+  private async onSubmit (): Promise<any> {
+    if (this.submitId === CorrectionTypes.CORRECT_NAME) {
+      const correctedCompanyName = { legalName: this.companyName }
+      this.setNameRequest({ ...this.nameRequest, ...correctedCompanyName })
+      this.emitDone(CorrectionTypes.CORRECT_NEW_NR)
+    }
+  }
+
+  /** Inform parent the process is complete. */
+  @Emit('done')
+  private emitDone (type: CorrectionTypes = null): void {
+    if (!type) this.resetForm()
+  }
+
+  /** Inform parent when form is valid and ready for submission. */
+  @Watch('valid')
+  @Emit('isValid')
+  private emitValid (): boolean {
+    return this.isFormValid
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+  .v-expansion-panel-content ::v-deep .v-expansion-panel-content__wrap {
+    padding: 0;
+  }
+</style>
--- a/src/components/Company/CompanyName/CorrectCompanyName.vue
+++ b/src/components/Company/CompanyName/CorrectCompanyName.vue
@@ -28,13 +28,10 @@ export default class CorrectCompanyName extends Vue {
   /** Form Submission Prop */
   @Prop({ default: '' }) formType: string
 
-  // Global state
-  @State(state => state.stateModel.nameRequest)
-  readonly nameRequest!: NameRequestIF
-
   @Action setNameRequest!: ActionBindingIF
 
   @Getter getApprovedName!: string
+  @Getter getNameRequest!: NameRequestIF
 
   private valid = false
   private companyName = ''
@@ -64,7 +61,7 @@ export default class CorrectCompanyName extends Vue {
   private async onSubmit (): Promise<any> {
     if (this.formType === CorrectionTypes.CORRECT_NAME) {
       const correctedCompanyName = { legalName: this.companyName }
-      this.setNameRequest({ ...this.nameRequest, ...correctedCompanyName })
+      this.setNameRequest({ ...this.getNameRequest, ...correctedCompanyName })
       this.emitDone(CorrectionTypes.CORRECT_NEW_NR)
     }
   }
--- a/src/components/Company/CompanyName/CorrectCompanyName.vue
+++ b/src/components/Company/CompanyName/CorrectCompanyName.vue
@@ -26,7 +26,7 @@ import { CorrectionTypes } from '@/enums'
 @Component({})
 export default class CorrectCompanyName extends Vue {
   /** Form Submission Prop */
-  @Prop({ default: '' }) formType: string
+  @Prop({ default: '' }) formType: CorrectionTypes
 
   @Action setNameRequest!: ActionBindingIF
 
--- a/src/components/Company/CompanyName/CorrectCompanyName.vue
+++ b/src/components/Company/CompanyName/CorrectCompanyName.vue
@@ -5,7 +5,6 @@
         <v-text-field
           v-model="companyName"
           filled
-          req
           persistent-hint
           :rules="companyNameRules"
           data-test="business-identifier"
@@ -25,9 +24,9 @@ import { ActionBindingIF, BusinessInformationIF, NameRequestIF } from '@/interfa
 import { CorrectionTypes } from '@/enums'
 
 @Component({})
-export default class CorrectNameRequest extends Vue {
+export default class CorrectCompanyName extends Vue {
   /** Form Submission Prop */
-  @Prop({ default: false }) submitId: string
+  @Prop({ default: '' }) formType: string
 
   // Global state
   @State(state => state.stateModel.nameRequest)
@@ -60,14 +59,10 @@ export default class CorrectNameRequest extends Vue {
     return this.valid
   }
 
-  private resetForm () {
-    this.$refs.correctNameForm.resetValidation()
-  }
-
   /** Watch for form submission and emit results. */
-  @Watch('submitId')
+  @Watch('formType')
   private async onSubmit (): Promise<any> {
-    if (this.submitId === CorrectionTypes.CORRECT_NAME) {
+    if (this.formType === CorrectionTypes.CORRECT_NAME) {
       const correctedCompanyName = { legalName: this.companyName }
       this.setNameRequest({ ...this.nameRequest, ...correctedCompanyName })
       this.emitDone(CorrectionTypes.CORRECT_NEW_NR)
@@ -76,9 +71,7 @@ export default class CorrectNameRequest extends Vue {
 
   /** Inform parent the process is complete. */
   @Emit('done')
-  private emitDone (type: CorrectionTypes = null): void {
-    if (!type) this.resetForm()
-  }
+  private emitDone (type: CorrectionTypes): void {}
 
   /** Inform parent when form is valid and ready for submission. */
   @Watch('valid')
--- a/src/components/Company/CompanyName/CorrectCompanyName.vue
+++ b/src/components/Company/CompanyName/CorrectCompanyName.vue
@@ -17,16 +17,16 @@
 <script lang="ts">
 // Libraries
 import { Component, Prop, Watch, Emit, Vue } from 'vue-property-decorator'
-import { Action, Getter, State } from 'vuex-class'
+import { Action, Getter } from 'vuex-class'
 
 // Interfaces && enums
-import { ActionBindingIF, BusinessInformationIF, NameRequestIF } from '@/interfaces'
+import { ActionBindingIF, NameRequestIF } from '@/interfaces'
 import { CorrectionTypes } from '@/enums'
 
 @Component({})
 export default class CorrectCompanyName extends Vue {
   /** Form Submission Prop */
-  @Prop({ default: '' }) formType: CorrectionTypes
+  @Prop({ default: null }) formType: CorrectionTypes
 
   @Action setNameRequest!: ActionBindingIF
 
--- a/src/components/Company/CompanyName/CorrectNameOptions.vue
+++ b/src/components/Company/CompanyName/CorrectNameOptions.vue
@@ -1,18 +1,28 @@
 <template>
   <v-container id="name-options-container">
-    <h4 class="name-options-info mb-3">You can correct the company name in one of the following ways:</h4>
+    <p class="name-options-info mb-5" v-if="!isOneOption">
+      You can correct the company name in one of the following ways:
+    </p>
     <v-expansion-panels v-model="panel">
       <v-expansion-panel
         v-for="(item,i) in displayedOptions"
         :key="i"
+        :disabled="isOneOption"
+        @click="identifyForm(item.id)"
       >
-        <v-expansion-panel-header class="name-options-header">{{item.description}}</v-expansion-panel-header>
+        <v-expansion-panel-header class="px-0" :class="{'name-options-header': isOneOption}">
+          <span class="names-option-title">{{item.title}}</span>
+          <template v-slot:actions>
+            <v-icon color="primary">mdi-menu-down</v-icon>
+          </template>
+        </v-expansion-panel-header>
         <v-expansion-panel-content class="name-options-content">
+          <v-label color="primary">{{item.description}}</v-label>
           <v-container>
             <component
               :is="item.component"
               :key="item.id"
-              :submit="submit"
+              :submitId="submitId"
               @done="emitDone($event)"
               @isValid="isFormValid = $event"
             />
@@ -24,7 +34,8 @@
     <div class="action-btns my-3">
       <v-btn
         id="done-btn"
-        large color="primary"
+        large
+        color="primary"
         @click="submitNameCorrection"
         :disabled="!isFormValid"
         :loading="isLoading"
@@ -32,7 +43,7 @@
         <span>Done</span>
       </v-btn>
 
-      <v-btn id="cancel-btn" large @click="emitCancel"><span>Cancel</span></v-btn>
+      <v-btn id="cancel-btn" large outlined color="primary" @click="emitCancel"><span>Cancel</span></v-btn>
     </div>
   </v-container>
 </template>
@@ -42,7 +53,7 @@
 import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
 
 // Components
-import CorrectNameRequest from './CorrectNameRequest.vue'
+import { CorrectNameRequest, CorrectCompanyName } from './'
 
 // Interfaces & Enums
 import { CorrectNameOptionIF } from '@/interfaces'
@@ -57,7 +68,8 @@ import { CorrectionTypes } from '@/enums'
  */
 @Component({
   components: {
-    CorrectNameRequest
+    CorrectNameRequest,
+    CorrectCompanyName
   }
 })
 export default class CorrectNameOptions extends Vue {
@@ -67,23 +79,28 @@ export default class CorrectNameOptions extends Vue {
   // local properties
   private displayedOptions: Array<CorrectNameOptionIF> = []
   private panel = null as number
-  private submit = false
+  private submitId = ''
+  private currentPanelId = ''
   private isLoading = false
   private isFormValid = false
   private correctionNameOptions: Array<CorrectNameOptionIF> = [
     {
       id: CorrectionTypes.CORRECT_NAME,
-      description: 'Edit the company name',
-      component: ''// CorrectName
+      title: 'Edit the company name',
+      description: 'Correct typographical errors int he existing company name.',
+      component: CorrectCompanyName
     },
     {
       id: CorrectionTypes.CORRECT_NAME_TO_NUMBER,
-      description: 'Use the incorporation number as the name',
+      title: 'Use the incorporation number as the name',
+      description: '',
       component: '' // CorrectNameToNumber
     },
     {
       id: CorrectionTypes.CORRECT_NEW_NR,
-      description: 'Use a new name request number',
+      title: 'Use a new name request number',
+      description: 'Enter the new Name Request Number (e.g., NR1234567) and either the applicant phone number OR the ' +
+        'applicant email that was used when the name was requested.',
       component: CorrectNameRequest
     }
   ]
@@ -93,20 +110,35 @@ export default class CorrectNameOptions extends Vue {
     this.displayedOptions = this.correctionNameOptions.filter(
       option => this.correctionNameChoices.includes(option.id)
     )
-    if (this.correctionNameChoices.length === 1) this.panel = 0 // open by default if only 1 option
+    // open by default and assign id if only 1 option
+    if (this.isOneOption) {
+      this.panel = 0
+      this.currentPanelId = this.displayedOptions[0].id
+    }
+  }
+
+  private get isOneOption (): boolean {
+    return this.correctionNameChoices.length === 1
   }
 
   /** Trigger form submission */
   private submitNameCorrection (): void {
     this.isLoading = true
-    this.submit = !this.submit
+    this.submitId = this.currentPanelId
+  }
+
+  /** Identify the current form */
+  private identifyForm (id: string) {
+    this.currentPanelId = id
+    this.isFormValid = false
   }
 
   /** Inform Parent name correction process is done. */
   @Emit('done')
-  private emitDone (isSaved: boolean): void {
+  private emitDone (type: CorrectionTypes): void {
     this.isLoading = false
-    if (isSaved) this.panel = null
+    this.submitId = ''
+    if (type) this.panel = null
   }
 
   /** cancel name correction */
@@ -121,6 +153,14 @@ export default class CorrectNameOptions extends Vue {
   #name-options-container {
     padding: 0;
 
+    .name-options-header {
+      align-items: start;
+    }
+
+    .names-option-title {
+      color: #1669BB
+    }
+
     .action-btns {
       display: flex;
       justify-content: flex-end;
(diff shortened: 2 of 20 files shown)


src/components/DefineCompany/YourCompany.vue
```diff
@@ -203,10 +216,36 @@ export default class YourCompany extends Mixins(DateMixin, EntityFilterMixin, Le
       : 'Unknown'
   }
 
-  /** Handle the company name change data */
-  private async nameChangeHandler (isSaved: boolean): Promise<any> {
-    // Handle name corrections vs a snapshot of the original IA here for UI indicators.
-    if (isSaved) this.companyNameChanges = false
+  /** Compare current to corrected data and update UI.  */
+  private async nameChangeHandler (type: CorrectionTypes): Promise<any> {
```
cameron-freshworks, line 220: I added a switch case here to handle the different comparisons based off what type of name correction occurred. Without a Named Company to test with, i was 100% sure if the comparison was going to be the same, so currently the switch cases are using the same comparison method, but this could change. So this may seem redundant now, but could be utilized when we introduce the third scenario and/or have some named companies to test with. If unneeded, will refactor in the near future.

cameron-freshworks:
For testing out the components, i have an NR that isn't expired to verify the updated data in the UI. 
NR 0479445 / lorna.trent@gov.bc.ca

src/components/Company/CompanyName/CorrectCompanyName.vue
```diff
@@ -0,0 +1,96 @@
...
+import { Action, Getter, State } from 'vuex-class'
+
+// Interfaces && enums
+import { ActionBindingIF, BusinessInformationIF, NameRequestIF } from '@/interfaces'
+import { CorrectionTypes } from '@/enums'
+
+@Component({})
+export default class CorrectNameRequest extends Vue {
+  /** Form Submission Prop */
+  @Prop({ default: false }) submitId: string
```
severinbeauvais, line 30: Type mismatch.

src/components/Company/CompanyName/CorrectCompanyName.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+  <v-form ref="correctNameForm" v-model="valid" lazy-validation>
+    <v-layout row>
+      <v-flex>
+        <v-text-field
+          v-model="companyName"
+          filled
+          req
```
severinbeauvais, line 8: Is this redundant, since you aleady have a rule to make this required?

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
```diff
@@ -17,16 +19,16 @@ export interface NameRequestDetailsIF {
...
-  emailAddress: string
-  phoneNumber: string
-  addressLine1: string
-  addressLine2: string
-  addressLine3: string
-  city: string
-  countryTypeCode: string
-  postalCode: string
-  stateProvinceCode: string
+  firstName?: string
```
severinbeauvais, line 22: It feels like this IF doesn't have much point if everything is optional. Would it make more sense to have the IF itself optional? (I'm not sure.)

src/store/index.ts
```diff
@@ -112,7 +114,8 @@ export function getVuexStore () {
       mutateOfficeAddresses,
       mutateOrgPersonList,
       mutateShareClasses,
-      mutateUserEmail
+      mutateUserEmail,
+      mutateOriginalIA
```
severinbeauvais, line 118: I've been trying to keep this alphabetized :)

src/store/state/state-model.ts
```diff
@@ -27,6 +27,8 @@ export const stateModel: StateModelIF = {
     identifier: ''
   },
   nameRequest: {
+    legalType: '',
```
severinbeauvais, line 30: Do you have a type for "none"? (or perhaps null?)

src/components/Company/CompanyName/CorrectCompanyName.vue
```diff
@@ -0,0 +1,96 @@
...
+      const correctedCompanyName = { legalName: this.companyName }
+      this.setNameRequest({ ...this.nameRequest, ...correctedCompanyName })
+      this.emitDone(CorrectionTypes.CORRECT_NEW_NR)
+    }
+  }
+
+  /** Inform parent the process is complete. */
+  @Emit('done')
+  private emitDone (type: CorrectionTypes = null): void {
+    if (!type) this.resetForm()
```
severinbeauvais, line 80: Is type ever falsey here?

src/components/Company/CompanyName/CorrectCompanyName.vue
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private resetForm () {
+    this.$refs.correctNameForm.resetValidation()
+  }
+
+  /** Watch for form submission and emit results. */
+  @Watch('submitId')
+  private async onSubmit (): Promise<any> {
+    if (this.submitId === CorrectionTypes.CORRECT_NAME) {
```
severinbeauvais, line 70: Why is the property named "id" when it's a type?

src/components/Company/CompanyName/CorrectCompanyName.vue
```diff
@@ -0,0 +1,96 @@
...
+// Libraries
+import { Component, Prop, Watch, Emit, Vue } from 'vue-property-decorator'
+import { Action, Getter, State } from 'vuex-class'
+
+// Interfaces && enums
+import { ActionBindingIF, BusinessInformationIF, NameRequestIF } from '@/interfaces'
+import { CorrectionTypes } from '@/enums'
+
+@Component({})
+export default class CorrectNameRequest extends Vue {
```
lekshmimallika-aot, line 28: Should the class name be CorrectCompanyName?

src/components/Company/CompanyName/CorrectCompanyName.vue
cameron-freshworks: Fixed in next commit!

src/components/Company/CompanyName/CorrectCompanyName.vue
cameron-freshworks: It is! Will fix in next commit

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: o i didn't know we could do that, i think that WOULD make more sense.

src/store/index.ts
cameron-freshworks: oops, will fix in next commit!

src/store/state/state-model.ts
cameron-freshworks: i think null makes sense here, will update!

src/components/Company/CompanyName/CorrectCompanyName.vue
cameron-freshworks: poor naming, will update this!

src/components/Company/CompanyName/CorrectCompanyName.vue
cameron-freshworks: Absolutely, thank you 😅

src/components/Company/CompanyName/CorrectCompanyName.vue
cameron-freshworks: Not in this component. In the other it made sense to handle Name Request Failures but here it is redundant. Will update this, good catch!

src/components/Company/CompanyName/CorrectCompanyName.vue
cameron-freshworks: Refactored `submitId` to `formType` in all places in next commit.

src/components/DefineCompany/YourCompany.vue
```diff
@@ -8,52 +8,65 @@
     <div class="section-container px-8">
       <!--TODO: Replace container content with Name Request Summary when it is ready -->
       <v-layout row>
-        <v-flex md4>
-          <label><strong>Company Name</strong></label>
+        <v-flex md3>
+          <v-layout column>
```
severinbeauvais, line 12: This may not be the time to figure out the layout but I'm bringing it up anyway to think about.

First, is this code really needed for a single column?

Second, I think all rows in this section should be in a single v-layout block (so that all columns are the same width).

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: Updated in next commit

src/components/DefineCompany/YourCompany.vue
```diff
@@ -183,11 +199,8 @@ export default class YourCompany extends Mixins(DateMixin, EntityFilterMixin, Le
   private folioNumberChanges = false
   private nameTranslationChanges = false
   private officeAddressChanges = false
-  private correctNameChoices = [
-    CorrectionTypes.CORRECT_NEW_NR,
-    CorrectionTypes.CORRECT_NAME_TO_NUMBER,
-    CorrectionTypes.CORRECT_NAME
-  ]
+  private correctNameChoices = []
```
severinbeauvais, line 202: Array type?

src/components/Company/CompanyName/CorrectNameOptions.vue
```diff
@@ -67,23 +80,28 @@ export default class CorrectNameOptions extends Vue {
...
     {
       id: CorrectionTypes.CORRECT_NAME,
-      description: 'Edit the company name',
-      component: ''// CorrectName
+      title: 'Edit the company name',
+      description: 'Correct typographical errors int he existing company name.',
+      component: CorrectCompanyName
     },
     {
       id: CorrectionTypes.CORRECT_NAME_TO_NUMBER,
```
lekshmimallika-aot, line 95: How is this handled? I see that the other two components name request correction and new NR alters the store state in those components itself. Shouldnt this option also alter the name request object in the store or is it handled some other way?

lekshmimallika-aot (approved):
Looks great to me overall

src/store/state/state-model.ts
cameron-freshworks: Actually, i think it is ok to initialize as an empty string here isn't it? The base type is a string

src/components/DefineCompany/YourCompany.vue
```diff
@@ -162,11 +171,18 @@ export default class YourCompany extends Mixins(DateMixin, EntityFilterMixin, Le
...
 
   // Global state
   @State(state => state.stateModel.defineCompanyStep.valid)
   readonly valid!: boolean
 
+  @State(state => state.stateModel.originalIA)
+  readonly originalIA!: IncorporationFilingIF
+
+  @State(state => state.stateModel)
+  readonly stateModel!: StateModelIF
```
severinbeauvais, line 184: Unless we need a LOT of properties from an object, I think we should use getters instead of looking at the whole state, which is really looking into the store data structure -- if possible let's consider the store a black box and just use setters/getters along with interface types.
(This can be fixed later so it doesn't hold up this commit.)

src/components/Company/CompanyName/CorrectNameRequest.vue
```diff
@@ -1,57 +1,90 @@
...
 
 @Component({})
 export default class CorrectNameRequest extends Mixins(NameRequestMixin) {
   /** Form Submission Prop */
-  @Prop({ default: false }) submit: boolean
+  @Prop({ default: false }) submitId: string
+
+  // Global state
+  @State(state => state.stateModel.nameRequest)
+  readonly nameRequest!: NameRequestIF
```
severinbeauvais, line 78: We should just have a getter for this data.
Same for "businessInformation" below.

src/components/DefineCompany/YourCompany.vue
cameron-freshworks: Updated!

src/components/Company/CompanyName/CorrectNameRequest.vue
```diff
@@ -100,32 +141,35 @@ export default class CorrectNameRequest extends Mixins(NameRequestMixin) {
...
+            legalName: this.getNrApprovedName(response),
+            applicant: {
+              phoneNumber: this.entityPhone,
+              emailAddress: this.entityEmail
+            }
+          }
+          this.setNameRequest({ ...this.nameRequest, ...nrCorrection })
+          this.emitDone(CorrectionTypes.CORRECT_NEW_NR)
+        }).catch(() => {
+          this.emitDone()
```
severinbeauvais, line 164: No error feedback to user?
No console log for debugging?

src/components/Company/CompanyName/CorrectNameOptions.vue
cameron-freshworks: This is the last Name Correction option that i'm currently working on in ticket #4587. It will do something very similar!

src/components/DefineCompany/YourCompany.vue
cameron-freshworks: no no i think that makes a lot of sense. In this case we're looking at something pretty specific, or at least one section of the state. Will update this to 'get' the specific value for comparison.

src/components/Company/CompanyName/CorrectNameRequest.vue
cameron-freshworks: kk, copy that, will update this!

src/components/Company/CompanyName/CorrectNameRequest.vue
cameron-freshworks: the error is handled during the request. If it fails during the request the dialog will appear with the type of error. This is simply to inform the parent that process ( whether success or fail ) is now complete.

src/components/DefineCompany/YourCompany.vue
cameron-freshworks: This specific column is for keeping the label and v-chip in alignment.

I definitely agree though, this layout could certainly use another look, especially as we continue on with adding more correction options to this `yourCompany` component.

src/components/Company/CompanyName/CorrectNameOptions.vue
```diff
@@ -67,23 +80,28 @@ export default class CorrectNameOptions extends Vue {
   // local properties
   private displayedOptions: Array<CorrectNameOptionIF> = []
   private panel = null as number
-  private submit = false
+  private submitId = ''
+  private currentPanelId = ''
```
severinbeauvais, line 84: Are these strings or numbers?
(PS You can initialize numbers to NaN if needed.)

cameron-freshworks: This has been refactored in next commit. The `id` was misleading, it is a type.

src/components/Company/CompanyName/CorrectCompanyName.vue
severinbeauvais: Should be
```
formType: CorrectionTypes
```

src/store/state/state-model.ts
severinbeauvais: This is fine.
I can't remember what I was thinking now.
Carry on.

src/components/Company/CompanyName/CorrectCompanyName.vue
cameron-freshworks: gah you are right. Will amend this. 🤦

src/components/Company/CompanyName/CorrectCompanyName.vue
severinbeauvais: I see this is now fixed.

src/components/Company/CompanyName/CorrectCompanyName.vue
cameron-freshworks: Fixed

src/components/DefineCompany/YourCompany.vue
severinbeauvais: OK, we've thought about it... and we can revisit later.

src/components/Company/CompanyName/CorrectNameRequest.vue
severinbeauvais: Oh right. And I think I asked you about this previously.
Could you please add a comment saying the request handles its own errors?

cameron-freshworks: Of course!

src/components/Company/CompanyName/CorrectNameRequest.vue
severinbeauvais: So much better now! 👍

src/components/Company/CompanyName/CorrectCompanyName.vue
severinbeauvais: It's still defaulted to '' ?

cameron-freshworks: Hmm, do i need a default? It would feel weird to default to a specific type?

src/components/Company/CompanyName/CorrectNameOptions.vue
```diff
@@ -42,7 +53,8 @@
 import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
 
 // Components
-import CorrectNameRequest from './CorrectNameRequest.vue'
+import CorrectNameRequest from '@/components/Company/CompanyName/CorrectNameRequest.vue'
+import CorrectCompanyName from '@/components/Company/CompanyName/CorrectCompanyName.vue'
```
severinbeauvais, line 57: If you use an index file then you could import both components in 1 line.

src/components/Company/CompanyName/CorrectNameOptions.vue
```diff
@@ -67,23 +80,28 @@ export default class CorrectNameOptions extends Vue {
   // local properties
   private displayedOptions: Array<CorrectNameOptionIF> = []
   private panel = null as number
-  private submit = false
+  private formType = ''
+  private currentFormType = ''
```
severinbeauvais, line 84: Are these strings or CorrectionTypes? If the latter then init as null?
See also `identifyForm()` below.

src/components/Company/CompanyName/CorrectNameRequest.vue
```diff
@@ -1,58 +1,83 @@
...
+      <v-flex md1 class="pa-3 pl-0">
+        <v-btn x-small fab outlined disabled color="#1A5A96" class="step-icon">1</v-btn>
+      </v-flex>
+      <v-flex>
+        <v-text-field
+          v-model="nameRequestNumber"
+          filled
+          label="Enter the NR Number"
+          hint="Example: NR 1234567"
+          req
```
severinbeauvais, line 13: Not needed?

src/components/Company/CompanyName/CorrectNameRequest.vue
```diff
@@ -1,58 +1,83 @@
...
+  NameRequestIF,
+  NrCorrectionIF
+} from '@/interfaces'
+import { CorrectionTypes } from '@/enums'
 
 @Component({})
 export default class CorrectNameRequest extends Mixins(NameRequestMixin) {
   /** Form Submission Prop */
-  @Prop({ default: false }) submit: boolean
+  @Prop({ default: '' }) formType: CorrectionTypes
```
severinbeauvais, line 73: Should default be null instead?

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
```diff
@@ -1,5 +1,7 @@
 // Name Request State interface
 export interface NameRequestIF {
+  legalType: string
```
severinbeauvais, line 3: Could we use enum EntityTypes for this?
NB There's a (slow) move to use the same enum in FE and BE for this. I started something in Filings UI but it's incomplete overall. See https://github.com/bcgov/business-filings-ui/blob/master/src/enums/legalTypes.ts

src/store/getters/state-getters.ts
```diff
@@ -94,6 +99,11 @@ export const isNamedBusiness = (state: any): boolean => {
   return !!state.stateModel.nameRequest.nrNumber
 }
 
+/** The NR number of a name request. */
```
severinbeauvais, line 102: It's the whole NR object, not just the number.

src/components/Company/CompanyName/CorrectCompanyName.vue
severinbeauvais: Null?

cameron-freshworks: Sure!

severinbeauvais (approved):
Looks really good overall.

I like the component break-down. It makes for smaller/simpler components that do only specific things.

What about ... unit tests?

src/components/Company/CompanyName/CorrectNameOptions.vue
cameron-freshworks: Due to the index file, the file where we import and having the files exporting all from the same directory was actually causing some weirdness here.

src/components/Company/CompanyName/CorrectNameOptions.vue
cameron-freshworks: Will update

src/components/Company/CompanyName/CorrectNameRequest.vue
cameron-freshworks: Fixed!

src/components/Company/CompanyName/CorrectNameRequest.vue
cameron-freshworks: Updaetd!

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: For sure, will apply the one we have for entityTypes 👍

src/store/getters/state-getters.ts
cameron-freshworks: Fixed

src/components/Company/CompanyName/CorrectNameOptions.vue
cameron-freshworks: > Looks really good overall.
> 
> I like the component break-down. It makes for smaller/simpler components that do only specific things.
> 
> What about ... unit tests?

This is a great question, i feel like these tickets were a little under-estimated due to being our first correction within a new Edit-UI. That being said, i've been playing catch up and have put them off. Are we ok to create a tech debt ticket, bring it in for next spring (if approved) and i can certainly explain why to the team.

cameron-freshworks:
> Looks really good overall.
> 
> I like the component break-down. It makes for smaller/simpler components that do only specific things.
> 
> What about ... unit tests?

This is a great question, i feel like these tickets were a little under-estimated due to being our first correction within a new Edit-UI. That being said, i've been playing catch up and have put them off. Are we ok to create a tech debt ticket, bring it in for next spring (if approved) and i can certainly explain why to the team.

src/components/Company/CompanyName/CorrectNameOptions.vue
severinbeauvais: OK. If I remember then I will have a look at this later.

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
severinbeauvais: Sounds good.

severinbeauvais:
> This is a great question, i feel like these tickets were a little under-estimated due to being our first correction within a new Edit-UI. That being said, i've been playing catch up and have put them off. Are we ok to create a tech debt ticket, bring it in for next spring (if approved) and i can certainly explain why to the team.

I agree with you. Some of the work I did last week was significant but not obvious from the ticket(s).

I'm OK with deferring the unit tests for the short term (and estimating them separately).

src/components/Company/CompanyName/CorrectCompanyName.vue
severinbeauvais: Gaah, is the comment still correct for this prop?

src/components/Company/CompanyName/CorrectNameOptions.vue
```diff
@@ -67,23 +80,28 @@ export default class CorrectNameOptions extends Vue {
   // local properties
   private displayedOptions: Array<CorrectNameOptionIF> = []
   private panel = null as number
```
severinbeauvais, line 82: I'm not sure what this resolves to. Normally for a "non-value" for a number you'd use NaN.

In this case it may be falsey but that's a dangerous check because 0 is also falsey but is a valid number!

See https://www.w3schools.com/jsref/jsref_isnan.asp

cameron-freshworks: kk i'll look at this!

cameron-freshworks:
> > This is a great question, i feel like these tickets were a little under-estimated due to being our first correction within a new Edit-UI. That being said, i've been playing catch up and have put them off. Are we ok to create a tech debt ticket, bring it in for next spring (if approved) and i can certainly explain why to the team.
> 
> I agree with you. Some of the work I did last week was significant but not obvious from the ticket(s).
> 
> I'm OK with deferring the unit tests for the short term (and estimating them separately).

Excellent, sounds good. 
Also, even though i can do some unit tests on the individual subcomponents, it will also make sense to test the parent components ( options and yourCompany ) once we have all three sub options as well. 
Which will be very soon, once this PR is resolved i will put up the other 👍

src/components/Company/CompanyName/CorrectNameOptions.vue
```diff
@@ -67,23 +80,28 @@ export default class CorrectNameOptions extends Vue {
...
-  private submit = false
+  private formType = null
+  private currentFormType = null
```
severinbeauvais, line 84: What type are these?

Sorry for being strict/picky but I think we should use types as much as possible to better catch errors during development.

Here, the type cannot be inferred from the initial value, so you should specify it. (In TS with Angular you get warned when specifying a type that is already inferred, so I now leave it out in those situations.)

My basic guideline: I should be able to hover over any property and get some useful type information. Only use 'any' sparingly.

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
```diff
@@ -1,7 +1,11 @@
+import { EntityTypes } from '@/enums'
+
 // Name Request State interface
 export interface NameRequestIF {
+  legalType: EntityTypes
+  legalName: string
   nrNumber: string
-  entityType: string
+  entityType: EntityTypes
```
severinbeauvais, line 8: Why do we have both legalType and entityType?

If it's legit and the property can't be named something better, please add comments here.

src/components/Company/CompanyName/CorrectNameOptions.vue
cameron-freshworks: No problem at all! 

Updated 👍

severinbeauvais:
> Also, even though i can do some unit tests on the individual subcomponents, it will also make sense to test the parent components ( options and yourCompany ) once we have all three sub options as well.
> Which will be very soon, once this PR is resolved i will put up the other 👍

OK, create tickets as needed.

(It would be neat to test the parents without needing to mount the children eh?)

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: This i'm not sure of, name request  ( as it was for create ) utilized entityType i think? 
What the Correction schema looks for is `legalType`. I didn't look into removing entityType entirely or instances of where it is utilized as of yet.

severinbeauvais: The schema specifies only `legalType`.

A sample IA I'm looking at uses `legalType`.

I think we should get rid of "entityType" from the NR and convert back and forth as needed (using parseFiling() and buildFiling()).

cameron-freshworks: Done and done 👍

cameron-freshworks:
> > Also, even though i can do some unit tests on the individual subcomponents, it will also make sense to test the parent components ( options and yourCompany ) once we have all three sub options as well.
> > Which will be very soon, once this PR is resolved i will put up the other 👍
> 
> OK, create tickets as needed.
> 
> (It would be neat to test the parents without needing to mount the children eh?)

1 Ticket should be good to cover all the bases from Company Name Changes! 

:) We will certainly try

src/components/Company/CompanyName/CorrectNameOptions.vue
cameron-freshworks: Yea, i'm not too sure about this one.

severinbeauvais: Oh, hehe, looks like I used "null" for "no panel" in Filings UI.

The Vuetify docs just say this is of type "any" :/

Well, leave it for now then. Just be aware not to check for "if (panel)".

vysakh-menon-aot (approved):

PR #23, opened 2020-09-17
Correction - articles component

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4765

*Description of changes:*
Modified the agreement type component to include correction features
Added/Modified tests

### DIFF
--- a/src/components/IncorporationAgreement/AgreementType.vue
+++ b/src/components/IncorporationAgreement/AgreementType.vue
@@ -75,7 +75,7 @@
 <script lang="ts">
 // Libraries
 import { Component, Emit, Vue, Prop, Watch } from 'vue-property-decorator'
-import { Action, State } from 'vuex-class'
+import { Action, Getter } from 'vuex-class'
 
 // Interfaces
 import { ActionBindingIF, IncorporationFilingIF } from '@/interfaces'
@@ -86,11 +86,8 @@ export default class AgreementType extends Vue {
   private incorporationAgreementTypeResource = AgreementTypeResource
 
   // State
-  @State(state => state.stateModel.incorporationAgreementStep.agreementType)
-  readonly agreementTypeState: string | null
-
-  @State(state => state.stateModel.originalIA)
-  readonly originalIA!: IncorporationFilingIF
+  @Getter getAgreementType!: string
+  @Getter getOriginalIA!: IncorporationFilingIF
 
   // Global setters
   @Action setIncorporationAgreementStepData!: ActionBindingIF
@@ -102,7 +99,7 @@ export default class AgreementType extends Vue {
 
   // Lifecycle methods
   private created (): void {
-    this.agreementType = this.agreementTypeState
+    this.agreementType = this.getAgreementType
   }
 
   mounted (): void {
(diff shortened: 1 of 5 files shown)


src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -61,18 +89,16 @@ export default class AgreementType extends Vue {
   @State(state => state.stateModel.incorporationAgreementStep.agreementType)
   readonly agreementTypeState: string | null
 
-  @Prop({ default: false })
-  private showErrorSummary: boolean
-
-  @Prop({ default: false })
-  private isSummary: boolean
+  @State(state => state.stateModel.originalIA)
+  readonly originalIA!: IncorporationFilingIF
```
severinbeauvais, line 93: I think we should stop using State (since that it "looking into the store structure") and instead use getters.
I think Cam created a getter for Original IA already.
Please also use a getter for Agreement Type.

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -102,6 +139,18 @@ export default class AgreementType extends Vue {
         .summaryDescription
     } else { return '' }
   }
+
+  @Emit('haveChanges')
```
severinbeauvais, line 143: Can we change this to saving it in the store instead?
Also we will need a getter for App to use to display/hide the Fee Summary. (Similar getter to `isApplicationValid` I think?)

severinbeauvais (commented):
Wow, fast and looks good!
I'm hoping for a couple of simple changes to keep things simple and consistent with other components currently being developed. Thanks!

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: Sure. Forgot about that. Will do that

vysakh-menon-aot (approved):

src/components/IncorporationAgreement/AgreementType.vue
cameron-freshworks: I don't think i implemented a Getter for the entirety of the Original IA, i actually did as Lekshmi did above. We could refactor that tho! See also line 180 in `YourCompany` if you do create an Original IA getter.

cameron-freshworks (approved):
Looks good, just a few comments!

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: Done

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: Looked into this. I think we will have changes to the company, agreement type, correction and App components for this. I will either createa a follow up PR once Cameron's PR is in. It will be easy to do all that together.

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: Hah, she implemented it! Well done :)

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: OK.
I looked into also and there are 2 things to think about:
1. There is a `haveChanges` store property to keep track of whether the user has done any changes (to prevent their exit if they have unsaved changes). I think the component-has-changed flags can replace this.
2. App will have to check whether there are any correction changes or alteration changes or ??? when deciding whether to display the Fee Summary. This is probably a simple getter with A || B || C but it doesn't exist yet.

severinbeauvais (commented):
I think I have other comments to post...

severinbeauvais:
> I think I have other comments to post...

Nope, lost them. Sec...

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -1,82 +1,105 @@
...
   @Action setHaveChanges!: ActionBindingIF
 
   // Local properties
   private agreementType: string | null = null
+  private showAgreementTypeForm = false
 
   // Lifecycle methods
   private created (): void {
-    this.agreementType = this.agreementTypeState
+    this.agreementType = this.getAgreementType
```
severinbeauvais, line 102: Are you sure Agreement Type is set when this component is created? I'm not so sure.

I don't know if this is preferable, but you could set `{ immediate: true }` on your watcher below, which will cause it to fire when this component is created.

severinbeauvais (approved):

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: I think it is a good idea to add immediate .. that way I dont need that setter in the created life cycle method

PR #24, opened 2020-09-17
Save and File Corrections

cameron-freshworks:
*Issue #:* /bcgov/entity#4946

*Description of changes:*

* Updates the filing templates to build and parse Correction Filings

* Updated Interfaces for future correction interchangeability 

* Updating the legal api mixin to handle Correction filings

* Refactored Correction view to assume we are always resuming a correction draft created in the filings UI when launching a new IA Correction.

*Notes:*
* In order to achieve a complete end 2 end flow with IA Correction, there is still work to be done in the Filings Ui to handle...
* Creating the Draft IA Correction before routing to the Edit-UI with the Correction filing Id.
* Resuming a draft ( only staff ) with the Correction filing Id.
* Functionality to Delete the Draft ( only staff ).

* The work described above is captured in story #4582 


lekshmimallika-aot (approved):
Looks good!

src/App.vue
```diff
@@ -94,7 +94,7 @@
                 @fetchError="fetchErrorDialog = true"
                 @haveData="haveData = true"
                 @filingData="filingData = $event"
-                @haveChanges="showFeeSummary = $event"
+                @haveChanges="stateChangeHandler($event)"
```
severinbeauvais, line 97: OK for now but we've been talking about getting of this event in favour of store flags. Maybe.

src/constants/constants.ts
```diff
@@ -2,3 +2,4 @@ export const ADDRESSCHANGED = 'addressChanged'
 export const INCORPORATION_APPLICATION = 'incorporationApplication'
 export const ISFUTUREEFFECTIVE = 'isFutureEffective'
 export const ISIMMEDIATE = 'isImmediate'
+export const CORRECTION = 'correction'
```
severinbeauvais, line 5: I think you should import Filing Types from another project instead.

This will also allow us to verify that header=="incorporationApplication" when we load the corrected filing (which you can add to onAppReady() in Correction.vue).

vysakh-menon-aot (approved):

src/interfaces/filing-interfaces/incorporationApplication-interface.ts
```diff
@@ -0,0 +1,26 @@
+import { IncorporationAddressIf, OrgPersonIF, ShareClassIF } from '@/interfaces'
+
+/** Incorporation Application filing loaded from / saved to the Legal API. */
+export interface IncorporationApplicationIF {
+  nameRequest: {
+    legalType: string
+    nrNumber?: string
+    legalName?: string // only set when there is an NR
```
severinbeauvais, line 8: This comment also applies to line 7.

src/constants/constants.ts
severinbeauvais: Or at least separate the "loose constants" from the "filing type" constants :)

src/views/Correction.vue
```diff
@@ -179,7 +172,19 @@ export default class Correction extends Mixins(DateMixin, FilingTemplateMixin, L
 
     // try to fetch data
     try {
-      if (this.correctionId) {
+      // set current entity type
+      this.setEntityType(EntityTypes.BCOMP)
+
+      // initialize Fee Summary data
+      this.emitFilingData([{
```
severinbeauvais, line 179: FYI (future TODO): we need to set/clear the filing data according to whether we have changes.

src/App.vue
cameron-freshworks: kk, copy that. Agile!

src/interfaces/filing-interfaces/incorporationApplication-interface.ts
cameron-freshworks: Updated in next commit!

src/constants/constants.ts
cameron-freshworks: Interested, do we have a Constants declaration file available like this? Are you thinking we relocate this to the bcrs-shared components and import? 
If that is the case, we should also do that with A LOT of our interfaces. That would be highly useful when if available across all the UI's :)

src/views/Correction.vue
cameron-freshworks: Copy that, i will add a comment.

src/constants/constants.ts
severinbeauvais: We have some constants in Filings UI, as well as enums/types.

I wasn't thinking of moving this to Shared Components but it sounds like a good idea (except that it's an enum instead of a component) (and we'd need to import it using npm)...?

If someone has a good idea how to do this, please share. For now let's be Agile and not change it yet :)

src/views/Correction.vue
severinbeauvais: Or... make it work?

severinbeauvais: I'm OK either way.

cameron-freshworks: Absolutely, will do!

src/constants/constants.ts
cameron-freshworks: Let's do a little research on this, i can see value in that. Having interfaces/enums/constants shared across the 3 UI's would be amazing for consistency and cut down on the code.

severinbeauvais: OK but if it takes significant time, please create a new ticket.

PS This feels similar to our common schemas...
https://github.com/bcgov/business-schemas

src/views/Correction.vue
cameron-freshworks: For sake of expediency, added a TODO here. Will be applicable when implementing Alterations and beyond.

src/constants/constants.ts
severinbeauvais: Future...

src/App.vue
severinbeauvais: Future...

severinbeauvais (approved):

PR #27, opened 2020-09-18
4549 Insert staff payments into incorp app correction and make this a common component

vysakh-menon-aot:
*Issue #:* /bcgov/entity#4549

*Description of changes:*
 - Integrated StaffPayment shared component
 - Reading and Writing Staff Payment API data
 - Padding fix for router-view container in App (It was displaying with Zero space in left side of the screen)
 - Update the filing data that is sent to the Fee Summary component from Staff payment component


src/components/common/StaffPayment.vue
```diff
@@ -4,17 +4,88 @@
...
+
+  private onStaffPaymentData (event) {
+    let staffPaymentData: StaffPaymentIF = { ...this.getStaffPayment, ...event }
+    switch (staffPaymentData.option) {
+      case StaffPaymentOptions.FAS:
+        staffPaymentData = {
+          option: StaffPaymentOptions.FAS,
+          routingSlipNumber: staffPaymentData.routingSlipNumber,
+          isPriority: staffPaymentData.isPriority,
+          bcolAccountNumber: '',
```
severinbeauvais, line 56: In Filings UI we don't include these properties and I'm not sure what the API will do with them present (even if empty).
I suggest you update the IF to specify which properties are optional, and then don't specify them here (ditto in BCOL and NO_FEE cases).

src/components/common/StaffPayment.vue
```diff
@@ -4,17 +4,88 @@
...
+          isPriority: staffPaymentData.isPriority,
+          routingSlipNumber: ''
+        }
+        break
+
+      case StaffPaymentOptions.NO_FEE:
+        staffPaymentData = {
+          option: StaffPaymentOptions.NO_FEE,
+          routingSlipNumber: '',
+          isPriority: false,
```
severinbeauvais, line 77: Although this wasn't set in the Filings UI for NO_FEE, this may in fact be correct ... Please verify with save and resume, and when filing is functional, with an actual filing.

src/components/common/StaffPayment.vue
```diff
@@ -4,17 +4,88 @@
...
+  }
+})
 export default class StaffPayment extends Vue {
+  // Global getters
+  @Getter getStaffPayment!: StaffPaymentIF
+
+  // Global setters
+  @Action setStaffPayment!: ActionBindingIF
+
+  private staffPaymentFormValid: boolean = null
```
severinbeauvais, line 40: Null? Or false?

src/store/state/state-model.ts
```diff
@@ -103,5 +103,13 @@ export const stateModel: StateModelIF = {
         agreementType: ''
       }
     }
+  },
+  staffPayment: {
+    option: NaN,
+    routingSlipNumber: '',
```
severinbeauvais, line 109: As per my previous comment, would it better to _not_ assign some of these properties?
(And not set priority to false?)

src/components/common/StaffPayment.vue
```diff
@@ -4,17 +4,88 @@
...
+
+  /** Emits Valid event. */
+  @Emit('emitValid')
+  private emitValid (): boolean {
```
severinbeauvais, line 44: This is never called.
Possibly the `@valid` should call it?
Actually, we said we'd save validity in the store right?

severinbeauvais (commented):
Please see my comments.

src/components/common/StaffPayment.vue
vysakh-menon-aot: > Null? Or false?

It shouldn't be a problem as it won't be emitted immediately, but will change it to false since it make more sense

src/components/common/StaffPayment.vue
vysakh-menon-aot: It won't be included in API call. That has been handled in filing-template-mixin.ts

src/components/common/StaffPayment.vue
severinbeauvais: Yes please.

src/components/common/StaffPayment.vue
severinbeauvais: Hmm, you shouldn't be getting a TS error since those properties are defined as optional.

vysakh-menon-aot: True, but it was not optional in shared component. That's how it is been implemented in filing-ui.

src/components/common/StaffPayment.vue
vysakh-menon-aot: while calling API we are setting waiveFees as true if NO_FEE

src/components/common/StaffPayment.vue
vysakh-menon-aot: Fixed it

src/components/common/StaffPayment.vue
severinbeauvais: OK, that's fine.

src/store/state/state-model.ts
severinbeauvais: Never mind. Answered in another conversation.

src/components/common/StaffPayment.vue
severinbeauvais: Ah, OK.
It would be good for the IF to be the same in the component and in this project, yes?

src/components/common/StaffPayment.vue
```diff
@@ -4,17 +4,89 @@
...
-@Component({})
+// Components
+import { StaffPayment as StaffPaymentComponent } from '@bcrs-shared-components/staff-payment'
+
+// Mixins, Interfaces and Enums
+import { FilingTemplateMixin } from '@/mixins'
+import { ActionBindingIF } from '@/interfaces'
+
+import { StaffPaymentIF } from '@bcrs-shared-components/interfaces'
+import { StaffPaymentOptions } from '@bcrs-shared-components/enums'
```
cameron-freshworks, line 26: Very cool, would love to see universal interfaces/enums across all our UI's one day.

cameron-freshworks (approved):
Looks good 👍

lekshmimallika-aot (approved):

severinbeauvais (approved):
The Filing Data changes look good.

src/components/common/StaffPayment.vue
severinbeauvais: OK for now. We'll resolve this later.

PR #36, opened 2020-09-24
Framework updates and fixes

lekshmimallika-aot:
*Issue #:* /bcgov/entity#5092

*Description of changes:*

- Added new flag at the steplevel that indicates whether a correction has been made to that component
- New getter that looks at those change flags to identify whether any correction has been done to the application
- Tied the staff payment validity to the store to enable/disable file and pay
- Changed the showFee summary to look at whether any modification has been done to the filing rather than 'haveChanges'. 
   This fixes the problem of fee summary not getting displayed while resuming a draft with correction
- Modified actions buttons to look at the validity and change
- Removed haveCorrection from tombstone

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -177,7 +177,6 @@ export default class App extends Mixins(BcolMixin, CommonMixin, DateMixin, Filin
   @Action setCurrentDate!: ActionBindingIF
   @Action setAuthRoles: ActionBindingIF
   @Action setHaveChanges!: ActionBindingIF
-  @Action setHaveCorrection!: ActionBindingIF
   @Action setAccountInformation!: ActionBindingIF
   @Action setKeycloakRoles!: ActionBindingIF
   @Action setUserInfo: ActionBindingIF
--- a/src/App.vue
+++ b/src/App.vue
@@ -93,7 +93,6 @@
                 @profileReady="profileReady = true"
                 @fetchError="fetchErrorDialog = true"
                 @haveData="haveData = true"
-                @haveChanges="stateChangeHandler($event)"
               />
             </v-col>
             <v-col cols="12" lg="3" style="position: relative">
@@ -101,8 +100,7 @@
                 <aside>
                   <affix
                     relative-element-selector=".col-lg-9"
-                    :offset="{ top: 86, bottom: 12 }"
-                  >
+                    :offset="{ top: 86, bottom: 12 }">
                     <sbc-fee-summary
                       :filingData="[...getFilingData]"
                       :payURL="payApiUrl"
@@ -142,7 +140,7 @@ import * as Views from '@/views'
 import * as Dialogs from '@/components/dialogs'
 
 // Mixins, interfaces, etc
-import { BcolMixin, DateMixin, FilingTemplateMixin, LegalApiMixin } from '@/mixins'
+import { BcolMixin, CommonMixin, DateMixin, FilingTemplateMixin, LegalApiMixin } from '@/mixins'
 import { FilingDataIF, ActionBindingIF, ConfirmDialogType } from '@/interfaces'
 import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
 
@@ -157,7 +155,7 @@ import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
     ...Views
   }
 })
-export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixin, LegalApiMixin) {
+export default class App extends Mixins(BcolMixin, CommonMixin, DateMixin, FilingTemplateMixin, LegalApiMixin) {
   // Refs
   $refs!: {
     confirm: ConfirmDialogType
@@ -172,6 +170,7 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
   @Getter getUserRoles!: string
   @Getter getUserUsername!: string
   @Getter getFilingData!: FilingDataIF
+  @Getter isFilingChanged!: boolean
 
   // Global setters
   @Action setBusinessId!: ActionBindingIF
@@ -198,7 +197,6 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
   private saveErrors: Array<object> = []
   private saveWarnings: Array<object> = []
   private fileAndPayInvalidNameRequestDialog: boolean = false
-  private showFeeSummary: boolean = false
 
   // FUTURE: change profileReady/appReady/haveData to a state machine?
 
@@ -248,10 +246,14 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
     return Boolean(sessionStorage.getItem(SessionStorageKeys.KeyCloakToken))
   }
 
-  /** Handle State Changes */
-  private stateChangeHandler (hasStateChanges: boolean): void {
-    this.showFeeSummary = hasStateChanges
-    this.setHaveCorrection(hasStateChanges)
+  private get showFeeSummary (): boolean {
+    const defaultFilingData = {
+      filingTypeCode: null,
+      entityType: null,
+      priority: false,
+      waiveFees: false
+    }
+    return this.isFilingChanged && !this.isSame(this.getFilingData, defaultFilingData)
   }
 
   /**
@@ -462,7 +464,6 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
   private resetFlags (): void {
     this.appReady = false
     this.haveData = false
-    this.showFeeSummary = false
     this.bcolObj = null
     this.nameRequestInvalidErrorDialog = false
     this.invalidIncorporationApplicationDialog = false
--- a/src/App.vue
+++ b/src/App.vue
@@ -245,6 +245,9 @@ export default class App extends Mixins(BcolMixin, CommonMixin, DateMixin, Filin
     return Boolean(sessionStorage.getItem(SessionStorageKeys.KeyCloakToken))
   }
 
+  /** Safety check to ensure that fee summary component is not loaded until there
+   *  is a valid filing type and entity code.
+   */
   private get showFeeSummary (): boolean {
     const defaultFilingData = {
       filingTypeCode: null,
--- a/src/components/IncorporationAgreement/AgreementType.vue
+++ b/src/components/IncorporationAgreement/AgreementType.vue
@@ -96,22 +96,23 @@ export default class AgreementType extends Vue {
   private showAgreementTypeForm = false
 
   mounted (): void {
-    this.setIncorporationAgreementStepData({
-      valid: !!this.agreementType,
-      agreementType: this.agreementType
-    })
-
+    this.setIncorporationAgreementData()
     // now that all data is loaded, wait for things to stabilize and reset flag
     Vue.nextTick(() => this.setHaveChanges(false))
   }
 
   private setAgreementType (): void {
-    this.setIncorporationAgreementStepData({
-      valid: !!this.agreementType,
-      agreementType: this.agreementType
-    })
+    this.setIncorporationAgreementData()
     this.emitHaveChanges()
     this.showAgreementTypeForm = false
+  }
+
+  private setIncorporationAgreementData (): void {
+    this.setIncorporationAgreementStepData({
+      valid: !!this.agreementType,
+      changed: this.hasAgreementTypeChange,
+      agreementType: this.agreementType
+    })
   }
 
   private resetAgreementType (): void {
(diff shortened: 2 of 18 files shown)


src/App.vue
```diff
@@ -248,10 +245,14 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
...
-    this.showFeeSummary = hasStateChanges
-    this.setHaveCorrection(hasStateChanges)
+  private get showFeeSummary (): boolean {
+    const defaultFilingData = {
+      filingTypeCode: null,
+      entityType: null,
+      priority: false,
+      waiveFees: false
+    }
+    return this.isFilingChanged && !this.isSame(this.getFilingData, defaultFilingData)
```
severinbeauvais, line 255: I don't understand why you need to check for different filing data...?

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -96,22 +96,23 @@ export default class AgreementType extends Vue {
   private showAgreementTypeForm = false
 
   mounted (): void {
-    this.setIncorporationAgreementStepData({
-      valid: !!this.agreementType,
-      agreementType: this.agreementType
-    })
-
+    this.setIncorporationAgreementData()
```
severinbeauvais, line 99: Is this needed, since the watch on line 141 has "immediate=true" (meaning it fires when the component is created I think) ?

src/components/YourCompany/YourCompany.vue
```diff
@@ -241,17 +244,19 @@ export default class YourCompany extends Mixins(DateMixin, EntityFilterMixin, Le
...
       this.nameTranslationChanges ||
       this.officeAddressChanges
-    )
+    this.setDefineCompanyStepChanged(haveChanges)
+    this.emitHaveChanges(haveChanges)
   }
+
+  /** Emits Have Changes event. */
+  @Emit('haveChanges')
+  private emitHaveChanges (haveChanges: boolean): void {}
```
severinbeauvais, line 259: Why did you split up the original method into 2 methods?

src/components/common/Actions.vue
```diff
@@ -4,7 +4,7 @@
     <div class="buttons-left">
       <!-- disable Save button for now -->
       <v-btn id="save-btn" large
-        :disabled="!isEntityType || isBusySaving"
+        :disabled="!isEntityType || isBusySaving || !isFilingValid"
```
severinbeauvais, line 7: Previously we allowed saving even when the filing was invalid. Please check with Linda and/or Scott whether we want to make this change. (Personally I can't decide.)

src/components/common/Actions.vue
```diff
@@ -24,7 +24,7 @@
     <div class="buttons-right">
       <v-fade-transition hide-on-leave>
         <v-btn id="file-pay-btn" large color="primary"
-          :disabled="!getHaveCorrection || isBusySaving"
+          :disabled="!isFilingChanged || isBusySaving || !isFilingValid || !isStaffPaymentValid"
```
severinbeauvais, line 27: Why isn't "staff payment validity" simply part of the filing validity?

src/store/getters/state-getters.ts
```diff
@@ -203,17 +203,16 @@ export const getStaffPayment = (state: any): any => {
   return state.stateModel.staffPayment
 }
 
+/** The staff payment validity. */
+export const isStaffPaymentValid = (state: any): any => {
```
severinbeauvais, line 207: Please set return type.

src/store/getters/state-getters.ts
```diff
@@ -241,14 +240,18 @@ export const isBusySaving = (state: any): boolean => {
...
+    state.stateModel.incorporationAgreementStep.changed)
 }
 
-/** Whether all the correction sections are valid. */
-export const isCorrectionValid = (state: any): boolean => {
-  // TODO: add other sections here
+/** Whether any correction/alteration sections have changed. */
+export const isFilingValid = (state: any): boolean => {
+  // Add sections that can have only invalid data like people and roles and share structure
+  // Define company, Agreement Type wont allow saving and invalid state to the store
```
severinbeauvais, line 255: Staff payment validity should go here I think.

src/store/state/state-model.ts
```diff
@@ -113,10 +115,12 @@ export const stateModel: StateModelIF = {
     folioNumber: '',
     isPriority: false
   },
+  staffPaymentValidity: false,
```
severinbeauvais, line 118: Please consider combining this and the staffPayment object above into a step object (same as the others).

src/components/common/Actions.vue
lekshmimallika-aot: I am c hecking the filing validity in save and resume. If that is not needed it can be part of the filing validity.

src/components/common/Actions.vue
lekshmimallika-aot: I will check that

src/store/getters/state-getters.ts
lekshmimallika-aot: As explained in the other comment, if we want to allow invalid filing to be saved I can remove the check from save action buttons and payment validity can be here.

src/App.vue
lekshmimallika-aot: because the isfilingChanged returns even before filing data get set. Looking at the code I dont understand why that happens. When that happens Fee Summary component errors out. So as a safety check I added this

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: Probably not. I will check

src/components/YourCompany/YourCompany.vue
lekshmimallika-aot: To set the change flag in the defineCompanyStep. I dont think it is right to set a state with in an emit method even though functionality will work that way

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -96,22 +96,23 @@ export default class AgreementType extends Vue {
...
     // now that all data is loaded, wait for things to stabilize and reset flag
     Vue.nextTick(() => this.setHaveChanges(false))
```
severinbeauvais, line 101: The global haveChanges flag was to prevent a user from exiting with unsaved changes, but it got complicated and I think it's not working correctly -- for example, does it currently identify unsaved changes, or any change from the original IA?

I think the best thing is to initialize this flag (to false) when the Correction view has finished loading, and then set it true when any data changes (in mutations-model.ts). (Some mutations already set the flag to true.)

Setting this flag in this component is not needed. (This code is left over from Create UI.)

src/components/common/Actions.vue
severinbeauvais: What I mean is that "staff payment validity" should be part of the "is filing valid" getter.

src/store/getters/state-getters.ts
severinbeauvais: Can payment validity be here, whether or not we allow invalid filing to be saved?

I think the only difference would be whether you check "isFilingValid" for the save buttons.

If I misunderstand, please explain. Thx!

src/App.vue
severinbeauvais: I think the store does think async (ie, in another tick) which is why things don't seem to get set right away.

Please add a comment explaining this safety check.

src/components/YourCompany/YourCompany.vue
severinbeauvais: OK

src/store/getters/state-getters.ts
```diff
@@ -241,14 +240,18 @@ export const isBusySaving = (state: any): boolean => {
...
-/** Whether any correction sections have changed. */
-export const isCorrectionChanged = (state: any): boolean => {
+/** Whether any correction/alteration sections have changed. */
+export const isFilingChanged = (state: any): boolean => {
   // TODO: add other sections here
-  return (state.stateModel.peopleAndRoles.changed)
+  return (state.stateModel.peopleAndRoles.changed ||
+    state.stateModel.defineCompanyStep.changed ||
+    state.stateModel.createShareStructureStep.changed ||
+    state.stateModel.incorporationAgreementStep.changed)
```
cameron-freshworks, line 249: Good consolidation of the states 👍

src/store/getters/state-getters.ts
cameron-freshworks: If staff payment validity is a check on the filing validity as a whole, then i believe it should be conditional based on the 'staff' role. Unless that's something we want to think about later. To be honest, i'm not sure what the plan is for client driven corrections..

cameron-freshworks (approved):
Just a few comments / questions. Overall looks great, i want to pull it in lol

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: Fixed

src/App.vue
lekshmimallika-aot: Done

src/components/common/Actions.vue
lekshmimallika-aot: Scott confirmed that saving should be enabled for invalid filing. So removed this condition from save / save and resume

src/components/common/Actions.vue
lekshmimallika-aot: Done

src/store/getters/state-getters.ts
lekshmimallika-aot: Not needed as this is now part of isFilingValid

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: Agree that it is overly complicated and very difficult to understand. I have removed that from component. Also added the mutation in setNameRequest which was not present before. I think the functionality works as expected now. Please confirm

src/store/getters/state-getters.ts
lekshmimallika-aot: I think we can do that when we do self serve alteration.

src/store/state/state-model.ts
lekshmimallika-aot: Done

src/store/getters/state-getters.ts
severinbeauvais: Resolving this conversation... I'll look at your update code ASAP.

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: Resolving this conversation... I'll look at your updated code ASAP.

src/store/state/state-model.ts
severinbeauvais: Way better - thanks!

src/components/common/Actions.vue
```diff
@@ -24,7 +24,7 @@
     <div class="buttons-right">
       <v-fade-transition hide-on-leave>
         <v-btn id="file-pay-btn" large color="primary"
-          :disabled="!getHaveCorrection || isBusySaving"
+          :disabled="!isFilingChanged || isBusySaving || !isFilingValid"
```
severinbeauvais, line 27: Does anything prevent File and Pay when a component is in editing mode?

src/interfaces/store-interfaces/state-interface.ts
```diff
@@ -1,9 +1,8 @@
 import {
   AccountInformationIF, CertifyIF, DefineCompanyIF, NameRequestIF, TombStoneIF, PeopleAndRolesIF,
   ShareStructureIF, DateTimeIF, IncorporationAgreementIF, BusinessInformationIF, IncorporationFilingIF,
-  FilingDataIF
+  FilingDataIF, StaffPaymentStepIF
 } from '@/interfaces'
```
severinbeauvais, line 5: Maybe we should simplify this import as I've done elsewhere using `*` and `...` :)

src/components/common/Actions.vue
lekshmimallika-aot: Not yet. This is something that needs to be implemented. Almost all components use 'isEditing' locally within component.

lekshmimallika-aot: It has to be another PR

severinbeauvais: OK. Please create another PR or ticket for this.

lekshmimallika-aot: https://github.com/bcgov/entity/issues/5115

src/interfaces/store-interfaces/state-interface.ts
lekshmimallika-aot: Will implement it in the next PR

lekshmimallika-aot:
Merging this PR and the outstanding comment will be incorporated in the PR for wiring Detail component

PR #43, opened 2020-10-01
4764 Correct Share Structure

cameron-freshworks:
*Issue #:* /bcgov/entity#4764

*Description of changes:*

* Renamed ListShareStructure into ShareStructure.

* Implemented functionality to handle the varying states of Share Classes and Series

* Consolidated the ADD/EDIT Class and Series Shares form into the ShareStructure component and renamed EditShareStructure.

* Created common chip component to display states. 
( I didn't implement in all use cases as it was creating conflicts with ongoing work. Can be implemented case by case refactors )

*Notes: *

* Temporarily disabled unit testing.
( Currently WIP, but wanted to get eyes on this  sooner than later. )

* Please be CRITICAL in review. This was a very large/complex bit of work done while i have been on cold medicine 😆.

* Will be applying the Unit tests in a follow up PR. 


lekshmimallika-aot (approved):
Looks good to me!

src/components/common/ActionChip.vue
```diff
@@ -0,0 +1,27 @@
...
+</template>
+
+<script lang="ts">
+import { Component, Prop, Vue } from 'vue-property-decorator'
+import { ActionTypes } from '@/enums'
+
+@Component({})
+export default class ActionChip extends Vue {
+  @Prop({ default: {} })
+  private actionableItem: {}
```
severinbeauvais, line 19: In the future, you might be able to use a partial type definition here to ensure that property "action" exists on the object. Dunno if this can be done; just a thought.

https://www.google.com/search?q=typescript+partial

src/mixins/filing-template-mixin.ts
```diff
@@ -66,11 +66,22 @@ export default class FilingTemplateMixin extends Vue {
...
-    // if filing and paying, filter out removed orgs/persons and omit the 'action' property
+    // if filing and paying, filter out removed entities and omit the 'action' property
     let parties = this.getPeopleAndRoles
+    let shareStructure = this.getShareClasses
     if (!isDraft) {
       parties = parties.filter(x => x.action !== 'removed')
         .map((x) => { const { action, ...rest } = x; return rest })
+
+      // Filter out class actions
+      shareStructure = shareStructure.filter(x => x.action !== 'removed')
```
severinbeauvais, line 77: Oh snap. All 3 of these should use the ActionTypes enum :-}

severinbeauvais: And please add a comment for orgs/persons. Thx.

cameron-freshworks: oo right, i'll update them all 👍

src/components/common/ActionChip.vue
cameron-freshworks: Great idea, will look into this

src/mixins/filing-template-mixin.ts
```diff
@@ -109,7 +120,7 @@ export default class FilingTemplateMixin extends Vue {
         },
         parties,
         shareStructure: {
-          shareClasses: this.getShareClasses
+          shareClasses: shareStructure
```
severinbeauvais, line 123: Or you could name it the same and use the object property value shorthand.
https://alligator.io/js/object-property-shorthand-es6/

src/views/Alteration.vue
```diff
@@ -13,11 +13,14 @@
...
       :isSummary="true"
     />
-    <list-share-class
+
+    <share-structure
       class="mt-10"
       :shareClasses="getShareClasses"
-      :isSummary="true"
+      @setShareClass="setShareClasses($event)"
+      @haveChanges="shareStructChanges = $event"
```
severinbeauvais, line 21: I believe we're getting away from setting data (or managing a ton of state data) in the views. The ShareStructure component can simply store its data (and whether it has changes).

severinbeauvais: Same comment for the Correction view.

Have a look at `isFilingChanged()`, `isFilingValid()`, etc.

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,490 @@
...
+                  :label="shareStructure.type + ' Name [Shares]'"
+                  :hint="'Enter the name of the  '+ shareStructure.type.toLowerCase() +
+                  '  - the words &quot;Shares&quot; is automatically added'"
+                  id="txt-name"
+                  v-model="shareStructure.name"
+                  :rules="getNameRule()"
+                  suffix="Shares"
+                  persistent-hint/>
+
+                <v-divider class="separator" />
```
severinbeauvais, line 35: In other places, separators have a left and right margin (which I think looks better). You may want to add class "mx-4" here and elsewhere.

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,490 @@
...
+                          </v-select>
+                        </v-col>
+                      </v-row>
+                    </template>
+                  </v-radio>
+                  <v-radio :value="true" label="No par value" id="radio-no-par"/>
+                </v-radio-group>
+
+                <div v-show="isSeries">
+                    <v-row v-if="shareStructure.hasParValue">
```
severinbeauvais, line 105: Change indent width to "2 spaces" not 4. (It does this to me sometimes too.)

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,490 @@
...
+                        </v-col>
+                        <v-col cols="6">
+                            <v-text-field
+                            id="series-currency"
+                            label="Currency"
+                            :value="`${getCurrencyNameByCode(shareStructure.currency)} (${shareStructure.currency})`"
+                            :disabled="true"/>
+                        </v-col>
+                    </v-row>
+                    <v-label id='lbl-no-par' v-else>No par value</v-label>
```
severinbeauvais, line 122: Change ' to ".

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,490 @@
...
+                </div>
+
+                <v-divider class="separator" />
+
+                <div class="form__row">{{hasSeriesShares}}
+                  <v-checkbox
+                     id="special-rights-check-box"
+                    :label="'This share ' + shareStructure.type.toLowerCase() + ' has special rights or restrictions'"
+                    v-model="shareStructure.hasRightsOrRestrictions"
+                    @click="confirmSeriesRemoval"
```
severinbeauvais, line 132: I prefer to use () on event handlers to make it clear that:
- it's a method
- it doesn't (or does) use formal parameters (eg, $event)

src/views/Alteration.vue
cameron-freshworks: ok, no prob. Will refactor a few things here, on `Correction` and in the `ShareStructure` component!

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,490 @@
...
+
+                  <v-btn
+                    large
+                    id="done-btn"
+                    class="form-primary-btn"
+                    color="primary"
+                    @click="validateForm()"
+                    :disabled="!formValid"
+                  >
+                    Done
```
severinbeauvais, line 148: No change required, but if you're wondering why we somethings use `<span>` around button text, it's to make it easier to fetch just the text in the unit tests... otherwise the text may contain some line returns (so you have to "expect" that the string "contains" the string instead of "equals" the string). FYI

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,490 @@
...
+    if (!this.hasNoParValue) {
+      return [v => !!v || 'Currency is required']
+    }
+    return []
+  }
+
+  /** Called when component is created. */
+  private created (): void {
+    if (this.initialValue) {
+      this.shareStructure = { ...this.initialValue }
```
severinbeauvais, line 295: If this component is inside a "v-if" in its parent then the initialValue prop is probably set, otherwise I have found that initial props are NOT set until the data is read it (from API), meaning that `created()` has little value in some components. YMMV but maybe worth checking.

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,490 @@
...
+        this.shareStructure.hasRightsOrRestrictions = false
+      }).catch(() => {
+        // if we get here, Cancel was clicked
+        this.shareStructure.hasRightsOrRestrictions = true
+      })
+    }
+  }
+
+  // Getters
+  get isClass (): boolean {
```
severinbeauvais, line 394: FWIW, I like having my getters near the top of the script since they are kind of like "local properties".

Here, while reviewing, I had to "find on page" to find this declaration as it wasn't where I was expecting.

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,490 @@
+<template>
+  <div>
+
+    <confirm-dialog
+      ref="confirm"
+      attach="#app"
+    />
+
+    <v-expand-transition id="addShareStructureContainer">
```
severinbeauvais, line 9: It might be nice to:
- move this id to the upper div
- name it after the file name
- attach the Confirm Dialog to it (I don't know if `#app` is visible in this file)

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -1,427 +1,731 @@
...
-                          </v-select>
-                        </v-col>
-                      </v-row>
+  <v-card flat id="share-structure">
+
+    <!-- Summary Section -->
+    <div id="share-summary">
+      <!-- Summary Header -->
+      <div class="share-summary-header" >
+        <v-icon>mdi-file-tree</v-icon>
```
severinbeauvais, line 8: Header icon colour should be #38598A.
Please fix any other instances that you find.

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -1,427 +1,731 @@
...
-    }
-    return []
+  // Share Class Functionality
+  /**
+   * Initialize the Add Share Class Form
+   */
+  private initNewShareClass (): void {
+    this.currentShareStructure = { ...this.newShareClass }
+    this.currentShareStructure.priority =
+      this.shareClasses.length === 0 ? 1 : this.shareClasses[this.shareClasses.length - 1].priority + 1
```
severinbeauvais, line 355: The conditional (ternary) operator is supposed to make simple code easier to read.
Maybe here an if-then-else would be easier to read.
Or parentheses around the if and else bits?

Ditto on line 358 below.

Another option is to separate this into 3 lines, eg:
```
val = condition
   ? if_truthy
   : if_falsy
```

src/mixins/filing-template-mixin.ts
severinbeauvais: I wonder if lodash `omit()` instead of `map()` would improve this code?

severinbeauvais (approved):
Wow, lots of work here 👍 

I know you ported much of code but I do have some comments to make it better.

Overall, looks good.

src/components/common/ActionChip.vue
cameron-freshworks: Updated in next commit

src/mixins/filing-template-mixin.ts
cameron-freshworks: Hmm maybe? As far as speed/performance though i'm not sure which approach would be best. 
Updated to enum in next commit 👍

src/mixins/filing-template-mixin.ts
cameron-freshworks: Great idea! Updated in next commit

src/views/Alteration.vue
cameron-freshworks: done and done, refactored so the ShareStructure component handles state in both Correction and Alteration.

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: Sure! Updated in next commit

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: thanks! Fixed in next commit

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: Good catch, thanks! fixed in next commit

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: Good point, will remember this 👍  Fixed!

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: I was wondering about this actually!

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: Will have to look at this a bit more

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: They have been relocated in next commit!

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: Yes, the attach was an error, good catch!
Fixed in next commit

src/components/ShareStructure/ShareStructure.vue
cameron-freshworks: done and done

src/components/ShareStructure/EditShareStructure.vue
severinbeauvais: Or you can put the text on the same line as the button element tags, like Vysakh did this morning.

src/components/ShareStructure/ShareStructure.vue
cameron-freshworks: Agreed, it was lengthy! Updated in commit

src/components/ShareStructure/EditShareStructure.vue
severinbeauvais: Yeah, just check the initial value in created(). If it's null/undefined then this code does nothing (and your watcher will handle its update).

cameron-freshworks: Oh i see. There is definitely a value there when we initiate an 'EDIT' of a share structure or class.

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: kk sounds good. I will do this when i finish updating the unit tests ( I'm thinking separate PR at this point 😁 )

src/views/Alteration.vue
severinbeauvais: Please delete obsolete getters and setters in Alteration.vue. Thx.

src/mixins/filing-template-mixin.ts
severinbeauvais: Thanks for the enum update.

Since you repurposed the previous orgs/person code block, could you please add a comment to the parties code block similar to the other "Filter out..." ones? Thx.

src/components/ShareStructure/EditShareStructure.vue
severinbeauvais: np

src/components/ShareStructure/EditShareStructure.vue
severinbeauvais: OK, I'll leave this with you.

severinbeauvais (approved):

src/mixins/filing-template-mixin.ts
cameron-freshworks: Definitely, will do!

src/views/Alteration.vue
cameron-freshworks: You bet, removed them from Correction, i will double check Alteration.

cameron-freshworks: Fixed in next commit!

PR #44, opened 2020-10-02
4850 Name translation component

vysakh-menon-aot:
*Issue #:* /bcgov/entity#4850

*Description of changes:*
 - Name translation component without 2nd level undo and label for (add, removed and modified)


src/components/YourCompany/NameTranslations/AddNameTranslation.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <div>
+    <confirm-dialog
+      ref="confirmTranslationDialog"
```
severinbeauvais, line 4: Please add an id to the top-level div for this component, and add an "attach" property to this dialog. This will attach the dialog to this component instead of the app level, ie its code will be in this component's DOM tree, which allows you to test the dialog (since elements at the app level are not available when you mount the component).

src/components/YourCompany/NameTranslations/AddNameTranslation.vue
```diff
@@ -0,0 +1,147 @@
...
+      </v-row>
+      <v-row>
+        <v-col>
+          <div class="action-btns">
+          <v-btn large color="primary"
+            id="name-translation-btn-ok"
+            :disabled="!nameTranslationForm"
+            @click="addTranslation"
+          >
+            OK
```
severinbeauvais, line 28: FYI, this renders (in the DOM) "OK" on a separate line, which makes it trickier to unit test because you can't select just the text so you can't use `toBe()` and instead have to use the less precise `contains()` to verify it.

This may not matter here, but that's why we either have the button text on the same line as the button element, or within its own span (which we can select during testing), allowing us to verify the text exactly.

src/components/YourCompany/NameTranslations/AddNameTranslation.vue
```diff
@@ -0,0 +1,147 @@
...
+          </v-btn>
+          <v-btn large outlined color="primary"
+            id="name-translation-btn-cancel"
+            @click="cancelNameTranslation"
```
severinbeauvais, line 32: I prefer to see the event handler with its (), to reinforce the fact that it's a method (and in this case, takes no parameters).

(I know this is a valid short-hand but I'm trying to make reviewing easier ;) )

src/components/YourCompany/NameTranslations/CorrectNameTranslation.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <name-translation
+        :nameTranslations="getNameTranslations"
```
severinbeauvais, line 3: Please index only 2 spaces.

src/components/YourCompany/NameTranslations/CorrectNameTranslation.vue
```diff
@@ -0,0 +1,52 @@
...
+})
+export default class CorrectNameTranslation extends Vue {
+  // Getters
+  @Getter getNameTranslations!: Array<string>
+  @Getter getOriginalIA!: IncorporationFilingIF
+
+  // Setters
+  @Action setNameTranslations!: ActionBindingIF
+
+  private get originalNameTranslations () {
```
severinbeauvais, line 37: Please add a return type to this getter.

This is how I did it:
```
private get originalOfficeAddresses (): IncorporationAddressIf {
  return (this.getOriginalIA.incorporationApplication.offices as IncorporationAddressIf)
}
```

src/components/YourCompany/NameTranslations/CorrectNameTranslation.vue
```diff
@@ -0,0 +1,52 @@
...
+    return this.getOriginalIA.incorporationApplication.nameTranslations?.new
+  }
+
+  private updateNameTranslations (nameTranslations: Array<string>): void {
```
severinbeauvais, line 41: Unused method.

src/components/YourCompany/NameTranslations/CorrectNameTranslation.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <name-translation
+        :nameTranslations="getNameTranslations"
+        :originalNameTranslations="originalNameTranslations"
+        @nameTranslationsChange="setNameTranslations"
```
severinbeauvais, line 5: In this case I would prefer to see
```
@nameTranslationsChange="setNameTranslations($event)"
```

@ Other Devs: If you strongly disagree with my preferred style for event handlers, let's discuss... I could change my mind 😀

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
+<template>
+  <div id="name-translation">
+    <confirm-dialog
+      ref="confirmTranslationDialog"
+      attach="#name-translation"
```
severinbeauvais, line 5: 👍

lekshmimallika-aot (commented):
Is building the filing data in template mixin part of another pull request?

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
...
+    />
+    <v-layout class="mt-3" v-if="!isEditing">
```
severinbeauvais, line 7: Typically, it's better if the parent component sets the top margin.

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
...
+      <v-flex md3>
```
severinbeauvais, line 8: Check your layout with different screen widths (even though we're not trying hard to be mobile-friendly)... in other places I changed from "md" to "xs" to get satisfactory resizing.

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
...
+        <div v-for="(name, index) in drafTranslations" :key="`name_translation_${index}`">{{name}}</div>
+      </v-flex>
+      <v-flex md7 v-else>
+        No name translations
+      </v-flex>
+      <v-flex md2 class="align-right" v-if="!hasNameTranslationChange">
+        <v-btn
+          text color="primary"
+          @click="isEditing = true"
+        >
```
severinbeauvais, line 26: I recommend adding an id to buttons that appear only once, or a class to buttons that can appear multiple times, for testing purposes!

src/components/YourCompany/NameTranslations/CorrectNameTranslation.vue
cameron-freshworks: If that event is receiving multiple props from the child emitter, it makes sense to do it like `setNameTranslations` ( cuz i don't know a better way to handle multiples :P ), but if it is a single prop being emit i think i also prefer to see the `$event`. 
TLDR: No objects to either approach 😁

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
...
+            <p>Name translations must use the Latin Alphabet (English, French, etc.).
+              Names that use other writing systems must spell the name phonetically in English or French.
+            </p>
+            <v-btn outlined color="primary" @click="isAddingNameTranslation=true" :disabled="isAddingNameTranslation">
+              <v-icon>mdi-plus</v-icon>
+              <span>Add Name Translation</span>
+            </v-btn>
+          </v-flex>
+        </v-layout>
+        <v-layout pb-5>
```
severinbeauvais, line 83: Does "pb-5" work here or does it need to be a class?

Also, I find it's better for lower elements to define a top margin (or padding) instead of for upper elements to define bottom whitespace. I find this makes layout easier since you never get unwanted whitespace below an element.

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
...
+
+  /** Add or update a name translation
+   *
+   * @param name The name to add
+   */
+  private addName (name: string): void {
+    // Handle name translation adds or updates
+    this.editIndex > -1
+      ? this.drafTranslations[this.editIndex] = name
+      : this.drafTranslations.push(name)
```
lekshmimallika-aot, line 227: Does this allow you to identify the new and changed translations separately? You would want  to know that in filing template mixin

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
...
+    this.isAddingNameTranslation = true
+  }
+
+  /** Remove a name translation
+   *
+   * @param index Index number of the name translation to remove
+   */
+  private removeNameTranslation (index: number): void {
+    this.drafTranslations.splice(index, 1)
+
```
lekshmimallika-aot, line 251: Same here. Will you be able to populate removed translation without another level of comparison in the template mixin?

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
...
+  private addName (name: string): void {
+    // Handle name translation adds or updates
+    this.editIndex > -1
+      ? this.drafTranslations[this.editIndex] = name
+      : this.drafTranslations.push(name)
+
+    // Emit haveChanges
+    // this.emitHaveChanges(this.hasNameTranslationChange)
+
+    this.cancelNameTranslation()
```
severinbeauvais, line 232: This code doesn't make sense to me (and you have some commented out code)... Why are you canceling the name translation during add?

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,314 @@
...
+        yes: 'Save changes',
+        no: null,
+        cancel: 'Cancel without saving'
+      }
+    ).then(async (confirm) => {
+      if (confirm) {
+        this.setNameTranslations()
+      }
+    }).catch(() => {
+      this.onNameTranslationsPropValueChanged()
```
severinbeauvais, line 214: Would it make more sense to actually update the prop value here instead of calling the watch handler?

severinbeauvais (approved):
Great looking code! Just a few comments but none blocking.

vysakh-menon-aot:
> Is building the filing data in template mixin part of another pull request?

It was already present in the code

lekshmimallika-aot:
> > Is building the filing data in template mixin part of another pull request?
> 
> It was already present in the code

 nameTranslations: {
          new: this.stateModel.nameTranslations
        },

is what I see in the template mixin. For correction you will not have just new entries but modified and ceased entries as well.

example: 'nameTranslations': {
        'new': ['MÉDIAS DE TRANSPORT INC.'],
        'modified': [{
            'oldValue': 'A1 LTD.',
            'newValue': 'SOCIÉTÉ GÉNÉRALE'
        }],
        'ceased': ['B1', 'B2']
    }

src/components/YourCompany/NameTranslations/CorrectNameTranslation.vue
severinbeauvais: @cameron-freshworks https://www.google.com/search?q=vue+emit+multiple+parameters

src/components/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: Is this true? I think we just send the entire updated IA to the API, with no special flags for new or changed.

src/components/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: Ah yes, normally you would keep track of which translations were deleted and then filter them out in the Filing Template Mixin.
But here, you just delete the translation (no Undo), right?

severinbeauvais:
> example: 'nameTranslations': {
> 'new': ['MÉDIAS DE TRANSPORT INC.'],
> 'modified': [{
> 'oldValue': 'A1 LTD.',
> 'newValue': 'SOCIÉTÉ GÉNÉRALE'
> }],
> 'ceased': ['B1', 'B2']
> }

I think you are confusing `filing.incorporationApplication` (the full updated/corrected IA) and `filing.correction.diff` (which the BE constructs).

The UI doses not build the diff section.

src/components/YourCompany/NameTranslations/NameTranslation.vue
vysakh-menon-aot: Identifying Add, Modified and Removed are not part of this ticket

lekshmimallika-aot:
> > example: 'nameTranslations': {
> > 'new': ['MÉDIAS DE TRANSPORT INC.'],
> > 'modified': [{
> > 'oldValue': 'A1 LTD.',
> > 'newValue': 'SOCIÉTÉ GÉNÉRALE'
> > }],
> > 'ceased': ['B1', 'B2']
> > }
> 
> I think you are confusing `filing.incorporationApplication` (the full updated/corrected IA) and `filing.correction.diff` (which the BE constructs).
> 
> The UI doses not build the diff section.

No. Very clear on what I said. There is a reason why this cant be all data. This is a section that is applicable for both alteration and correction unlike people and roles or office address. Even though share structure is common like this alteration is supposed to file the whole share structure( not just the changes). So all of these can be deleted and recreated.

For name translation, alteration gives the data in the mode described above and the same processor processes both. We cannot delete and recreate name translations as it is done for other components. I had explained this to Vysakh

Please see the sample data of alteration in the schemas. Correction was not updated to include sample modified and ceased entries.

severinbeauvais:
> For name translation, alteration gives the data in the mode described above and the same processor processes both. We cannot delete and recreate name translations as it is done for other components. I had explained this to Vysakh
> 
> Please see the sample data of alteration in the schemas. Correction was not updated to include sample modified and ceased entries.

Oh! You are correct (sorry!). The Name Translations schema has other sections for "modified" and "ceased".

https://github.com/bcgov/business-schemas/blob/master/src/registry_schemas/schemas/name_translations.json

It looks like we need to keep track of corrected/added/ceased name translations. This is a good bit more work than what is currently implemented, so this needs to be brought up to our POs for approval.

src/components/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: OK, thanks.

Please also update the buttons on lines 105 and 111.

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,310 @@
...
+            <list-name-translation
+              v-if="drafTranslations && drafTranslations.length > 0"
+              :isAddingNameTranslation="isAddingNameTranslation"
+              :translationList="drafTranslations"
+              @editNameTranslation="editNameTranslation($event)"
+              @removeNameTranslation="removeNameTranslation($event)"
+            />
+          </v-flex>
+        </v-layout>
+        <v-layout pt-5>
```
severinbeauvais, line 102: Does this prop actually work, or should this be a class value?

src/components/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: ^^^ Actually, Name Translations are not done this way. See the schema. There is a filing section for new, modified and ceased.

(I don't know how unchanged translations are stored...?)

src/components/YourCompany/NameTranslations/NameTranslation.vue
vysakh-menon-aot: Yes it does

severinbeauvais: OK. I just learned something!

src/components/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: I think this conversation is superceded by Lekshmi's (and my) comments below.

src/components/YourCompany/NameTranslations/ListNameTranslation.vue
```diff
@@ -0,0 +1,210 @@
...
+        <v-chip v-if="translation.action === actionTypes.ADDED"
+          x-small label color="#1669BB" text-color="white">ADDED</v-chip>
+        <v-chip v-if="translation.action === actionTypes.EDITED"
+          x-small label color="#1669BB" text-color="white">CORRECTED</v-chip>
+        <v-chip v-if="translation.action === actionTypes.REMOVED"
+          x-small label color="#E0E0E0" text-color="grey darken-1">REMOVED</v-chip>
+        </v-col>
+
+        <!-- Actions Column -->
+        <v-col v-if="translation.action === actionTypes.EDITED || translation.action === actionTypes.REMOVED">
```
severinbeauvais, line 29: What about ADDED? Is that handled via the else below (Edit button)?

BTW, check with Scott as I think he wants the dropdown edit button to be labeled "Correct".

src/components/YourCompany/NameTranslations/ListNameTranslation.vue
```diff
@@ -0,0 +1,210 @@
...
+
+@Component({})
+export default class ListNameTranslation extends Vue {
+  @Prop({ default: () => [] })
+  private translationList: NameTranslationDraftIF[]
+
+  @Prop({ default: false })
+  private isAddingNameTranslation: boolean
+
+  private actionTypes = ActionTypes
```
severinbeauvais, line 135: I recommend you capitalize the left side, as it's still a "type". It just needs to be in the script so the Vue template can see it.

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,360 @@
...
+      <v-flex xs3>
+        <label><strong>Name Translation(s)</strong></label>
+        <v-flex md1>
+            <v-chip v-if="hasNameTranslationChange" x-small label color="#1669BB" text-color="white">
+                Corrected
+            </v-chip>
+          </v-flex>
+      </v-flex>
+      <v-flex xs7 v-if="draftTranslations && draftTranslations.filter(x => x.action !== 'removed').length">
+        <div v-for="(translation, index) in draftTranslations.filter(x => x.action !== 'removed')"
```
severinbeauvais, line 17: The filter code here and in the line above would be better as a getter due to caching.

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,360 @@
...
+  private setNameTranslations (): void {
+    this.emitNameTranslations(this.draftTranslations)
+    this.emitHaveChanges(this.hasNameTranslationChange)
+    this.isEditing = false
+  }
+
+  private resetNameTranslations (): void {
+    this.draftTranslations = this.nameTranslations
+      .filter(x => x.action !== 'added')
+      .map(a => {
```
severinbeauvais, line 194: Nice use of 'map' :)

severinbeauvais: So it looks like you don't undo added names? Is this Scott's design?

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,360 @@
...
+    this.emitHaveChanges(false)
+    this.isEditing = false
+  }
+
+  private cancelNameTranslationCorrection () {
+    const nameTranslations = this.nameTranslations || []
+    const hasUnsavedData = this.draftTranslations.length !== nameTranslations.length ||
+      !this.draftTranslations.every((translation, index) => {
+        return nameTranslations[index].value === translation.value &&
+          nameTranslations[index].action === translation.action
```
severinbeauvais, line 211: I had to look at this code for a couple of minutes to understand it, haha.
Could you please add a comment saying what you're looking for to determine if there is unsaved data?

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,360 @@
...
+        yes: 'Save changes',
+        no: null,
+        cancel: 'Cancel without saving'
+      }
+    ).then(async (confirm) => {
+      if (confirm) {
+        this.setNameTranslations()
+      }
+    }).catch(() => {
+      this.draftTranslations = this.nameTranslations ? this.nameTranslations.map(a => Object.assign({}, a)) : []
```
severinbeauvais, line 237: Yep, good catch -- your previous `[...x]` code only did a shallow copy of the array. THIS code returns an array with new items (but remember that assign() also does a shallow copy so you have to watch that your items don't have sub-objects).
FWIW, an alternative is Lodash cloneDeep(). Up to you.

src/components/YourCompany/YourCompany.vue
```diff
@@ -63,37 +63,9 @@
...
-              id="btn-correct-name-translations"
-              @click="nameTranslationChanges = true"
-            >
-              <v-icon small>mdi-pencil</v-icon>
-              <span>Correct</span>
-            </v-btn>
-          </div>
-        </v-flex>
-      </v-layout>
+      <correct-name-translation class="mt-3"
```
severinbeauvais, line 66: Nice cleanup / new sub-component :)

src/mixins/filing-template-mixin.ts
```diff
@@ -81,6 +84,25 @@ export default class FilingTemplateMixin extends Vue {
         shareClasses[index].series = share.series?.filter(x => x.action !== ActionTypes.REMOVED)
           .map((x) => { const { action, ...rest } = x; return rest })
       }
+
+      // Filter out and modify name translation
+      const translations = this.stateModel.nameTranslations as NameTranslationDraftIF[]
```
severinbeauvais, line 89: This code looks fine but please break it out into a new function. THIS method should do 1 thing but it has grown :)

(Maybe the other filters above should be functioned out too.)

src/mixins/filing-template-mixin.ts
```diff
@@ -171,7 +191,19 @@ export default class FilingTemplateMixin extends Vue {
     this.setNameRequest(filing.incorporationApplication.nameRequest)
 
     // Set Name Translations
-    this.setNameTranslations(filing.incorporationApplication.nameTranslations?.new)
+    if (filing.incorporationApplication.nameTranslations instanceof Array) {
```
severinbeauvais, line 194: If I understand correctly, this is to handle a couple of possible structures here. Please add a comment explaining what you expect to see here (your assumption).

Is there any way to NOT have this situation? Is it specifically to handle a real IA vs a draft? If feels complicated.

tests/unit/ListNameTranslations.spec.ts
```diff
@@ -12,8 +12,9 @@ import { getVuexStore } from '@/store'
 import { createLocalVue, mount } from '@vue/test-utils'
 
 // Components
-import { ListNameTranslations } from '@/components/YourCompany'
+import { ListNameTranslation } from '@/components/YourCompany'
```
severinbeauvais, line 15: Let's all keep in mind that it would be nice to have a "Correction.spec.ts" file where we can test the "resume draft" functionality and verify that all data is read in as expected.

I'm not asking you to do it in this ticket. It should probably be a new (tech debt) ticket.

severinbeauvais (approved):
Looks OK to me.

src/components/YourCompany/NameTranslations/NameTranslation.vue
vysakh-menon-aot: Its not specified in design

src/components/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: Great comment, thanks!

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,369 @@
...
+        yes: 'Save changes',
+        no: null,
+        cancel: 'Cancel without saving'
+      }
+    ).then(async (confirm) => {
+      if (confirm) {
+        this.setNameTranslations()
+      }
+    }).catch(() => {
+      this.draftTranslations = this.nameTranslations ? this.nameTranslations.map(a => cloneDeep(a)) : []
```
severinbeauvais, line 246: cloneDeep is recursive so you can just do
```
this.draftTranslations = this.nameTranslations ? cloneDeep(this.nameTranslations) : []
```

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -0,0 +1,369 @@
...
+
+  private setNameTranslations (): void {
+    this.emitNameTranslations(this.draftTranslations)
+    this.emitHaveChanges(this.hasNameTranslationChange)
+    this.isEditing = false
+  }
+
+  private resetNameTranslations (): void {
+    this.draftTranslations = this.nameTranslations
+      .filter(x => x.action !== 'added')
```
severinbeauvais, line 199: Should use enum here.

src/mixins/filing-template-mixin.ts
```diff
@@ -159,6 +163,29 @@ export default class FilingTemplateMixin extends Vue {
...
   }
 
+  /**
+   * Prepare name translations for non draft correction
+   */
+  prepareNameTranslations () : NameTranslationIF {
+    const translations = this.stateModel.nameTranslations as NameTranslationDraftIF[]
+    return {
+      new: translations
+        .filter(x => x.action === 'added')
```
severinbeauvais, line 173: Should use enum here.

Please also change all 'edited' to enum.

src/components/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: Also on line 323.

src/mixins/filing-template-mixin.ts
```diff
@@ -171,7 +198,21 @@ export default class FilingTemplateMixin extends Vue {
     this.setNameRequest(filing.incorporationApplication.nameRequest)
 
     // Set Name Translations
-    this.setNameTranslations(filing.incorporationApplication.nameTranslations?.new)
+    if (filing.incorporationApplication.nameTranslations instanceof Array) {
+      // If it's an array that means it's a draft which is saved from edit-ui by staff.
+      this.setNameTranslations(filing.incorporationApplication.nameTranslations)
+    } else {
+      // If it's an object that means it's an initial draft created from filing-ui and has a structure of an IA.
```
severinbeauvais, line 205: Thanks for the helpful comments but now I'm confused... Why do we have 2 different possibilities here? And if there are unchanged translations, what do they look like?

Translations may be a special case but it was my understanding that a corrected IA is a complete IA object in a correction body.

vysakh-menon-aot: All unchanged translations will still be in array with action as null (draft).

vysakh-menon-aot: NameTranslationDraftIF[] is used only for draft state. NameTranslationIF is aligned to the schema of IA

severinbeauvais: Can you please post an example of the IA structure with: some new/modified/ceased translations, and some unchanged ones?

vysakh-menon-aot: When staff initiate a correction: `'nameTranslations': {'new': ['ABC Ltd.', 'Société Générale']}`
Then edit ui converts it into` [{'value':'ABC Ltd.', 'oldValue': null, 'action': null},  {'value': 'Société Générale', 'oldValue': null, 'action': null}]` here to keep track of changes in draft mode.
Then while filing this array will be converted to `{'new': [], 'modified': {'newValue': 'XYZ Ltd.', 'oldValue': 'ABC Ltd.'}, 'ceased': ['Société Générale']}`
If staff add any translation it will be in the `new` array

severinbeauvais: Can you please post an example of the IA structure with: some new/modified/ceased translations, and some unchanged ones?

vysakh-menon-aot: `{'new': ['i am new'], 'modified': {'newValue': 'XYZ Ltd.', 'oldValue': 'ABC Ltd.'}, 'ceased': ['Société Générale']}`
unchanged once will not be part of correction filing, hence stripped out while filing

severinbeauvais: OK.
So I think this goes against the idea that a correction includes a COMPLETE IA filing structure.
@lekshmimallika-aot Does this seem right to you? Does the back end retain the unchanged translations?

lekshmimallika-aot: This looks ok to me. Yes backend retain the unchanged translations

lekshmimallika-aot (approved):

src/mixins/filing-template-mixin.ts
severinbeauvais: OK for now

tests/unit/ListNameTranslations.spec.ts
severinbeauvais: New tech debt ticket for Correction view / save and resume:
https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/5213

src/components/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: I think this conversation is superseded by other comments below.

PR #46, opened 2020-10-05
Update state-getters.ts

severinbeauvais:
*Issue #:* None

*Description of changes:*
- Temp fix (until certify component is implemented) to allow File and Pay when other components are valid.


lekshmimallika-aot (approved):

cameron-freshworks (approved):
Looks good. Is the `detail.valid` check also ready to go?

severinbeauvais:
> Looks good. Is the `detail.valid` check also ready to go?

I think. Its value is true so it's not blocking the File button.

cameron-freshworks:
> > Looks good. Is the `detail.valid` check also ready to go?
> 
> I think. Its value is true so it's not blocking the File button.

Copy that, lets get it in then!

PR #49, opened 2020-10-05
Name Request Correction Form Validators  Update

cameron-freshworks:
*Issue #:* /bcgov/entity#5082

*Description of changes:*

* Simplified and  updated the form validations for phone and email on name request search. 

### DIFF
--- a/src/components/YourCompany/CompanyName/CorrectNameRequest.vue
+++ b/src/components/YourCompany/CompanyName/CorrectNameRequest.vue
@@ -94,12 +94,12 @@ export default class CorrectNameRequest extends Mixins(NameRequestMixin) {
     v => this.validateNameRequestNumber(v) || 'Name Request Number is invalid'
   ]
   private entityPhoneNumberRules = [
-    v => this.isInputEntered(v, 'phone') || 'Phone number is required',
+    v => (v !== '' || !this.entityEmail) || 'Phone number is required',
     v => !(v?.length > 12) || 'Phone number is invalid'
   ]
   private entityEmailRules = [
-    v => this.isInputEntered(v, 'email') || 'Email is required',
-    v => this.isValidateEmail(v) || 'Email is Invalid'
+    v => (v !== '' || !this.entityPhone) || 'Email is required',
+    v => (v === '' || this.isValidateEmail(v)) || 'Email is Invalid'
   ]
 
   mounted (): void {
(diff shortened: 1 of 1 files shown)


src/components/YourCompany/CompanyName/CorrectNameRequest.vue
```diff
@@ -94,12 +94,12 @@ export default class CorrectNameRequest extends Mixins(NameRequestMixin) {
     v => this.validateNameRequestNumber(v) || 'Name Request Number is invalid'
   ]
   private entityPhoneNumberRules = [
-    v => this.isInputEntered(v, 'phone') || 'Phone number is required',
+    v => (v !== '' || !this.entityEmail) || 'Phone number is required',
```
severinbeauvais, line 97: This looks backwards... please add a comment that it's one or the other, or something :)

severinbeauvais (approved):
Okey doke.

vysakh-menon-aot (approved):

src/components/YourCompany/CompanyName/CorrectNameRequest.vue
cameron-freshworks: oh no your right, i can do this better. Will refactor!

lekshmimallika-aot (approved):

src/components/YourCompany/CompanyName/CorrectNameRequest.vue
cameron-freshworks: Ok, fixed my wonky rules. They didn't make any sense. Refactored and Simplified :P

severinbeauvais: Does the new code still enforce "either phone number or email"?

severinbeauvais: And do you have a unit test for this, so we find out right away if we break it in the future?

cameron-freshworks: Yes, the form itself will never be valid unless one or the other are present.
No, will add one!

cameron-freshworks: I can explain as well, this fix was more of a UX fix as the core of the validation wasn't broken prior. 

There is two layers of validation, the form itself and the input fields individually. Prior, the input fields were validating as if they were required, which technically neither is required, but one or the other needs to be there to complete ( validating at the form level ). So i removed the 'required rule' at the input level, and just relied on the form validation to check at least one of the values is present.

PR #57, opened 2020-10-07
Default Detail Comment Fix

cameron-freshworks:
*Issue #:* /bcgov/entity#5203

*Description of changes:*

* Updated Getters and Detail comment component.

* Updated filing template mixin to handle adding and parsing default detail comment.

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -41,6 +41,7 @@ export default class FilingTemplateMixin extends Vue {
   @Getter getFolioNumber!: string
   @Getter getStaffPayment!: StaffPaymentIF
   @Getter getDetailComment!: string | null
+  @Getter getDefaultCorrectionDetailComment!: string
 
   // Global setters
   @Action setBusinessContact!: ActionBindingIF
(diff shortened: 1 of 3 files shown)


src/store/getters/state-getters.ts
```diff
@@ -254,8 +254,9 @@ export const isFilingValid = (state: any): boolean => {
     state.stateModel.staffPaymentStep.valid)
 }
 
-export const getOriginalIAFilingDate = (state: any): boolean => {
-  return state.stateModel.originalIA.header.date.split('T')[0]
+export const getDefaultCorrectionDetailComment = (state: any): string => {
+  const filingDate = state.stateModel.originalIA.header.date.split('T')[0]
```
severinbeauvais, line 258: Should you have a fallback value if original IA has not been loaded yet? (I know the object exists in the state model but there date=''.)

src/mixins/filing-template-mixin.ts
```diff
@@ -251,7 +252,9 @@ export default class FilingTemplateMixin extends Vue {
       certifiedBy: filing.header.certifiedBy
     })
 
-    this.setDetailComment(filing.correction.comment)
+    const comment: string = filing.correction.comment || ''
+    const detailComment = comment.split('\n').slice(1).join('\n')
```
severinbeauvais, line 256: This is complex enough that I think it needs a comment.
Might as well add a comment to line 110 to say what you're doing there too :)

severinbeauvais (approved):

vysakh-menon-aot (approved):

src/mixins/filing-template-mixin.ts
cameron-freshworks: Absolutely, will do!

src/store/getters/state-getters.ts
cameron-freshworks: Great call!

PR #63, opened 2020-10-08
5207 Fee Module Bug ( Various State Fixes )

cameron-freshworks:
*Issue #:* /bcgov/entity#5207

*Description of changes:*

* Implements watchers in Contact info and People and Roles to handle 'changed' states when resuming from drafts.

* Bug fix in Share Structure that wouldn't restore series shares after they had been shuffled around in the list placement.

*NOTE:* 

* Developer question...   want me to incorporate this tickets work ( Disabling Actions) into this PR, should only add a few hrs.. 

https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/5115

### DIFF
--- a/src/components/PeopleAndRoles/PeopleAndRoles.vue
+++ b/src/components/PeopleAndRoles/PeopleAndRoles.vue
@@ -439,6 +439,14 @@ export default class PeopleAndRoles extends Vue {
   private onPeopleAndRoles (): void {
     this.setPeopleAndRolesValid(this.hasValidRoles)
   }
+
+  /**
+   * Sets the 'changes' flag when a user has made changes.
+   */
+  @Watch('hasChanges')
+  private onPeopleAndRolesChanges (): void {
+    this.setPeopleAndRolesChanged(this.hasChanges)
+  }
 }
 </script>
 
--- a/src/components/YourCompany/BusinessContactInfo.vue
+++ b/src/components/YourCompany/BusinessContactInfo.vue
@@ -247,8 +247,6 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
     this.emitHaveChanges(this.hasBusinessContactInfoChange)
   }
 
-  @Watch('')
-
   @Emit('contactInfoChange')
   private emitContactInfo (contactInfo: BusinessContactIF): void { }
 
--- a/src/components/YourCompany/BusinessContactInfo.vue
+++ b/src/components/YourCompany/BusinessContactInfo.vue
@@ -164,15 +164,14 @@
 
 <script lang="ts">
 // Libraries
-import { Component, Vue, Prop, Watch, Emit, Mixins } from 'vue-property-decorator'
+import { Component, Prop, Watch, Emit, Mixins } from 'vue-property-decorator'
 import { mask } from 'vue-the-mask'
 
 // Interfaces
-import { ActionBindingIF, BusinessContactIF, IncorporationFilingIF } from '@/interfaces'
+import { BusinessContactIF } from '@/interfaces'
 
 // Mixins
 import { CommonMixin } from '@/mixins'
-import { Action, Getter } from 'vuex-class'
 
 @Component({
   directives: { mask }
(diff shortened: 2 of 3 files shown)


severinbeauvais:
> * Developer question...   want me to incorporate this tickets work ( Disabling Actions) into this PR, should only add a few hrs..

No, that's targeted for next sprint and we have enough work in this sprint :D

src/components/YourCompany/BusinessContactInfo.vue
```diff
@@ -247,6 +240,13 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
     return this.businessContact.email !== this.originalBusinessContact.email
   }
 
+  // Watchers
+  @Watch('businessContact', { immediate: true })
```
severinbeauvais, line 244: If you don't watch-deep then this method will not be called if a sub-property changes; it will only be called if the main handle changes.

src/components/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -439,6 +439,14 @@ export default class PeopleAndRoles extends Vue {
   private onPeopleAndRoles (): void {
     this.setPeopleAndRolesValid(this.hasValidRoles)
   }
+
+  /**
+   * Sets the 'changes' flag when a user has made changes.
+   */
+  @Watch('hasChanges')
+  private onPeopleAndRolesChanges (): void {
+    this.setPeopleAndRolesChanged(this.hasChanges)
```
severinbeauvais, line 448: On the one hand, this will store the changed flag so we know that we can enable some actions.

On the other hand, this (and other components like it) are not actually identifying that changes were made since the last save -- the idea being that we could prompt the user for unsaved changes. Now if they resume a draft with changes, they will get prompted. Solving this will be tricky. Maybe we can live with this for now.

src/components/YourCompany/BusinessContactInfo.vue
cameron-freshworks: Interesting, what i found was with `deep` is it was reflecting the change in the Fee Summary as soon as you edited the field, and by removing it it would update once the form was 'done' editing etc

severinbeauvais: That's weird, since Fee Summary (or its data) isn't part of the businessContact object. Maybe you just saw a timing issue. Either way, I think you need "deep" or this won't recognize changes to businessContact fields.

cameron-freshworks: The fee summary is actioned by the Getter that looks at the `changed` prop in the various areas of correction. In this case, changing anything in the `YourCompany` will reflect in the `defineCompanyStep.changed` becoming true. 

Can confirm, with `deep` it reflects in the Fee summary as soon as any edits happen within the text field. Without, the changes in the fee summary reflect when selecting 'done'.

Will look into it more, for a better explanation!

cameron-freshworks: So after another look..
Inside the watcher we were creating a direct reference from the model to the prop, which is why `deep: true` was causing the Prop to change whenever the model changed. In turn causing the watcher to call itself often, where we only wanted it to be reactive to the store values, ie `businessContact`. 
So, put `deep: true` back in, but used a shallow copy of the prop to initialize the `contact` data opposed to a reference to the prop itself.

lekshmimallika-aot: Yes after the initial commit it was a direct reference to the prop. But I had fixed this in PR https://github.com/bcgov/business-edit-ui/pull/36/files and fee summary was working correctly for business contact save and resume. Looks like it got overwritten by some other PR

lekshmimallika-aot (approved):

src/components/YourCompany/BusinessContactInfo.vue
severinbeauvais: Thanks for the fix!

severinbeauvais (approved):

PR #67, opened 2020-10-09
Misc Bug fixes

cameron-freshworks:
*Issue #:* /bcgov/entity#5202

*Description of changes:*

* Quick fix for a getter bug that wasn't evaluating beyond the first Share Class
* Updated Unit testing for above

* Added verbiage for Share Structure

* Updated logic to only validate the NR at file and pay if changes were made to the NR.

* Fixed unwanted population of NR data in NR Change form

### DIFF
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -68,6 +68,7 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
   @Getter getEffectiveDate!: Date
   @Getter isFilingChanged!: boolean
   @Getter isFilingValid!: boolean
+  @Getter hasNewNr!: boolean
 
   // Global setters
   @Action setIsSaving!: ActionBindingIF
@@ -140,6 +141,19 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
 
     this.setIsFilingPaying(true)
     let filingComplete
+
+    /** If it is a named company IA, validate NR before filing submission. This method is different
+     * from the processNameRequest method in App.vue. This method shows a generic message if
+     * the Name Request is not valid and clicking ok in the pop up redirects to the Manage Businesses
+     * dashboard */
+    if (this.isNamedBusiness && this.hasNewNr) {
+      try {
+        await this.validateNameRequest(this.getNameRequestNumber)
+      } catch (error) {
+        this.setIsFilingPaying(false)
+        return
+      }
+    }
 
     try {
       const filing = await this.buildIaCorrectionFiling(false)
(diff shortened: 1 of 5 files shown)


severinbeauvais (approved):

severinbeauvais:
PS Too bad there wasn't a unit test for this :P

cameron-freshworks:
> PS Too bad there wasn't a unit test for this :P

haha i can do that!

vysakh-menon-aot (approved):

src/components/common/Actions.vue
```diff
@@ -141,19 +141,6 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
     this.setIsFilingPaying(true)
     let filingComplete
 
-    /** If it is a named company IA, validate NR before filing submission. This method is different
-     * from the processNameRequest method in App.vue. This method shows a generic message if
-     * the Name Request is not valid and clicking ok in the pop up redirects to the Manage Businesses
-     * dashboard */
-    if (this.isNamedBusiness) {
-      try {
-        await this.validateNameRequest(this.getNameRequestNumber)
```
severinbeauvais, line 150: Don't we still need this if they are using a new NR?

cameron-freshworks: This will validate during that process of adding a new NR. Do you think they would need to Re-validate again when filing and paying?

severinbeauvais: Not 100% sure.

On the one hand, I've heard that corrections are for staff only (for now) and will be done in one shot (no save and resume), so this reduces the chance of the NR no longer being valid during the session.

On the other hand, we do support save and resume, so it's possible that the user (staff or other) tries to file and pay on a different day than when they added the new NR, so this check could prevent data errors in that case.

Maybe this should be asked of Scott and/or BAs/POs...

cameron-freshworks: Ok, sounds good! I will inquire on this, as you mentioned above i feel like this is low enough risk to leave out at the moment.  
If it wasn't blocking Named Company filings that didn't include NR changes, i would 100% leave it in but in the meantime, it looks like a check we can do without at the moment.

vysakh-menon-aot (approved):

src/store/getters/state-getters.ts
```diff
@@ -256,7 +265,7 @@ export const isFilingValid = (state: any): boolean => {
 
 export const getDefaultCorrectionDetailComment = (state: any): string => {
   const filingDate = state.stateModel.originalIA.header.date.split('T')[0] || ''
-  return `Correction for Incorporation Application. Filed on ${filingDate}`
+  return `Correction for Incorporation Application filed on ${filingDate}`
```
severinbeauvais, line 268: I know this wasn't specced, but what do you think about a period at the end?

cameron-freshworks: Updated!

PR #71, opened 2020-10-15
EntityInfo unit tests and updates

cameron-freshworks:
*Issue #:* /bcgov/entity#5074

*Description of changes:*

* Fixed and updated unit tests for entityInfo component for edit UI.

* Updated some id's in the entityInfo component to make sense with the edit ui.

### DIFF
--- a/tests/unit/entityInfo.spec.ts
+++ b/tests/unit/entityInfo.spec.ts
@@ -6,43 +6,76 @@ import Vuetify from 'vuetify'
 import { getVuexStore } from '@/store'
 
 // Components
-import { mount } from '@vue/test-utils'
+import { createLocalVue, mount } from '@vue/test-utils'
 import { EntityInfo } from '@/components/common'
 
+// Other
+import mockRouter from './MockRouter'
+import VueRouter from 'vue-router'
+
 Vue.use(Vuetify)
-
 const vuetify = new Vuetify({})
+document.body.setAttribute('data-app', 'true')
 
-describe.skip('Entity Info component with an NR', () => {
-  let wrapper: any
-  let store: any
+describe('Entity Info component in a Correction as a named company', () => {
+  let wrapper
+  let store
+
+  const mockFiling = {
+    header: {
+      name: 'incorporationApplication',
+      filingId: 12345
+    },
+    business: {
+      identifier: 'BC1234567',
+      legalType: 'BEN'
+    },
+    incorporationApplication: {
+      contactPoint: {
+        email: 'mock@email.com',
+        phone: '123-456-7890'
+      },
+      nameRequest: {
+        legalType: 'BEN',
+        nrNumber: 'NR 1234567',
+        legalName: 'My Mock Name Inc.'
+      },
+      offices: {},
+      parties: [],
+      shareClasses: [],
+      incorporationAgreement: {}
+    }
+  }
 
   beforeEach(() => {
     store = getVuexStore()
+    let state = store.state.stateModel
 
-    // Entity type will always be set with or without an NR
-    store.state.stateModel.tombstone.entityType = 'BEN'
-    // Temp Id will always be set with or without an NR
-    store.state.stateModel.tombstone.businessId = 'T1234567'
-    store.state.stateModel.nameRequest.nrNumber = 'NR 1234567'
-    store.state.stateModel.nameRequest.details.approvedName = 'Xyz Ltd.'
+    state.tombstone.keycloakRoles = ['staff']
+    state.businessInformation = mockFiling.business
+    state.tombstone.businessId = mockFiling.business.identifier
+    state.defineCompanyStep.businessContact = mockFiling.incorporationApplication.contactPoint
+    state.originalIA.incorporationApplication.nameRequest = mockFiling.incorporationApplication.nameRequest
 
-    wrapper = mount(EntityInfo, { vuetify, store })
+    const localVue = createLocalVue()
+    localVue.use(VueRouter)
+    const router = mockRouter.mock()
+    router.push({ name: 'correction' })
+
+    wrapper = mount(EntityInfo, {
+      localVue,
+      vuetify,
+      store,
+      router
+    })
   })
 
-  afterEach(() => {
-    wrapper.destroy()
+  afterEach(async () => {
+    await wrapper.destroy()
   })
 
-  it('renders the Name Request header when the EntityType(BC) is present', async () => {
-    expect(wrapper.vm.$el.querySelector('#nr-header').textContent)
-      .toContain('BC Benefit Company Incorporation Application')
-
-    expect(wrapper.vm.$el.querySelector('#entity-legal-name').textContent)
-      .toContain('Xyz Ltd.')
-
-    expect(wrapper.vm.$el.querySelector('#entity-nr-number').textContent)
-      .toContain('NR 1234567')
+  it('renders the EntityInfo Component', async () => {
+    expect(wrapper.find(EntityInfo).exists()).toBe(true)
   })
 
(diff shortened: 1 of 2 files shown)


vysakh-menon-aot (approved):

tests/unit/entityInfo.spec.ts
```diff
@@ -6,43 +6,76 @@ import Vuetify from 'vuetify'
...
+      vuetify,
+      store,
+      router
+    })
   })
 
-  afterEach(() => {
-    wrapper.destroy()
+  afterEach(async () => {
+    await wrapper.destroy()
```
severinbeauvais, line 74: My hover help (from Vetur extension I think) informs me that "await has no effect on the type of this expression". In other words, "destroy" is not async (does not return a promise).

If you need to wait for things to stability here, use `await Vue.nextTick()`, otherwise please remove this code.

tests/unit/entityInfo.spec.ts
```diff
@@ -6,43 +6,76 @@ import Vuetify from 'vuetify'
...
-  it('renders the Name Request header when the EntityType(BC) is present', async () => {
-    expect(wrapper.vm.$el.querySelector('#nr-header').textContent)
-      .toContain('BC Benefit Company Incorporation Application')
-
-    expect(wrapper.vm.$el.querySelector('#entity-legal-name').textContent)
-      .toContain('Xyz Ltd.')
-
-    expect(wrapper.vm.$el.querySelector('#entity-nr-number').textContent)
-      .toContain('NR 1234567')
+  it('renders the EntityInfo Component', async () => {
```
severinbeauvais, line 77: Not a big deal (just cleanup) but the arrow function within is not async.

Ditto for others in this file.

tests/unit/entityInfo.spec.ts
```diff
@@ -53,58 +86,119 @@ describe.skip('Entity Info component with an NR', () => {
...
     const crumb3 = breadcrumbs.at(4)
 
-    // TODO: add test for "Staff Dashboard" when user is staff
-    expect(crumb1.text()).toStrictEqual('Manage Businesses Dashboard')
+    expect(crumb1.text()).toStrictEqual('Staff Dashboard')
     expect(divider.text()).toStrictEqual('>')
-    expect(crumb2.text()).toStrictEqual('Xyz Ltd.')
-    expect(crumb3.text()).toStrictEqual('BC Benefit Company Incorporation Application')
+    expect(crumb2.text()).toStrictEqual('My Mock Name Inc.')
+    expect(crumb3.text()).toStrictEqual('Correction - Incorporation Application')
```
severinbeauvais, line 92: I think you only need to use `toStrictEqual` for objects/arrays. Otherwise you can just use `toBe`.

tests/unit/entityInfo.spec.ts
```diff
@@ -53,58 +86,119 @@ describe.skip('Entity Info component with an NR', () => {
...
-})
 
-describe.skip('Entity Info component without an NR', () => {
-  let wrapper: any
-  let store: any
-
-  beforeEach(() => {
-    store = getVuexStore()
+  it('renders the business name and numbers', async () => {
+    expect(wrapper.vm.$el.querySelector('#entity-legal-name').textContent)
```
severinbeauvais, line 96: I would prefer to see us move to the new wrapper format here, eg,
`wrapper.find('#entity-legal-name').text()`.

(Otherwise we're reaching in to the element and calling the JS query function directly.)

tests/unit/entityInfo.spec.ts
```diff
@@ -6,43 +6,76 @@ import Vuetify from 'vuetify'
...
+      },
+      offices: {},
+      parties: [],
+      shareClasses: [],
+      incorporationAgreement: {}
+    }
+  }
 
   beforeEach(() => {
     store = getVuexStore()
```
severinbeauvais, line 51: `store` only needs to be called once, which you can do above, near where you get `vuetify`.

tests/unit/entityInfo.spec.ts
```diff
@@ -6,43 +6,76 @@ import Vuetify from 'vuetify'
...
-    store.state.stateModel.tombstone.entityType = 'BEN'
-    // Temp Id will always be set with or without an NR
-    store.state.stateModel.tombstone.businessId = 'T1234567'
-    store.state.stateModel.nameRequest.nrNumber = 'NR 1234567'
-    store.state.stateModel.nameRequest.details.approvedName = 'Xyz Ltd.'
-
-    wrapper = mount(EntityInfo, { vuetify, store })
+    let state = store.state.stateModel
+
+    state.tombstone.keycloakRoles = ['staff']
```
severinbeauvais, line 54: You can move this common stuff, which only needs to be called once, into a `beforeAll` method.

Nice use of `state` to shorten the assignments BTW 👍 
(It might even be a bit quicker as you don't have to de-reference the store object every time.)

tests/unit/entityInfo.spec.ts
```diff
@@ -53,58 +86,119 @@ describe.skip('Entity Info component with an NR', () => {
...
+      parties: [],
+      shareClasses: [],
+      incorporationAgreement: {}
+    }
+  }
 
-    expect(wrapper.vm.$el.querySelector('#entity-legal-name').textContent)
-      .toContain('Numbered Benefit Company')
+  beforeEach(() => {
+    store = getVuexStore()
```
severinbeauvais, line 142: Same as above for `store` and state assignments below.

severinbeauvais (approved):
Looks good. Just a few suggestions :)

tests/unit/entityInfo.spec.ts
cameron-freshworks: done and done, updated in next commit.

tests/unit/entityInfo.spec.ts
cameron-freshworks: Copy that, will clean this up. Fixed in next commit.

tests/unit/entityInfo.spec.ts
cameron-freshworks: Updated in next commit.

tests/unit/entityInfo.spec.ts
cameron-freshworks: For sure, no problems there. I should have done this initially. Updated in next commit.

tests/unit/entityInfo.spec.ts
cameron-freshworks: fixed in next commit!

tests/unit/entityInfo.spec.ts
cameron-freshworks: nice, good optimization! Updated in next commit.

tests/unit/entityInfo.spec.ts
cameron-freshworks: Fixed in next commit.

cameron-freshworks:
> Looks good. Just a few suggestions :)

Great suggestions, thanks! Updated 👍

tests/unit/entityInfo.spec.ts
```diff
@@ -15,11 +15,12 @@ import VueRouter from 'vue-router'
 
 Vue.use(Vuetify)
 const vuetify = new Vuetify({})
+const store = getVuexStore()
+let state = store.state.stateModel
```
severinbeauvais, line 19: By the way, you never reassign `state`, so it can be a const.

severinbeauvais:
Very nice. Good cleanup!

tests/unit/entityInfo.spec.ts
severinbeauvais: You can fix it next time...

PR #77, opened 2020-10-22
Alterations Framework

cameron-freshworks:
*Issue #:* /bcgov/entity#4721

*Description of changes:*

* Entry into Alterations with just a business Identifier. NO DRAFT AT THIS POINT.

* Implement requests to Legal api and Auth api to retrieve relevant data for the Alteration Filing.

* Parse the collected data into existing store structure.

* Created store structure for original Alteration Filing which will serve for comparisons and restoration to original data.

* Display the Data in existing components that are functional.

*Notes:*

* Many if not all components will need some level of work to meet the requirements of alterations design.

* Need to Build an Alteration Filing and Save it to the Original Alteration Store ( for future diffs on drafts ).

* Need to handle App entry with an Alteration Draft ID. Code currently commented and will need refactoring.


src/interfaces/filing-interfaces/alterationFiling-interface.ts
```diff
@@ -0,0 +1,18 @@
+import { AlterationIF } from '@/interfaces'
+
+export interface AlterationFilingIF {
+  header: {
+    name: string
+    certifiedBy: string
+    date: string
+    filingId?: number // Optional as this is not required when building a filing - causes an error for new filings
+    folioNumber?: string // Optional to the user and only displayed for certain account types
+  }
```
severinbeauvais, line 10: No change needed now, but in future we will need to add staff payment fields in here.

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
```diff
@@ -0,0 +1,23 @@
...
+  businessData: {
+    business: BusinessInformationIF
+  },
+  businessAliases: {
+    aliases: Array<NameTranslationSnapshotIF>
+  },
+  businessAddresses: IncorporationAddressIf,
+  contactInfo: BusinessContactIF,
+  businessDirectors: {
+    directors: OrgPersonIF
```
severinbeauvais, line 20: Why nest this?

severinbeauvais: Or "businessData" and "businessAliases" above?

src/interfaces/stepper-interfaces/YourCompany/business-contact-interface.ts
```diff
@@ -1,6 +1,6 @@
 export interface BusinessContactIF{
     email: string
-    confirmEmail: string
+    confirmEmail?: string
```
severinbeauvais, line 3: Why is this optional? Isn't this the same scenario as for a correction?

src/mixins/filing-template-mixin.ts
```diff
@@ -164,6 +165,17 @@ export default class FilingTemplateMixin extends Vue {
...
   }
 
+  /**
+   * Builds an Alteration filing body from store data. Used when saving a filing.
+   * @param isDraft boolean indicating whether this is a draft
+   * @returns the Alteration filing body to save
+   */
+  buildAlterationFiling (isDraft: boolean): AlterationFilingIF {
+    // Alteration body to save TBD
+    const alteration: AlterationFilingIF = null
```
severinbeauvais, line 175: Or maybe `{}` ?

src/router/router.ts
```diff
@@ -29,12 +29,6 @@ export function getVueRouter () {
         name: 'signin',
         query: { redirect: to.fullPath }
       })
-    } else if (!isSigninRoute(to) && !isSignoutRoute(to) && !haveQueryParam(to.query)) {
-      // for normal routes, re-route along with query params
-      next({
-        name: to.name,
-        query: { ...from.query, ...to.query }
-      })
```
severinbeauvais, line 37: Did you test this change with Corrections?

src/mixins/legal-api-mixin.ts
```diff
@@ -131,4 +131,52 @@ export default class LegalApiMixin extends Mixins(FilingTemplateMixin) {
     const config = { baseURL: authUrl }
     return axios.get('users/@me', config)
   }
+
+  /**
+   * Fetch business data for current business.
+   * @param datatype The type of data to request.
+   * @returns a promise to return the business base info or of the specified data type.
```
severinbeauvais, line 138: Small typo.

src/mixins/legal-api-mixin.ts
```diff
@@ -131,4 +131,52 @@ export default class LegalApiMixin extends Mixins(FilingTemplateMixin) {
...
+  getContactInfo (): Promise<any> {
+    if (!this.getBusinessId) throw new Error('Invalid parameter \'businessIdentifier\'')
+
+    const url = `entities/${this.getBusinessId}`
+    const authUrl = sessionStorage.getItem('AUTH_API_URL')
+    const config = { baseURL: authUrl }
+
+    return axios.get(url, config)
+      .then(response => {
+        if (response && response.data) {
```
severinbeauvais, line 173: Should add check for `response.data.contacts` and/or empty array?

src/mixins/filing-template-mixin.ts
```diff
@@ -312,6 +324,38 @@ export default class FilingTemplateMixin extends Vue {
...
+    this.setBusinessInformation(businessSnapshot.businessData.business)
+
+    // Set Name Translations
+    this.setNameTranslations(
+      businessSnapshot.businessAliases.aliases?.map(x => {
+        return {
+          value: x.alias,
+          oldValue: null,
+          action: null
+        }
```
severinbeauvais, line 342: Nice, but would it make more sense to have oldValue and action nullable? (wondering not sayin')

src/interfaces/filing-interfaces/alterationFiling-interface.ts
cameron-freshworks: right, good catch, thanks!

src/views/Alteration.vue
```diff
@@ -1,55 +1,66 @@
...
+    CompletingParty,
+    Detail,
+    PeopleAndRoles,
+    ShareStructure,
+    StaffPayment,
+    YourCompany
   }
 })
 export default class Alteration extends Mixins(LegalApiMixin, FilingTemplateMixin) {
+  @State stateModel
```
severinbeauvais, line 60: Ack, get rid of this (and `@State`) 😬

src/views/Alteration.vue
```diff
@@ -142,6 +157,25 @@ export default class Alteration extends Mixins(LegalApiMixin, FilingTemplateMixi
...
   }
 
+  /** Fetch Business Snapshot */
+  private async fetchBusinessSnapshot (): Promise<any> {
+    const businessData = await this.getBusinessData()
+    const businessAliases = await this.getBusinessData(BusinessDataTypes.TRANSLATIONS)
+    const businessAddresses = await this.getBusinessData(BusinessDataTypes.ADDRESSES)
+    const businessDirectors = await this.getBusinessData(BusinessDataTypes.DIRECTORS)
+    const businessShareStructure = await this.getBusinessData(BusinessDataTypes.SHARE_CLASSSES)
+    const contactInfo = await this.getContactInfo()
```
severinbeauvais, line 167: You should do these in parallel!

https://github.com/bcgov/business-filings-ui/blob/acddf879c61ecda163ced5e8b236d2768ffb08f6/src/App.vue#L311

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
cameron-freshworks: This was the way the data is returned from the backend and also due to creating on a single request method for the varying datatypes. I will look at refactoring it, to de-structure prior to this point.

src/views/Alteration.vue
```diff
@@ -107,29 +118,33 @@ export default class Alteration extends Mixins(LegalApiMixin, FilingTemplateMixi
...
+      //
+      // // parse alteration filing into store
+      // this.parseAlteration(alterationFiling)
+      // } else {
+      //   // as we don't have the necessary query params, do not proceed
+      //   throw new Error('Invalid alteration filing ID')
+      // }
+
+      const businessSnapshot = await this.fetchBusinessSnapshot()
+      await this.parseBusinessSnapshot(businessSnapshot)
```
severinbeauvais, line 147: You should handle the "promise rejected" state (ie, catch).

src/interfaces/stepper-interfaces/YourCompany/business-contact-interface.ts
cameron-freshworks: This isn't an actual value saved in the back end right? I think this prop is purely for front end validations.

src/mixins/filing-template-mixin.ts
cameron-freshworks: yea :P Sorry this is just a dummy placeholder so i could uphold the return type 😆

src/views/Alteration.vue
severinbeauvais: ^^ would happen in case of fetch error

Your "get" methods could return an exception, which becomes a rejected promise when awaiting. (Or you could wrap everything in a try-catch block, but you should still handle it somehow.)

src/router/router.ts
cameron-freshworks: Yea, corrections doesn't ever catch in this block because we don't enter corrections without a query param.
Also removed the `corrected-id` because we are always entering with a draft. ie `correction-id`.

cameron-freshworks: Could certainly add in another check here, though i'm not sure what at this point.

src/mixins/legal-api-mixin.ts
cameron-freshworks: I see no spelling mist.... 😆  Fixed in next commit!

src/mixins/legal-api-mixin.ts
cameron-freshworks: Copy that, will do!

src/interfaces/stepper-interfaces/YourCompany/business-contact-interface.ts
severinbeauvais: You're right. And I now think having it optional makes more sense.

src/router/router.ts
severinbeauvais: What happens if we do enter a correction without a query param?

Also did you try loading an alteration and a correction while NOT first logged in. It should redirect you to login and the return here. (Might be tricky to test with IDIR login, sigh.)

src/mixins/legal-api-mixin.ts
severinbeauvais: Grate!

src/views/Alteration.vue
cameron-freshworks: yackkk, will do. Testin :P

src/views/Alteration.vue
cameron-freshworks: oh i like it! Will do, thats awesome.

vysakh-menon-aot (approved):

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
cameron-freshworks: Fixed in next commit!

src/router/router.ts
cameron-freshworks: For sure!

If we try to enter correction without a query param, it errors out while trying to load the app. Get's caught on the filingId. 
If we have this code block in however, it gets caught in a router error which is somewhat of a loop `Maximum Call stack exceeds..` etc etc. 

This is also the same for Alterations without a query param with this block in.

If we try and enter either without a token in session, it directs us to Auth, and returns us the route of origin as expected.

Short answer, i think in the present state we are ok without this check for a query param BUT we should handle the error where no query param is present or invalid for CORRECTIONS.

src/mixins/filing-template-mixin.ts
cameron-freshworks: I don't follow entirely. What would the difference be here between defining them as null vs having them as nullable?

src/views/Alteration.vue
cameron-freshworks: Fixed in next commit. Really like that approach!

src/views/Alteration.vue
cameron-freshworks: Wouldn't the thrown errors be caught by the parent Try/Catch block and emit from line 151?

src/mixins/legal-api-mixin.ts
cameron-freshworks: Updated in next commit 👍

src/views/Alteration.vue
cameron-freshworks: Fixed in next commit

src/mixins/filing-template-mixin.ts
severinbeauvais: If they were nullable, you wouldn't have to assign them to null here to avoid an object type error.

When I saw this, I wondered what those were used for, since you simply assigned them to null. I realize we are trying to use a structured type that matches what we ultimately send to the BE, but if the FE doesn't need them, maybe we don't need to assign them?

Anyway, no need to change this. I'm probably overthinking this 😁

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
```diff
@@ -0,0 +1,17 @@
...
+  ShareStructureIF
+} from '@/interfaces'
+
+export interface BusinessSnapshotIF {
+  business: BusinessInformationIF,
+  aliases: Array<NameTranslationSnapshotIF>,
+  IncorporationAddressIf,
+  directors: OrgPersonIF,
+  shareClasses: ShareStructureIF,
+  BusinessContactIF
```
severinbeauvais, line 16: This looks over-optimized 😛 
The sub-object should have a name (that isn't the same as the IF type), please.

src/views/Alteration.vue
severinbeauvais: Oops, I didn't notice you were already in a try-catch block.

Yes, yes they would.

src/views/Alteration.vue
```diff
@@ -142,6 +155,18 @@ export default class Alteration extends Mixins(LegalApiMixin, FilingTemplateMixi
...
 
+  /** Fetch Business Snapshot */
+  private async fetchBusinessSnapshot (): Promise<any> {
+    return Promise.all([
+      this.getBusinessData(),
+      this.getBusinessData(BusinessDataTypes.TRANSLATIONS),
+      this.getBusinessData(BusinessDataTypes.ADDRESSES),
+      this.getBusinessData(BusinessDataTypes.DIRECTORS),
+      await this.getBusinessData(BusinessDataTypes.SHARE_CLASSSES),
+      await this.getContactInfo()
```
severinbeauvais, line 166: I think you don't need the awaits here, since they already return a promise.

Also, you should add a check that you got an array with 6 responses, otherwise throw an error that something is missing.

src/views/Alteration.vue
severinbeauvais: Also remove unused import.

src/views/Alteration.vue
severinbeauvais: I recommend puttin the array length check in parseBusinessSnapshot() to:
- keep THIS method clean (just a return statement)
- protect parseBusinessSnapshot() from an out-of-bounds error if one of these fetches fails
- this would also clarify that parseBusinessSnapshot() expects an array of responses

src/mixins/filing-template-mixin.ts
```diff
@@ -312,6 +324,39 @@ export default class FilingTemplateMixin extends Vue {
...
+      })
+    )
+
+    // Set Office Addresses
+    this.setOfficeAddresses(businessSnapshot[2])
+
+    // Set People and Roles
+    this.setPeopleAndRoles(businessSnapshot[3].directors)
```
severinbeauvais, line 351: Nice! 👍

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
cameron-freshworks: Fixed in next commit with the unit tests 👍

src/views/Alteration.vue
cameron-freshworks: Sounds great! Added in the unit test commit 👍

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  BusinessContactIF,
+  BusinessInformationIF,
+  IncorporationAddressIf,
+  NameTranslationSnapshotIF,
+  getOrgPersonsIF,
```
severinbeauvais, line 6: Please follow naming convention (IF starts with capital letter).

src/mixins/filing-template-mixin.ts
```diff
@@ -1,20 +1,26 @@
...
-  NameTranslationDraftIF
+  NameTranslationDraftIF,
+  RoleIF
 } from '@/interfaces'
+
 import { StaffPaymentIF } from '@bcrs-shared-components/interfaces'
+
 // Constants
-import { ActionTypes, EntityTypes, FilingTypes } from '@/enums'
+import { ActionTypes, EntityTypes, FilingTypes, Roles } from '@/enums'
```
severinbeauvais, line 23: Could you please rename Roles to RoleTypes (as I've done in another project)?

src/mixins/filing-template-mixin.ts
```diff
@@ -312,6 +329,51 @@ export default class FilingTemplateMixin extends Vue {
...
+    this.setPeopleAndRoles(businessSnapshot[3].directors?.map(director => {
+      return {
+        ...director,
+        roles: [
+          {
+            roleType: director.role,
+            appointmentDate: director.appointmentDate,
+            cessationDate: null
+          }
+        ]
```
severinbeauvais, line 366: Pretty fancy!

I would prefer that you not spread `director` above, as this will result in invalid properties in the director record (OrgPersonIF) -- specifically, appointmentDate, cessationDate and role.

src/interfaces/stepper-interfaces/PeopleAndRoles/org-person-interface.ts
```diff
@@ -16,3 +16,15 @@ export interface OrgPersonIF {
   deliveryAddress?: AddressIF
   action?: ActionTypes
 }
+
+export interface getOrgPersonsIF {
+  appointmentDate: string,
+  cessationDate: string,
+  deliveryAddress: AddressIF,
+  mailingAddress: AddressIF,
+  officer: {
```
severinbeauvais, line 25: No `officer.id` here? This will need to be fixed (future sprint when we handle director changes).

src/mixins/filing-template-mixin.ts
```diff
@@ -1,20 +1,26 @@
...
+  BusinessSnapshotIF,
   CorrectionFilingIF,
   IncorporationFilingIF,
   OrgPersonIF,
   ShareClassIF,
   StateModelIF,
   NameTranslationIF,
-  NameTranslationDraftIF
+  NameTranslationDraftIF,
+  RoleIF
```
severinbeauvais, line 17: I think you don't actually use RoleIF (or Roles interface below) in this module -- can be deleted from imports.

src/mixins/filing-template-mixin.ts
cameron-freshworks: Bit outside scope but sure..
Also Unused import, removed 👍

src/mixins/filing-template-mixin.ts
cameron-freshworks: k

src/mixins/filing-template-mixin.ts
severinbeauvais: You're right. This cleanup can be done later. Thanks either way :)

src/views/Alteration.vue
severinbeauvais: FYI, the browser can handle several simultaneous/pending network calls. I don't know what the current max is but I think it's over 10; anything over that has to wait for a previous call to complete.

As you can surmise, this shortens the fetch time from "sum of all fetches" to "time of longest fetch". This also allows increased performance/optimization of the web server/API through caching or multiple instances (ie, scaling).

src/interfaces/stepper-interfaces/PeopleAndRoles/org-person-interface.ts
cameron-freshworks: Sounds good 👍  Currently no ID returned from the backend. Put it in, optional for now.

src/views/Alteration.vue
```diff
@@ -1,55 +1,60 @@
...
+    Detail,
+    PeopleAndRoles,
+    ShareStructure,
+    StaffPayment,
+    YourCompany
   }
 })
 export default class Alteration extends Mixins(LegalApiMixin, FilingTemplateMixin) {
   // Global getters
   @Getter isRoleStaff!: boolean
```
severinbeauvais, line 57: Unused getter?

src/mixins/filing-template-mixin.ts
severinbeauvais: Actually, `{}` probably won't work anyway as it won't contain the necessary properties for the interface.

cameron-freshworks: This is just a placeholder for a future method. I can remove the whole thing if you prefer?

src/mixins/filing-template-mixin.ts
severinbeauvais: Thanks! 🙏

src/views/Alteration.vue
cameron-freshworks: Yea, this was an excellent way to do it. It actually improved the laod time significantly 👍 Thanks!

severinbeauvais (approved):
AWESOME job on getting the alteration framework so complete! And thanks for the extra cleanup too. 👍 👍

src/mixins/filing-template-mixin.ts
severinbeauvais: No, I like it. This is definitely "framework stuff" that we will expand later.

PR #80, opened 2020-10-26
Implemented Certify

cameron-freshworks:
*Issue #:* /bcgov/entity#4851

*Description of changes:*

* Import and tie Certify Component to Store

* Fixed Unit tests(WIP)

* Fixed time conversion error in Date Mixin

*Notes:*

* Will need to Update Date Mixin in Filings/Create UI to correct bug.

* A small layout refactor is required for the Shared Component to bring in line with similar components, work is captured in this PR: https://github.com/bcgov/bcrs-shared-components/pull/11

### DIFF
--- a/src/components/common/Certify.vue
+++ b/src/components/common/Certify.vue
@@ -3,20 +3,75 @@
     <header>
       <h2>2. Certify</h2>
     </header>
-
     <span>Enter the legal name of the person authorized to complete and submit these changes.</span>
-
-    <div id="certify-container" class="mt-4">
-      Under construction!
-    </div>
+    <certify
+      :currentDate="readableCurrentDate"
+      :certified-by="getCertifyState.certifiedBy"
+      :is-certified="getCertifyState.valid"
+      :entity-display="readableEntityType"
+      :message="certifyMessage"
+      @update:certifiedBy="onCertifiedBy($event)"
+      @update:isCertified="onValid($event)"
+    />
   </section>
 </template>
 
 <script lang="ts">
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Mixins } from 'vue-property-decorator'
+import { Action, Getter } from 'vuex-class'
 
-@Component({})
-export default class Certify extends Vue {
+// Components
+import { Certify } from '@bcrs-shared-components/certify'
+
+// Interfaces & Mixins
+import { DateMixin, EnumMixin } from '@/mixins'
+import { ActionBindingIF, CertifyIF } from '@/interfaces'
+import { EntityTypes } from '@/enums'
+
+@Component({
+  components: {
+    Certify
+  }
+})
+export default class CertifySection extends Mixins(DateMixin, EnumMixin) {
+  @Getter getCertifyState!: CertifyIF
+  @Getter getCurrentDate!: string
+  @Getter getEntityType!: EntityTypes
+
+  @Action setCertifyState!: ActionBindingIF
+
+  readonly certifyMessage = 'Note: It is an offence to make a false of misleading statement in respeoct of a material' +
+    'fact in a record submitted to the Corporate Registry for filing. See section 427 of the Business Corporations Act.'
+
+  /** Get current date in a readable format */
+  private get readableCurrentDate (): string {
+    return this.toReadableDate(this.getCurrentDate)
+  }
+
+  /** Get the entity type in readable format */
+  private get readableEntityType (): string {
+    return this.entityTypeToDescription(this.getEntityType)
+  }
+
+  /** Handler for Valid change event. */
+  private onValid (val: boolean): void {
+    this.setCertifyState(
+      {
+        valid: val,
+        certifiedBy: this.getCertifyState.certifiedBy
+      }
+    )
+  }
+
+  /** Handler for CertifiedBy change event. */
+  private onCertifiedBy (val: string): void {
+    this.setCertifyState(
+      {
+        valid: this.getCertifyState.valid,
+        certifiedBy: val
+      }
+    )
+  }
 }
 </script>
 
--- a/src/components/common/Certify.vue
+++ b/src/components/common/Certify.vue
@@ -23,9 +23,11 @@ import { Action, Getter } from 'vuex-class'
 // Components
 import { Certify } from '@bcrs-shared-components/certify'
 
-// Interfaces & Mixins
+// Mixins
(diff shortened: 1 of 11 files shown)


src/components/common/Certify.vue
```diff
@@ -3,20 +3,75 @@
...
+  }
+})
+export default class CertifySection extends Mixins(DateMixin, EnumMixin) {
+  @Getter getCertifyState!: CertifyIF
+  @Getter getCurrentDate!: string
+  @Getter getEntityType!: EntityTypes
+
+  @Action setCertifyState!: ActionBindingIF
+
+  readonly certifyMessage = 'Note: It is an offence to make a false of misleading statement in respeoct of a material' +
```
severinbeauvais, line 43: Do we still have "resources" somewhere that contain these strings for different entity types?

cameron-freshworks: o geez, yea we do i think. Didn't even cross my mind. 
Will cross reference to see if that matches the current design and implement if so!

severinbeauvais (approved):
Looks good to me.
Unit tests?

cameron-freshworks:
> Looks good to me.
> Unit tests?

I'm on it!

src/components/common/Certify.vue
cameron-freshworks: Added in latest commit!

severinbeauvais: Awesome!

rarmitag (approved):

src/mixins/date-mixin.ts
```diff
@@ -28,9 +28,12 @@ export default class DateMixin extends Vue {
    * @returns the formatted date string
    */
   toReadableDate (date: string): string {
-    // cast to a workable dateString
+    // cast to a workable dateString 'America/Vancouver'
```
severinbeauvais, line 31: No, the tz depends on the locale for this code (ie, browser). This comment could mislead.

src/mixins/date-mixin.ts
```diff
@@ -28,9 +28,12 @@ export default class DateMixin extends Vue {
...
     // split into an array
-    let formatDate = (new Date(date).toDateString()).split(' ')
+    let newDate = new Date(date)
+    // Factor in the UTC offset so date displayed is local
+    newDate.setTime(newDate.getTime() + newDate.getTimezoneOffset() * 60 * 1000)
```
severinbeauvais, line 35: This change needs visibility in this commit/PR/ticket (ie, for traceability and external testing).

Also this method is now different from the same method in our other projects. What's the plan to update those?

---

Would it make more sense to have a **single** UTC-to-local method in this mixin (ie, convertUtcTimeToLocalTime () below) and use **this** method to simply do the formatting?

src/mixins/date-mixin.ts
cameron-freshworks: Sorry, that an error. I was testing some things out and missed removing that comment

src/mixins/date-mixin.ts
cameron-freshworks: I'm not sure if one method will be applicable to all? The dates seem to come in many different formats.

We can/will update this in the other repos as well :) It was an error that would be present in any of its use cases.

src/components/common/Certify.vue
```diff
@@ -3,20 +3,80 @@
...
+// Interfaces, Enums & Resources
+import { ActionBindingIF, CertifyIF, CertifyStatementIF } from '@/interfaces'
+import { EntityTypes } from '@/enums'
+
+@Component({
+  components: {
+    Certify
+  }
+})
+export default class CertifySection extends Mixins(DateMixin, EnumMixin) {
```
severinbeauvais, line 38: Please rename the file too :)

cameron-freshworks: Recommendations?  We can't have 2 `Certify` components :P

src/mixins/date-mixin.ts
severinbeauvais: I'm concerned about 2 methods in this mixin that apparently do something similar (and do it in different ways!).

severinbeauvais: I'm also concerned that the description for this method says the date comes in as YYYY-MM-DD (ie, no time and no timezone)... is it for sure a UTC date? If not then this is dangerous as it could be used on a local date and give an incorrect result.

src/components/common/Certify.vue
severinbeauvais: Actually we can, because this file is the section and the shared component isn't in this repo.

But since you renamed this class, could you also rename this file to `CertifySection.vue` ?

cameron-freshworks: I suppose we could but when importing them both for sake of testing, how can we differentiate one from the other there?

src/mixins/date-mixin.ts
cameron-freshworks: Hmm yea, not sure what to do here. My goal was to correct a bug found during unit testing, but if you think we need to look at this mixin and refactor as a whole we can certainly do that.

src/components/common/Certify.vue
severinbeauvais: Let's chat in person.

src/mixins/date-mixin.ts
severinbeauvais: What if you don't fix it here and instead create a ticket to review/fix this for all 3 projects?

cameron-freshworks: That sounds great, we should further discuss the solution!

src/components/common/Certify.vue
cameron-freshworks: Copy that!

src/mixins/date-mixin.ts
severinbeauvais: For posterity: We decided to NOT format the current date in the Certify component at this time (since it raised the issues above). This is now consistent with the other 2 UIs. A new ticket will be created to format the current date consistently (and correctly) in all 3 UIs at the same time.

severinbeauvais:
LGTM. Merge when you want.

cameron-freshworks:
> LGTM. Merge when you want.

Copy that. work to cover to the update date display formats across the UI's is captured in ticket: 
https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/5425

PR #81, opened 2020-10-27
5305 UX Editing Nested Series Behaviour Update

cameron-eyds:
*Issue #:* /bcgov/entity#5305

*Description of changes:*

* Allows inline editing of specific series opposed to hiding all the series associated with a class.

* Fixed Console error where component was loading before resource was set for certify.

*Addressed P2 updates.:*

* Changed colour of `cancel` btn in confirm dialog.

* Added a `.` to warning modal in Edit Share Strucutre.

### DIFF
--- a/src/components/ShareStructure/EditShareStructure.vue
+++ b/src/components/ShareStructure/EditShareStructure.vue
@@ -406,7 +406,7 @@ export default class EditShareStructure extends Mixins(CurrencyLookupMixin) {
       this.$refs.confirm.open(
         'Remove Share Series with Class',
         'A share series exists for this class. Removing the Special Rights or Restrictions for this class' +
-        ' will remove all associated share series',
+        ' will remove all associated share series.',
         {
           width: '45rem',
           persistent: true,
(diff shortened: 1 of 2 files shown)


vysakh-menon-aot (approved):

lekshmimallika-aot (approved):

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -791,6 +815,11 @@ export default class ShareStructure extends Mixins(CommonMixin) {
     }
   }
 
+  /** Map an empty 2d array to handle the unknown size of nested Share Classes and Series */
+  private mapEmpty2dArray (): Array<any> {
+    return new Array(50).fill(null).map(() => new Array(50).fill(null))
```
severinbeauvais, line 820: Not sure if you need this. I believe, in JS, arrays auto-resize so you can just ref an item (and don't have to pre-size the array).

severinbeauvais: https://stackoverflow.com/a/33903611/8679162

severinbeauvais: Or does this not work for 2D arrays?

cameron-freshworks: I did have issues managing this yea, specifically because it was a multidimensional array.
In js, i don't think multidimensional arrays really exist natively or are supported, it just boils down to arrays within an array. 
The issue was setting a baseline from which to work with / reset too. 
I can dig into it and see if there is another option tho!

severinbeauvais: Nah, it's OK for now.

Change it later if this keeps you up at night.

PR #90, opened 2020-11-21
5701 name translation schema change

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5701

*Description of changes:*

### DIFF
--- a/src/components/YourCompany/NameTranslations/NameTranslation.vue
+++ b/src/components/YourCompany/NameTranslations/NameTranslation.vue
@@ -173,9 +173,9 @@ export default class NameTranslation extends Mixins(CommonMixin) {
 
   private get hasPendingChange (): boolean {
     return this.draftTranslations.length !== this.nameTranslations.length ||
-      !this.draftTranslations.every((translation, index) => {
-        return this.nameTranslations[index].name === translation.name &&
-          this.nameTranslations[index].action === translation.action
+      this.draftTranslations.some((translation, index) => {
+        return this.nameTranslations[index].name !== translation.name ||
+          this.nameTranslations[index].action !== translation.action
       })
   }
 
--- a/src/components/YourCompany/NameTranslations/NameTranslation.vue
+++ b/src/components/YourCompany/NameTranslations/NameTranslation.vue
@@ -15,7 +15,7 @@
       </v-flex>
       <v-flex xs7 v-if="draftTranslations && translationsExceptRemoved.length">
         <div v-for="(translation, index) in translationsExceptRemoved"
-          :key="`name_translation_${index}`">{{translation.value}}</div>
+          :key="`name_translation_${index}`">{{translation.name}}</div>
       </v-flex>
       <v-flex xs7 v-else>
         No name translations
@@ -137,8 +137,7 @@ import { ConfirmDialog } from '@/components/dialogs'
 import { ListNameTranslation, AddNameTranslation } from '.'
 
 // Interfaces
-import { ActionBindingIF, ConfirmDialogType,
-  NameTranslationDraftIF, NameTranslationIF } from '@/interfaces'
+import { ActionBindingIF, ConfirmDialogType, NameTranslationIF } from '@/interfaces'
 
 // Enums
 import { ActionTypes } from '@/enums'
@@ -160,13 +159,13 @@ export default class NameTranslation extends Mixins(CommonMixin) {
   }
 
   @Prop({ default: () => { return [] as [] } })
-  private nameTranslations!: NameTranslationDraftIF[]
+  private nameTranslations!: NameTranslationIF[]
 
   // Actions
   @Action setEditingNameTranslations: ActionBindingIF
 
   // Properties
-  private draftTranslations: NameTranslationDraftIF[] = []
+  private draftTranslations: NameTranslationIF[] = []
   private isEditing: boolean = false
   private isAddingNameTranslation = false
   private editingNameTranslation = ''
@@ -175,7 +174,7 @@ export default class NameTranslation extends Mixins(CommonMixin) {
   private get hasPendingChange (): boolean {
     return this.draftTranslations.length !== this.nameTranslations.length ||
       !this.draftTranslations.every((translation, index) => {
-        return this.nameTranslations[index].value === translation.value &&
+        return this.nameTranslations[index].name === translation.name &&
           this.nameTranslations[index].action === translation.action
       })
   }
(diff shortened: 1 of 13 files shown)


cameron-freshworks (approved):
Looks good 👍

src/components/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -175,7 +174,7 @@ export default class NameTranslation extends Mixins(CommonMixin) {
   private get hasPendingChange (): boolean {
     return this.draftTranslations.length !== this.nameTranslations.length ||
       !this.draftTranslations.every((translation, index) => {
```
severinbeauvais, line 176: This could be improved by changing `!every(x===y)` to `some(x!==y)` :)

If you don't feel like making this change, we'll get to it later.

src/store/state/state-model.ts
```diff
@@ -128,9 +126,7 @@ export const stateModel: StateModelIF = {
       nameRequest: {
         legalType: ''
       },
-      nameTranslations: {
-        new: []
-      },
+      nameTranslations: [],
```
severinbeauvais, line 129: Do we have to worry about loading existing IAs (with old schema)?

src/mixins/filing-template-mixin.ts
```diff
@@ -187,21 +186,17 @@ export default class FilingTemplateMixin extends Vue {
...
-        .map(x => x.value)
-    } as NameTranslationIF
+  prepareNameTranslations () : NameTranslationIF[] {
+    return this.getNameTranslations?.filter(x => x.action === ActionTypes.ADDED || x.action === ActionTypes.EDITED)
+      .map(x => {
+        let nameTranslation = {
+          name: x.name
+        }
+        if (x.action === ActionTypes.EDITED) {
+          nameTranslation['id'] = x.id
```
severinbeauvais, line 196: I didn't see where "id" was originally assigned...?

If it orginally comes from the API, how does it get updated for new names?

src/mixins/filing-template-mixin.ts
```diff
@@ -216,21 +211,16 @@ export default class FilingTemplateMixin extends Vue {
...
-          return {
-            value: x,
-            oldValue: null,
-            action: null
-          }
-        })
-      )
-    }
+    this.setNameTranslations(
+      filing.incorporationApplication.nameTranslations?.map(x => {
```
severinbeauvais, line 215: The optional chaining means this value could be "undefined", so you probably want to add a `|| []` in here.

severinbeauvais: Hmm, actually, it looks like this "map" function is converting `{ id, name, oldName, action }` into `{ id, name, oldName, action }` -- in others words, is it really needed?

src/mixins/filing-template-mixin.ts
```diff
@@ -187,21 +186,17 @@ export default class FilingTemplateMixin extends Vue {
...
-        .map(x => {
-          return {
-            newValue: x.value,
-            oldValue: x.oldValue
-          }
-        }),
-      ceased: translations?.filter(x => x.action === ActionTypes.REMOVED)
-        .map(x => x.value)
-    } as NameTranslationIF
+  prepareNameTranslations () : NameTranslationIF[] {
```
severinbeauvais, line 189: I'd really like to see a unit test for the 3 methods you changed in this file.

Here's a test file for a mixin that you can copy:
https://github.com/bcgov/business-filings-ui/blob/master/tests/unit/common-mixin.spec.ts

severinbeauvais (commented):
The code looks good but I'm concerned about where "id" is set.

src/store/state/state-model.ts
vysakh-menon-aot: @thorwolpert verified that in prod we don't have any IA with name translation. So no need to consider that

src/mixins/filing-template-mixin.ts
vysakh-menon-aot: Based on schema id should not be included for new translation

severinbeauvais: OK. So what is ID used for?

vysakh-menon-aot: for modified translation

severinbeauvais: And where is it assigned?

vysakh-menon-aot: line no 217

severinbeauvais:
```
id: x.id,
```
But where did `x.id` come from? I don't see where id is assigned a number.

src/mixins/filing-template-mixin.ts
vysakh-menon-aot: For the first time (when we initiate a correction) the oldName and action props are not available in the api response, which creates an issue of not having these props in store. Due to missing props change event was not triggering if the action value is changed (at the time of Delete there is no other prop change except action). To handle this scenario I had to keep this structure.

src/mixins/filing-template-mixin.ts
vysakh-menon-aot: will look into it

src/mixins/filing-template-mixin.ts
vysakh-menon-aot: pk from database. FE wont be setting any value for id

src/mixins/filing-template-mixin.ts
severinbeauvais: OK, sounds fine. Could you please add a comment similar to your response, for when another developer wonders about this code? Thanks!

src/mixins/filing-template-mixin.ts
severinbeauvais: Oh OK. So the UI maintains it if the API provided it, otherwise it is empty?

vysakh-menon-aot: Yes. API does provide ID for the translations which is already been part of IA

src/mixins/filing-template-mixin.ts
vysakh-menon-aot: It may not be possible to include these in this ticket as we don't have an existing test file for this.

severinbeauvais: Admittedly, we haven't created test files for most mixins so far, which is why it would be nice to catch up a little here for code you updated.
However, I don't want to slow you down too much, so if you have higher priority work for today then please create a tech-debt ticket for some unit tests for this mixin, that you can maybe get around to soon.

severinbeauvais (approved):

lekshmimallika-aot (approved):

src/mixins/filing-template-mixin.ts
vysakh-menon-aot: I tried to include test cases for that. Then I realized to cover those methods I need to consider entire filing json. Which takes more time than what I have for this ticket. I was not pointing about not having test. I am just concerned about time :)

PR #99, opened 2021-01-18
5106 Alterations - Edit Company Name

cameron-freshworks:
*Issue #:* /bcgov/entity#5106

*Description of changes:*

- Updated Entity Info component.

- Stores an original business snapshot for the purposes of resetting to a base state and for value comparisons. 

- Parse and display additional Name Request and Business type data for Alterations.

- Update Change/Correct labels and indicators to be dynamic based on either Correction or Alteration filijngs.

- Added business type check to identify type mismatch. Store conflicts in a property and display Warning modal and UI icon when Name Request type is a mismatch to the current/original business type.

- Added some tooltip ovverides to emulate what we've done in Name Request with the Arrows.

- Some minor styling updates as per the Invision doc.
https://projects.invisionapp.com/d/main#/console/20562985/436252442/preview


vysakh-menon-aot (approved):

src/components/YourCompany/CompanyName/CorrectNameRequest.vue
```diff
@@ -58,19 +58,20 @@ import { Component, Prop, Watch, Emit, Mixins } from 'vue-property-decorator'
...
+import { CommonMixin, NameRequestMixin } from '@/mixins'
 
 // Interfaces & Enums
 import {
   ActionBindingIF,
   NameRequestApplicantIF,
   NameRequestIF,
-  NrCorrectionIF
+  NrCorrectionIF,
+  NrResponseIf
```
severinbeauvais, line 69: If -> IF

Or does it make more sense as If? If so then we should change for all interfaces.

src/components/YourCompany/CompanyName/CorrectNameRequest.vue
```diff
@@ -133,23 +134,35 @@ export default class CorrectNameRequest extends Mixins(NameRequestMixin) {
...
-      await this.validateNameRequest(this.nameRequestNumber, this.entityPhone, this.entityEmail)
-        .then(response => {
-          const nrCorrection: NrCorrectionIF = {
-            nrNumber: this.nameRequestNumber,
-            legalName: this.getNrApprovedName(response),
-            applicant: {
-              phoneNumber: this.entityPhone,
-              emailAddress: this.entityEmail
-            }
+      try {
```
severinbeauvais, line 137: Nice. This is cleaner than old code.

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
```diff
@@ -5,6 +5,8 @@ export interface NameRequestIF {
   legalType: EntityTypes
   legalName: string
   nrNumber: string
+  expiry?: string
+  status?: string
```
severinbeauvais, line 9: Can we use an enum type for this?

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
```diff
@@ -20,6 +22,8 @@ export interface NameRequestDetailsIF {
 
 // Name request applicant details interface
 export interface NameRequestApplicantIF {
+  address?: string
```
severinbeauvais, line 25: Is address really a string?

src/interfaces/correction-interfaces/correct-name-interface.ts
```diff
@@ -9,7 +9,27 @@ export interface CorrectNameOptionIF {
 }
 
 export interface NrCorrectionIF {
+  legalType: string,
```
severinbeauvais, line 12: Can we type this (and status and entity_type_cd below)?

(It may be that we can't because we don't know what the API sends us. But at some point we should then convert string -> type for our use.)

src/components/common/EntityInfo.vue
```diff
@@ -60,7 +62,7 @@ export default class EntityInfo extends Mixins(EnumMixin) {
       case RouteNames.CORRECTION:
         return 'Correction - Incorporation Application'
       case RouteNames.ALTERATION:
-        return 'Alteration'
+        return 'Company Information'
```
severinbeauvais, line 65: Should this getter be in common mixin also?

src/components/YourCompany/YourCompany.vue
```diff
@@ -199,10 +350,10 @@ export default class YourCompany extends Mixins(DateMixin, EntityFilterMixin, Le
 
   @Watch('getApprovedName')
   private onApprovedName ():void {
-    if (this.getApprovedName) {
+    if (this.getApprovedName && !this.isNumberedCompany) {
       this.correctNameChoices = [
         CorrectionTypes.CORRECT_NEW_NR,
-        CorrectionTypes.CORRECT_NAME,
+        this.isCorrection() && CorrectionTypes.CORRECT_NAME, // Only allow editable name changes for Corrections
```
severinbeauvais, line 360: I think this will leave a falsey (undefined I think) element in the array.
It might be better to dynamically create the array (eg, with push/pop/slice/etc) instead.

src/components/YourCompany/YourCompany.vue
```diff
@@ -114,29 +206,50 @@
 <script lang="ts">
 import { Component, Emit, Mixins, Watch } from 'vue-property-decorator'
 import { Action, Getter } from 'vuex-class'
-import { ActionBindingIF, BusinessContactIF, GetterIF, IncorporationFilingIF, NameRequestIF } from '@/interfaces'
+import {
+  ActionBindingIF,
+  BusinessContactIF,
+  BusinessSnapshotIF,
+  ConfirmDialogType,
```
severinbeauvais, line 213: ConfirmDialogType is an IF?

src/components/YourCompany/YourCompany.vue
```diff
@@ -62,8 +126,36 @@
...
+          <v-flex xs3 class="mt-6">
+            <v-layout column>
+              <label><strong>Name Request Applicant</strong></label>
+            </v-layout>
+          </v-flex>
+          <v-flex xs6 class="mt-6">
+            <template v-if="hasNewNr">
+              <div>
+                <span class="font-weight-bold">Name: </span>
+                <span>{{getNameRequest.applicant.fullName}}</span>
```
severinbeauvais, line 140: What happens if getNameRequest failed?

src/components/YourCompany/CompanyName/CorrectNameRequest.vue
cameron-freshworks: typo! Will fix, thanks

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: We can indeed!

src/interfaces/correction-interfaces/correct-name-interface.ts
```diff
@@ -9,7 +9,27 @@ export interface CorrectNameOptionIF {
...
+  status: string
   applicant: NameRequestApplicantIF | {}
 }
+
+export interface NrResponseIf {
+  // eslint-disable-next-line camelcase
+  entity_type_cd: string, // Ignore lint error on property name served from api.
+  expirationDate: string,
+  state: string,
+  applicants: {
```
severinbeauvais, line 25: I think applicants is an array...?

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: I will rename the prop name, in this case it is sort formatted address string. A combination of all the individual fields for display purposes.
So possibly a better name for the prop would be `fullAddress` or `fullDisplayAddress` something along those lines.

src/components/YourCompany/CompanyName/CorrectNameRequest.vue
cameron-freshworks: Fixed in next commit 👍

src/components/YourCompany/YourCompany.vue
```diff
@@ -155,41 +271,80 @@ export default class YourCompany extends Mixins(DateMixin, EntityFilterMixin, Le
...
+
   /** The company name (from NR, or incorporation number). */
   private get companyName (): string {
     if (this.getApprovedName) return this.getApprovedName
 
     return `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.`
   }
 
+  private get expiryDate (): string {
+    return new Date(this.getNameRequest.expiry).toDateString()
```
severinbeauvais, line 285: We always get in trouble here...
- Does this correctly convert UTC -> Pacific?
- Is this a consistent display date format within this project (and other Entity projects)?

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: Fixed in next commit

tests/unit/YourCompany.spec.ts
```diff
@@ -11,6 +11,7 @@ import { getVuexStore } from '@/store'
 import { createLocalVue, mount } from '@vue/test-utils'
 import { BusinessContactInfo, FolioNumber, OfficeAddresses, YourCompany } from '@/components/YourCompany'
 import { CorrectNameOptions } from '@/components/YourCompany/CompanyName'
+import flushPromises from 'flush-promises';
```
severinbeauvais, line 14: Unused import.

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: Updated in next commit

severinbeauvais: That would tie in nicely with the common mixin method `formatFullAddress()` :)

src/interfaces/correction-interfaces/correct-name-interface.ts
cameron-freshworks: kk, i'll check it out!

tests/unit/YourCompany.spec.ts
cameron-freshworks: thank you!

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Indeed we do, we seem to get a lot of different formats to deal with. 

This correctly converts the time given by the API to local time, as an example the expiry is set at Wed Mar 10 2021 07:59 GMT... which will convert into the day prior 1 minute before the end of the day. 

That being said, i think that is correct, as it WILL expire at the very end of the day that is displayed. 

To the last bullet, i'm not entirely sure. It is consistent with the design but i think we have some instances in the Filings/Create UI that are still in an older format.

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: > This was an intentional time to verify the UTC offset, ie, 7:59 UTC is the day before in Pacific (either PST or PDT, not sure).
> This is to ensure that code doesn't just truncate the time and assume the date is correct.
> 
> As long as this is well tested, I don't really care about the times/dates, but this is an area where we often fall down, so please test thoroughly.

Absolutely, in this scenario we don't do anything in the UI with this date afaik. Just display it. 
So if this offset is intentional and consistent from the backend, the ui will reflect the correct date locally. 
On that note though, since it is a very specific offset catering to PST, there could be oddities in other time zones.. but really this should be left to the back end 😩

cameron-freshworks:
> This was an intentional time to verify the UTC offset, ie, 7:59 UTC is the day before in Pacific (either PST or PDT, not sure).
> This is to ensure that code doesn't just truncate the time and assume the date is correct.
> 
> As long as this is well tested, I don't really care about the times/dates, but this is an area where we often fall down, so please test thoroughly.

Absolutely, in this scenario we don't do anything in the UI with this date afaik. Just display it. 
So if this offset is intentional and consistent from the backend, the ui will reflect the correct date locally. 
On that note though, since it is a very specific offset catering to PST, there could be oddities in other time zones.. but really this should be left to the back end 😩

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
cameron-freshworks: Definitely :)

src/components/common/EntityInfo.vue
cameron-freshworks: Yea, i think so, will update!
I forgot to mention the changes to entity info in the PR. Thanks for the reminder!

cameron-freshworks: Updated in next commit

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Ok, good point. Will check it out!

src/components/YourCompany/YourCompany.vue
severinbeauvais: IIRC, the NR expires at 00:00:00 on its expiry date,

severinbeauvais: ... which is counterintuitive.

We handle that by displays "days left" and "expires tomorrow" in the Filings UI. Might need to do that here if it's not clear. Ask UX for guidance if needed.

src/components/YourCompany/YourCompany.vue
cameron-freshworks: We ref the confirm dialog and define it on the component with this type and i believe it exports from the interfaces.

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Hmm, well the template is guarded by a check around the property `nrNumber`,. 
That being said, if somehow these other values are not returned from the API i could re-write these template values into a getter which will fail gracefully if the data isn't present. 
Will update

src/interfaces/correction-interfaces/correct-name-interface.ts
cameron-freshworks: I believe we can, updated!

src/components/YourCompany/YourCompany.vue
severinbeauvais: Or add a v-if for the results?

src/interfaces/correction-interfaces/correct-name-interface.ts
cameron-freshworks: It's oddly named, from namex it is an object but i agree.. misleading as there is only one Name Request applicant.

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Oh i see. There is a v-if already around these results templates or do you mean to protect each individual display property. 
That being said, if condition was met to display these templates and somehow the data wasn't there, i think all that happens is the data isn't there.

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Hmm, yea. I will bring it up during the UX review. Potentially a verbiage change there.

src/interfaces/correction-interfaces/correct-name-interface.ts
severinbeauvais: Oh, I was thinking of names[].

Just leave applicants then.

src/components/YourCompany/YourCompany.vue
severinbeauvais: I didn't see if there was a v-if to verify `getNameRequest.applicant` (should be be "applicants"? lol).

I think you actually get a template error if, say, getNameRequest or getNameRequest.applicant are null.

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Fixed in last commit

src/components/YourCompany/YourCompany.vue
cameron-freshworks: ah sorry, not on that property specifically. Ok, will adjust!

If 'getNameRequest`  is null.. ie the store properties it references, the template shouldn't be rendered.

severinbeauvais: If you think it's well enough protected then leave it. I was just askin'.

cameron-freshworks: lol.. i don't think so, with regards to `applicant`.
That is the property we assign those values too, whereas `applicants` was from Namex and in the interface that gave definition to the response (obviously `applicants` being misleading because it only returns an object).

I'm not sure if it is normal practice to make an interface for expected responses but i thought what the heck in this case.

cameron-freshworks: oh no, i agree i could protect them a bit better. 
I added a getter which optional chains the Name Request before referencing applicant 👍

severinbeauvais: > I'm not sure if it is normal practice to make an interface for expected responses but i thought what the heck in this case.

It is good practise, if not normal. Re: schema !!!

src/components/YourCompany/YourCompany.vue
```diff
@@ -62,8 +126,36 @@
...
+          <v-flex xs3 class="mt-6">
+            <v-layout column>
+              <label><strong>Name Request Applicant</strong></label>
+            </v-layout>
+          </v-flex>
+          <v-flex xs6 class="mt-6">
+            <template>
+              <div class="name-request-applicant-info">
+                <span class="font-weight-bold">Name: </span>
+                <span>{{nrApplicant.fullName}}</span>
```
severinbeauvais, line 140: This could still fail if nrApplicant is null (per getter). 

One option is to do a v-if="nrApplicant" over this block.

Another option is to change the getter to:
```
return this.getNameRequest?.applicant || {}
```

severinbeauvais: Oh I see, you've initialized applicant.

As long as state.applicant that doesn't get overwritten with null, that should be OK.

severinbeauvais: So many ways to skin a cat...

severinbeauvais (approved):
Merge when you're satisfied.

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Did you want me to action anything here? 
As you mentioned, expiring at 00:00:00 on the date we are given would mean displaying the Date at GMT not PST, since after conversion it becomes the day prior due to the Offset in the given date.

cameron-freshworks: As it is, the expiry date shown will be the last day it's useable, if we don't convert to local then it would become NOT useable on the date displayed. 
Sorry if there was a conversation around this preference earlier 😁

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Your telling me lol
It's good though!

src/components/YourCompany/YourCompany.vue
severinbeauvais: Now I'm thoroughly confused. For the record.

severinbeauvais: As long as:
- the NR is expired as of 00:00:00 on its expiry date, and
- the NR expiry date is correctly displayed to the user
Then I think we're good.

PR #103, opened 2021-01-26
UX/Style Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#5102 & /bcgov/entity#5106 

*Description of changes:*

* Some style, UX additions and fixes.

* Updated Benefit Company help display, to only show after selection.

### DIFF
--- a/src/assets/styles/base.scss
+++ b/src/assets/styles/base.scss
@@ -144,21 +144,12 @@ p {
 // Common Text styles
 .subtitle {
   color: $gray9;
-  font-size: 1rem;
-  line-height: 1.5rem;
   font-weight: bold;
 }
 
 .info-text {
   color: $gray7;
-  font-size: 1rem;
-  line-height: 1.5rem;
   font-weight: normal;
-}
-
-.small-text {
-  font-size: .875rem !important;
-  line-height: 1.375rem !important;
 }
 
 // Loading Indicator
--- a/src/assets/styles/base.scss
+++ b/src/assets/styles/base.scss
@@ -156,6 +156,11 @@ p {
   font-weight: normal;
 }
 
+.small-text {
+  font-size: .875rem !important;
+  line-height: 1.375rem !important;
+}
+
 // Loading Indicator
 .loading-container {
   display: flex;
--- a/src/assets/styles/theme.scss
+++ b/src/assets/styles/theme.scss
@@ -59,6 +59,6 @@ $BCgovFontColorInverted: #ffffff;
 
 // App Colors
 $app-blue: #1669bb; // same as the Vuetify theme primary
-$p-blue-text: #38598a; // same as the Vuetify theme secondary
+$app-dk-blue: #38598a; // same as the Vuetify theme secondary
 $app-red: #d3272c; // same as the Vuetify theme error
 $app-green: #1a9031, // same as the Vuetify theme success
--- a/src/assets/styles/theme.scss
+++ b/src/assets/styles/theme.scss
@@ -59,6 +59,6 @@ $BCgovFontColorInverted: #ffffff;
 
 // App Colors
 $app-blue: #1669bb; // same as the Vuetify theme primary
-$app-dk-blue: #38598a; // same as the Vuetify theme secondary
+$app-dk-blue: #38598a; // same as the Vuetify theme app-dk-blue
 $app-red: #d3272c; // same as the Vuetify theme error
 $app-green: #1a9031, // same as the Vuetify theme success
--- a/src/assets/styles/theme.scss
+++ b/src/assets/styles/theme.scss
@@ -59,5 +59,6 @@ $BCgovFontColorInverted: #ffffff;
 
 // App Colors
 $app-blue: #1669bb; // same as the Vuetify theme primary
+$p-blue-text: #38598a; // same as the Vuetify theme secondary
 $app-red: #d3272c; // same as the Vuetify theme error
 $app-green: #1a9031, // same as the Vuetify theme success
--- a/src/components/YourCompany/CorrectBusinessType.vue
+++ b/src/components/YourCompany/CorrectBusinessType.vue
@@ -15,10 +15,10 @@
       <v-flex xs8 v-if="!isEditingType">
         <template>
           <span class="info-text"
-                :class="{ 'hasConflict': isConflictingLegalType && !isNewBusinessType}"
+                :class="{ 'hasConflict': isConflictingLegalType && isNewName}"
           >{{getEntityDesc(getEntityType)}}
           </span>
-          <v-tooltip v-if="isConflictingLegalType && !isNewBusinessType"
+          <v-tooltip v-if="isConflictingLegalType && isNewName"
                      top
                      content-class="top-tooltip"
                      transition="fade-transition"
@@ -31,13 +31,15 @@
               Business Types do not match. The Name Request type must match the business type before you can continue.
             </span>
           </v-tooltip>
-          <template v-if="isNewBusinessType && confirmArticles">
+          <template v-if="isNewBusinessType">
             <p class="subtitle mt-2 pt-2">Benefit Company Articles</p>
-            <p class="info-text">
-              <v-icon color="success">mdi-check</v-icon>
+            <div class="confirmed-msg">
+              <v-icon color="success" class="confirmed-icon">mdi-check</v-icon>
+              <span class="info-text small-text confirmed-icon ml-2">
               The company has completed a set Benefit Company Articles containing a benefit provision, and a copy of
               these articles has been added to the company's record book.
-            </p>
+              </span>
+            </div>
           </template>
         </template>
       </v-flex>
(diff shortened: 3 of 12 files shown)


src/assets/styles/base.scss
```diff
@@ -156,6 +156,11 @@ p {
   font-weight: normal;
 }
 
+.small-text {
```
severinbeauvais, line 159: Instead, how about overriding some of the existing Vuetify classes (so we can just the existing classes)?

See https://vuetifyjs.com/en/styles/text-and-typography/#typography

Re: https://github.com/bcgov/namerequest/blob/cd01435b6ddb4f003ebe1ace8593e7ec483c1ef8/client/src/assets/scss/overrides.scss#L13
See

src/assets/styles/theme.scss
```diff
@@ -59,5 +59,6 @@ $BCgovFontColorInverted: #ffffff;
 
 // App Colors
 $app-blue: #1669bb; // same as the Vuetify theme primary
+$p-blue-text: #38598a; // same as the Vuetify theme secondary
```
severinbeauvais, line 62: How about a more generic name, like "app-dk-blue"?

Eg: https://github.com/bcgov/namerequest/blob/cd01435b6ddb4f003ebe1ace8593e7ec483c1ef8/client/src/assets/scss/theme.scss#L55

src/components/YourCompany/CorrectBusinessType.vue
```diff
@@ -65,61 +67,65 @@
...
-        <!-- Help Section Toggle -->
-        <div class="info-text help-toggle pt-2" @click="helpToggle = !helpToggle">
-          <v-icon class="pr-2 mt-n2" color="primary">mdi-help-circle-outline</v-icon>
-          <span v-if="!helpToggle">Learn More</span>
-          <span v-else>Hide Learn More</span>
-        </div>
+        <template v-if="isBenefit">
+          <div class="my-6">
+            <p class="subtitle">Benefit Company Articles</p>
+            <p class="info-text">Before submitting your alteration notice you <span class="font-weight-bold">must
```
severinbeauvais, line 73: Keep in mind that we might want to centralize and index these articles when there are more entity types in the future.
No action needed.

severinbeauvais: And provisions, lol.

src/components/YourCompany/CorrectBusinessType.vue
```diff
@@ -65,61 +67,65 @@
...
+              </div>
+            </section>
+
+            <!-- Confirm Articles Checkbox -->
+            <div class="pr-2">
+              <v-checkbox
+                v-model="confirmArticles"
+                id="confirm-articles-checkbox"
+                :label="`The company has completed a set Benefit Company Articles containing a benefit provision, and a
+              copy of these articles has been added to company's record book.`"
```
severinbeauvais, line 124: You might to extract this to a property or getter, to avoid fancy quotes and multiple lines. (Multiple lines are affected by the CSS `white-space` property, and it may be hard to tell what that property value is here.)

src/components/YourCompany/CorrectBusinessType.vue
```diff
@@ -260,6 +273,16 @@ export default class CorrectBusinessType extends Mixins(CommonMixin) {
     return this.getEntityType && this.getEntityType !== this.getOriginalSnapshot[0]?.business?.legalType
   }
 
+  /** Verify New Business name. */
+  private get isNewName (): boolean {
+    return this.getApprovedName && this.getApprovedName !== this.getOriginalSnapshot[0]?.business?.legalName
```
severinbeauvais, line 278: Personally I like brackets to clarify the logic and operator precedence.

src/components/YourCompany/CorrectBusinessType.vue
```diff
@@ -283,6 +306,13 @@ export default class CorrectBusinessType extends Mixins(CommonMixin) {
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
 
+.help-section {
+  .subtitle, .info-text {
+    font-size: .875rem !important;
+    line-height: 1.375rem !important;
```
severinbeauvais, line 312: It would be nice to use existing classes, if they exist (and are suitably declared).

src/components/YourCompany/CorrectBusinessType.vue
```diff
@@ -300,6 +330,13 @@ ol {
   }
 }
 
+.confirmed-msg {
+  display: flex;
+  .confirmed-icon, .confirmed-note {
+    display: block;
```
severinbeauvais, line 336: Or `d-flex` and `d-block` ?

My rule of thumb is to use CSS if:
- it's not possible/clean to put classes inline
- the class applies to multiple (repeated) elements

No action required, just food for thought.

src/main.ts
```diff
@@ -69,6 +69,7 @@ async function start () {
         themes: {
           light: {
             primary: '#1669bb', // same as $app-blue
+            secondary: '#38598a', // Same as $p-blue-text
```
severinbeauvais, line 72: Normally "secondary" is a gray colour (as in, "it's not the primary action").
Re: https://vuetifyjs.com/en/features/theme/#customizing

Does this change break any existing use of secondary?

tests/unit/CorrectBusinessType.spec.ts
```diff
@@ -97,6 +97,10 @@ describe('CorrectBusinessType in an Alteration', () => {
   })
 
   it('displays the confirm articles checkbox and enables Done btn when selected', async () => {
+    // Set the selected entity to Benefit Company
+    wrapper.vm.selectedEntityType = 'BEN'
```
severinbeauvais, line 101: Hmm, normally we don't set VM properties this way.

Maybe use `wrapper.setData()` ?

https://vue-test-utils.vuejs.org/api/wrapper/setdata.html

severinbeauvais: And you might not need the `nextTick()`.

src/mixins/common-mixin.ts
```diff
@@ -127,4 +127,23 @@ export default class CommonMixin extends Vue {
...
+
+    // Check if the input is of correct length
+    let match = cleaned.match(/^(\d{3})(\d{3})(\d{4})$/)
+
+    if (match) {
+      return '(' + match[1] + ') ' + match[2] + '-' + match[3]
+    }
+
+    return null
+  }
```
severinbeauvais, line 148: Please add numerous unit tests to exercise this (which also serves as examples on how it works).

src/main.ts
cameron-freshworks: Not that i noticed, what do you think a good name would be for a non-clickable, darker blue.

src/assets/styles/base.scss
cameron-freshworks: I'll look into it, but if the scope strays too far out globally it will deserve it's own ticket.
I'm much over time on these two tickets with continual style/ux iterations.

This class is used in one instance that i know about for now, so i'll move it to it's local component until we know it's going to be something more common.

src/main.ts
severinbeauvais: "app-dk-blue" ?

It just seems like "secondary" may be used for something different.

This project isn't too bad for colour names, but the idea is to have a small list of colours ("theme").

Yes, the fact that it's for something clickable could be part of the name. Yes, "app-xxx" is very generic, possibly overly generic, but I can't think of a better colour framework atm.

src/assets/styles/theme.scss
cameron-freshworks: Updated the name to `app-dk-blue`

src/components/YourCompany/CorrectBusinessType.vue
cameron-freshworks: For sure, when we have entities / documents etc to support, 100%

src/components/YourCompany/CorrectBusinessType.vue
cameron-freshworks: Will do!

src/main.ts
severinbeauvais: > Not that i noticed

Did you search for other uses of "secondary". If not, you could hijack it I guess. (And update any other places that should use it.)

src/components/YourCompany/CorrectBusinessType.vue
cameron-freshworks: Copy that!

src/assets/styles/base.scss
cameron-freshworks: Updated in next commit. Followed your recommendation and overrode `text-body-3` to accommodate our small size here.

src/components/YourCompany/CorrectBusinessType.vue
cameron-freshworks: updated in next commit

src/components/YourCompany/CorrectBusinessType.vue
cameron-freshworks: Done and done, this is gone in next commit.

src/components/YourCompany/CorrectBusinessType.vue
cameron-freshworks: I'll give it a try

src/main.ts
cameron-freshworks: kk, updated in next commit

tests/unit/CorrectBusinessType.spec.ts
cameron-freshworks: We've done this lots!
Can try, but i think it's more or less the same thing, setting a property local to the component.

src/mixins/common-mixin.ts
cameron-freshworks: I test the format that is output in the unit tests. I can test multiple inputs though if you feel there is value there. 

Note about this: Phone numbers and formats are still in discussion i think due to international numbers etc, i mentioned to Scott i'd throw something in to display it with brackets, so i expect this to change.

tests/unit/CorrectBusinessType.spec.ts
cameron-freshworks: Updated!

src/mixins/common-mixin.ts
cameron-freshworks: Added some more tests

vysakh-menon-aot (approved):

severinbeauvais (approved):
Looks good!

PR #105, opened 2021-02-04
Alteration Summary Framework 

cameron-freshworks:
*Issue #:* /bcgov/entity#5172

Visual Design for reference: https://projects.invisionapp.com/d/main/#/console/20562985/436257303/preview

*Description of changes:*

* Implements Summary components for Alterations.

* Implemented Summary for Change of Name and Change of Business Type 
*Note:* These tasks were covered in tickets #5173 and #5175, but will leave those tickets open in the event there is additional requirements to be met with those components specifically.

* Added some global getters to track alterations.

* Added functionality around fee summary actions.

* Confirm Dialogs for Cancelling or Restoring Alterations.

* Updated error handling when unable to retrieve an existing alteration filing.

* Additional Unit tests.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -432,25 +432,29 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
     // open confirmation dialog and wait for response
     this.$refs.confirm.open(
-      'Unsaved Changes',
-      'This will cancel your alteration filing. Do you want to exit?',
+      'Cancel Filings',
+      'Any changes to your company information requiring a fee will be cancelled.',
       {
         width: '45rem',
         persistent: true,
-        yes: 'Return to my application',
+        yes: 'Cancel my changes',
         no: null,
-        cancel: 'Exit without saving'
+        cancel: 'Keep my changes'
       }
-    ).then(() => {
-      // if we get here, Yes was clicked
-      // nothing to do
-    }).catch(async () => {
+    ).then(async () => {
       // Delete the draft filing
       this.getFilingId && await this.deleteFilingById(this.getFilingId)
-
-      // redirect to dashboard
-      const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      window.location.assign(dashboardUrl + this.getBusinessId)
+        .then(() => {
+          // redirect to dashboard
+          const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
+          window.location.assign(dashboardUrl + this.getBusinessId)
+        })
+        .catch((error) => {
+          this.$root.$emit('save-error-event', error)
+        })
+    }).catch(() => {
+      // if we get here, No was clicked
+      // nothing to do
     })
   }
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -84,7 +84,7 @@
             <v-col cols="12" lg="3" style="position: relative">
 
               <!-- Corrections still uses the unmodified fee summary -->
-              <template v-if="showFeeSummary && isCorrection()">
+              <template v-if="showFeeSummary && isCorrectionView()">
                 <aside>
                   <affix
                     relative-element-selector=".col-lg-9"
@@ -114,7 +114,7 @@
 
         <!-- Action bar is for Corrections ONLY -->
         <actions
-          v-if="isCorrection()"
+          v-if="isCorrectionView()"
           :key="$route.path"
           @goToDashboard="goToDashboard()"
         />
@@ -553,8 +553,8 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
       const filing = await this.buildAlterationFiling(isDraft)
 
       // Update or file the alteration if we have a filingId or create a draft if not.
-      this.getFilingId
-        ? filingComplete = await this.updateFiling(filing, isDraft)
+      filingComplete = this.getFilingId
+        ? await this.updateFiling(filing, isDraft)
         : await this.createAlteration(filing, isDraft)
 
       // clear flag
--- a/src/App.vue
+++ b/src/App.vue
@@ -56,7 +56,7 @@
 
     <!-- Initial Page Load Transition -->
     <transition name="fade">
-      <div class="loading-container" v-show="!haveData">
+      <div class="loading-container" v-show="!haveData && !fetchErrorDialog">
         <div class="loading__content">
           <v-progress-circular color="primary" size="50" indeterminate />
           <div class="loading-msg">Loading</div>
@@ -104,7 +104,7 @@
                 :filing-data="getFilingData"
                 :pay-api-url="payApiUrl"
                 :isBusySaving="isBusySaving"
-                :hasConflicts="isConflictingLegalType"
+                :hasConflicts="isConflictingLegalType && hasBusinessNameChanged"
                 :isSummaryMode="isSummaryMode"
                 @action="handleSummaryActions($event)"
               />
@@ -176,9 +176,14 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   @Getter getFilingData!: FilingDataIF
   @Getter haveChanges!: boolean
   @Getter isBusySaving!: boolean
-  @Getter isConflictingLegalType!: boolean
   @Getter isFilingChanged!: boolean
   @Getter isEditing!: boolean
+  @Getter isSummaryMode!: boolean
+
+  // Alteration flag getters
+  @Getter hasAlterationChanges!: boolean
+  @Getter hasBusinessNameChanged!: boolean
+  @Getter isConflictingLegalType!: boolean
 
   // Global setters
   @Action setAccountInformation!: ActionBindingIF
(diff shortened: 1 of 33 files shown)


vysakh-menon-aot (approved):

src/App.vue
```diff
@@ -430,7 +433,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
     // open confirmation dialog and wait for response
     this.$refs.confirm.open(
       'Unsaved Changes',
-      'You have unsaved changes in your Incorporation Application. Do you want to exit?',
+      'This will cancel your alteration filing. Do you want to exit?',
```
severinbeauvais, line 436: Does this code also apply to corrections?

src/components/YourCompany/YourCompany.vue
```diff
@@ -81,7 +86,7 @@
             <div class="actions mr-4">
               <!-- TODO: only show buttons for named company -->
               <v-btn
-                v-if="companyNameChanges"
+                v-if="companyNameChanges || (isAlteration() && hasBusinessNameChanged)"
```
severinbeauvais, line 89: `isAlteration()` threw me for a minute.
I initially wondered why it wasn't a getter (but not a big deal).
Then I wondered what it means... I suggest this be renamed to "isAlterationPage" (or ...Route or ...View).

severinbeauvais (commented):
I only got part way through this. I'll finish on Monday.

src/App.vue
cameron-freshworks: This dialog does not.

src/mixins/legal-api-mixin.ts
```diff
@@ -64,6 +64,21 @@ export default class LegalApiMixin extends Mixins(FilingTemplateMixin) {
...
 
+  /**
+   * Deletes a filing by its id.
+   * @returns a promise to delete the filing or return a failure.
+   */
+  async deleteFilingById (id: number): Promise<any> {
+    const url = `businesses/${this.getBusinessId}/filings/${id}`
+
+    return axios.delete(url)
+      .then().catch(error => {
```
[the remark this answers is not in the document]
cameron-freshworks: ok cool, will remove!

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Will refactor a bit!

src/mixins/legal-api-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: Or perhaps this would make more sense here?
```
await axios.delete(url)
   .catch(/* whatever */)
return
```

It depends on how you want to handle the async / promise.

src/components/YourCompany/YourCompany.vue
cameron-freshworks: Updated in next commit 👍 
Changed all the instance of isAlteration and isCorrection to `isAlterationView`

src/App.vue
```diff
@@ -441,10 +444,10 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
     ).then(() => {
       // if we get here, Yes was clicked
       // nothing to do
-    }).catch(() => {
-      // if we get here, Cancel was clicked
-      // ignore changes
-      this.setHaveChanges(false)
+    }).catch(async () => {
+      // Delete the draft filing
+      this.getFilingId && await this.deleteFilingById(this.getFilingId)
```
severinbeauvais, line 449: What if they previously had a draft, and they just want to cancel the current changes?

cameron-freshworks: I'll follow up with Scott, i believe the UX pin described the Cancel, as clearing the filing. Will confirm. 
If the cancel is just to return to the dashboard, it becomes simpler, no DELETE request is required.

cameron-freshworks: Yep, can confirm in the UXpin, it does say to Cancel the Alteration filing.

severinbeauvais: So no draft alteration?

cameron-freshworks: Yea, `Resume Later` saves the draft and returns to the dashboard.

cameron-freshworks: The user can Under / Edit any changes within the filing, maybe that is given path for undoing changes, opposed to saving changes per session.

severinbeauvais: The problem I see is if the user saves a draft, then resumes it, then makes a change, then decides to abort that change... then it deletes their saved draft!!!

src/views/Alteration.vue
```diff
@@ -1,33 +1,54 @@
...
+      <detail class="mt-10" />
 
-    <staff-payment class="mt-10" />
+      <certify-section class="mt-10" />
+
+      <staff-payment class="mt-10" />
+    </template>
+
+    <!-- Summary View -->
+    <template v-else>
```
severinbeauvais, line 30: This code looks good :)

tests/unit/AlterationSummary.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+Vue.use(Vuetify)
+const localVue = createLocalVue()
+const vuetify = new Vuetify({})
+
+describe('AlterationSummary', () => {
+  let wrapper: any
+  let store: any = getVuexStore()
+
+  const originalSnapShot = [
```
severinbeauvais, line 21: Why is this an array?

src/mixins/legal-api-mixin.ts
```diff
@@ -64,6 +64,21 @@ export default class LegalApiMixin extends Mixins(FilingTemplateMixin) {
...
+   * @returns a promise to delete the filing or return a failure.
+   */
+  async deleteFilingById (id: number): Promise<any> {
+    const url = `businesses/${this.getBusinessId}/filings/${id}`
+
+    return axios.delete(url)
+      .catch(error => {
+      // eslint-disable-next-line no-console
+        console.log('deleteFilingById() error - invalid response =', error)
+        throw new Error('Invalid API response')
```
severinbeauvais, line 78: Should consider error handling in future ticket.

What should we do if deleting fails? Maybe just put up a dialog and let user proceed (just an OK button).

tests/unit/AlterationSummary.spec.ts
cameron-freshworks: The collection of the business snapshot is multiple requests, so opposed to our filings, it comes in some separate chunks before we break it down.

src/mixins/legal-api-mixin.ts
severinbeauvais: This could probably use the same mechanism as "save-error-event".

src/App.vue
severinbeauvais: OK for now.
(As discussed, Scott will review and think about it.)

tests/unit/AlterationSummary.spec.ts
severinbeauvais: Is it an array of like objects?

severinbeauvais: I just looked it up.
What does it mean to have multiple snapshots?

src/components/Summary/NoFeeSummary.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+  <v-card flat id="no-fee-summary" v-if="isSummaryMode">
+    <div class="summary-header">
+      <v-layout row class="mx-0">
+        <v-flex xs9>
+          <v-icon class="my-n1" color="#38598A">mdi-handshake</v-icon>
```
severinbeauvais, line 6: Future: we need to think about the Vuetify theme colours so we can use a name here instead of hex code.

severinbeauvais (approved):
The rest looks really nice. Good job 👍

src/mixins/legal-api-mixin.ts
cameron-freshworks: Copy that, will do!

cameron-freshworks: Updated in next commit 👍

src/components/Summary/NoFeeSummary.vue
cameron-freshworks: ahh right, will update this now!

src/mixins/legal-api-mixin.ts
cameron-freshworks: Actually, need to update this one more time.

cameron-freshworks: Need to create a new error event, as the other types don't fit here.

This is going to take some more work, will have to get it in the morning at this point, my day ends when the kids get home.

cameron-freshworks:
> Does this work? I thought it only accepted Vuetify names (eg, primary)?

Not 100% but it appears it will accept any name if we define it in the theme.

severinbeauvais:
> > Does this work? I thought it only accepted Vuetify names (eg, primary)?
> 
> Not 100% but it appears it will accept any name if we define it in the theme.
> Edit*: Yes, it does work.

Oh, nice. It means we are not limited to primary, etc. In fact we could leave the default theme intact and just add our names.

src/App.vue
```diff
@@ -302,6 +315,16 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
       this.nameRequestErrorDialog = true
     })
 
+    // listen for invalid delete requests
+    this.$root.$on('delete-error-event', async (error: any) => {
+      console.log('Delete error =', error) // eslint-disable-line no-console
+      this.saveErrors = error?.response?.data?.errors || []
+      this.saveWarnings = error?.response?.data?.warnings || []
```
severinbeauvais, line 321: In future ticket, fix these variabes :)

cameron-freshworks: Ahh yes 🤦 . Will put in a small commit later today!

PR #108, opened 2021-02-24
5025 Contact Info Integration

cameron-freshworks:
*Issue #:* /bcgov/entity#5025

*Description of changes:*

* Integrated Contact info shared component into the Edit UI for Corrections and Alterations.

* Updated noFeeSumarry component.

* Added auth-api mixin to handle auth requests.

* Updated Unit testing.


src/mixins/auth-api-mixin.ts
```diff
@@ -0,0 +1,63 @@
...
+        phone: contactInfo.phone,
+        phoneExtension: contactInfo.extension
+      }
+      const authUrl = sessionStorage.getItem('AUTH_API_URL')
+      let url = `${authUrl}entities/${this.getBusinessId}/contacts`
+
+      return axios.put(url, contactPayload)
+    } catch (error) {
+      // TODO: Update error handling modal after design review
+      this.$root.$emit('save-error-event', error)
```
cameron-freshworks, line 60: There is no current design for a modal here. I'm assuming there will be one, so used the generic Save-error in the meantime. 
This modal is subject to change.

src/store/getters/state-getters.ts
```diff
@@ -344,3 +344,13 @@ export const hasBusinessNameChanged = (state: StateIF): boolean => {
 export const hasBusinessTypeChanged = (state: StateIF): boolean => {
   return state.stateModel.tombstone.entityType !== state.stateModel.originalSnapshot[0]?.business?.legalType
 }
+
+/** Check for changes between current contact and original contact. */
+export const hasContactInfoChange = (state: StateIF, rootState: any): boolean => {
+  const businessContact: BusinessContactIF = state.stateModel.defineCompanyStep.businessContact
+  const originalSnapshot: any = state.stateModel.originalSnapshot[5]
```
cameron-freshworks, line 351: @severinbeauvais I could use help here! 
I'm still having troubles defining a type on a specific index of an Array interface. I tried a few things and fell short. Any suggestions are welcome :)

package.json
```diff
@@ -14,10 +14,11 @@
   "dependencies": {
     "@babel/compat-data": "^7.8.6",
     "@bcrs-shared-components/certify": "^1.0.3",
+    "@bcrs-shared-components/contact-info": "1.0.3",
```
severinbeauvais, line 17: Please bump up the app patch version.

src/components/YourCompany/BusinessContactInfo.vue
```diff
@@ -1,293 +1,77 @@
...
-  </div>
+  <contact-info
+    :businessContact="getBusinessContact"
+    :originalBusinessContact="originalContact"
+    :hasBusinessContactInfoChange="hasBusinessContactInfoChange"
+    :edit-label="editLabel"
+    :edited-label="editedLabel"
+    @contactInfoChange="setContact($event)"
+    @haveChanges="emitHaveChanges($event)"
+  />
```
severinbeauvais, line 10: I noticed, while testing Contact Info in Storybook, that anew email address isn't "saved" to the read-only version. Does it work here?

package.json
cameron-freshworks: 1.0.3 is the latest! or do you mean something else?

src/components/YourCompany/BusinessContactInfo.vue
cameron-freshworks: Good point, the values displayed in the read only version are straight from store.
So in storybook changes made aren't reflected there since those values are not stored local to the component, although there could be a way to better mock this in storybook.

It does work when we are able to update the values we pass into the Shared Component, you bet.

severinbeauvais: Maybe the Storybook component needs a handler for `@contactInfoChange`.
(Maybe fix it next time we're in there?)

severinbeauvais: Maybe we could get the change notification working in SB also :)

src/components/YourCompany/BusinessContactInfo.vue
```diff
@@ -1,293 +1,77 @@
...
+
+  @Getter getBusinessContact!: BusinessContactIF
+  @Getter getOriginalIA!: IncorporationFilingIF
+  @Getter getOriginalSnapshot!: BusinessSnapshotIF
+
+  /** Get the original Contact info dependant on filing type. */
+  private get originalContact (): BusinessContactIF {
+    return this.isCorrectionView()
+      ? this.getOriginalIA.incorporationApplication.contactPoint
+      : this.getOriginalSnapshot[5]
```
severinbeauvais, line 48: I recommend a getter for this to abstract which array item this data is.

(ATM this requires too much knowledge of how snapshots work.)

src/components/YourCompany/BusinessContactInfo.vue
```diff
@@ -1,293 +1,77 @@
...
 
-  // Watchers
-  @Watch('businessContact', { deep: true, immediate: true })
-  private onContactPropValueChanged (): void {
-    this.contact = { ...this.businessContact }
-    this.emitHaveChanges(this.hasBusinessContactInfoChange)
+  /** Request to update contact info in Auth and Store. */
+  private async updateContactRequest (contactInfo: BusinessContactIF): Promise<void> {
+    await this.updateContactInfo(contactInfo)
+    this.setBusinessContact(contactInfo)
```
severinbeauvais, line 68: Should we await store setters? In NR some effort was made to handle the supposedly async store setters but in our other UIs we've never worried about this and I don't think we've seen any problems. Thoughts?

src/mixins/auth-api-mixin.ts
```diff
@@ -0,0 +1,63 @@
+// Libraries
+import { Component, Vue } from 'vue-property-decorator'
+import { Getter } from 'vuex-class'
+import { axios } from '@/utils'
+
+// Interfaces
+import { BusinessContactIF } from '@/interfaces'
+/**
+ * Mixin that provides integration with the Legal API.
```
severinbeauvais, line 9: Auth API

src/mixins/auth-api-mixin.ts
```diff
@@ -0,0 +1,63 @@
...
+    try {
+      const contactPayload = {
+        email: contactInfo.email,
+        phone: contactInfo.phone,
+        phoneExtension: contactInfo.extension
+      }
+      const authUrl = sessionStorage.getItem('AUTH_API_URL')
+      let url = `${authUrl}entities/${this.getBusinessId}/contacts`
+
+      return axios.put(url, contactPayload)
```
severinbeauvais, line 57: Wait, so we GET contact info from auth API but we PUT it to the legal API? Should the update method go in legal-api-mixin then?

src/mixins/auth-api-mixin.ts
severinbeauvais: Looks ok for now. 👍

src/store/getters/state-getters.ts
severinbeauvais: You can cast with "as".

Sometimes you have to cast to "unknown" first, then cast to target IF.

severinbeauvais: And instead of sprinkling this snapshot[x] code everywhere, this is another reason for global getters for it.

tests/unit/CorrectBusinessContactInfo.spec.ts
```diff
@@ -0,0 +1,129 @@
+// Libraries
```
severinbeauvais, line 1: Probably need to rename this, since you deleted the actual component by this name.

tests/unit/CorrectBusinessContactInfo.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    wrapper.destroy()
+  })
+
+  it('renders the CorrectBusinessContactInfo Component', async () => {
+    expect(wrapper.find(BusinessContactInfo).exists()).toBe(true)
+  })
+
+  it('loads the correct original contact info for a correction', async () => {
+    expect(wrapper.vm.originalContact.email = originalCorrectionContact.email)
+    expect(wrapper.vm.originalContact.phone = originalCorrectionContact.phone)
```
severinbeauvais, line 59: You should be able to use `expect(blah).toEqual()`.

tests/unit/CorrectBusinessContactInfo.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+  it('watches for business contact info changes ', async () => {
+    // Verify there is a diff between current and original contact data
+    expect(wrapper.vm.hasBusinessContactInfoChange).toBe(true)
+
+    // Call the set Contact method and set the data back to it's original
+    wrapper.vm.setContact(originalCorrectionContact)
+    await Vue.nextTick()
+
+    // Verify there is NO diff between current and original contact data
+    expect(wrapper.vm.hasBusinessContactInfoChange).toBe(false)
```
severinbeauvais, line 71: If it's quick, please also check for emitted events.

package.json
severinbeauvais: line 3

src/mixins/auth-api-mixin.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    return axios.get(url, config)
+      .then(response => {
+        if (response && response.data && response.data.contacts) {
+          // Always take the first contact.
+          return {
+            ...response.data.contacts[0],
+            confirmEmail: response.data.contacts[0].email,
+            extension: response.data.contacts[0].phoneExtension
+          } || []
```
andrepestana-aot, line 34: return the object or the empty array?

src/store/getters/state-getters.ts
severinbeauvais: Try `const originalSnapshot: OriginalSnapshotIF: state.stateModel.originalSnapshot[5] as unknown`.
(untested)

package.json
cameron-freshworks: ah, sorry. I misunderstood, thought you were referring to this import.

src/App.vue
```diff
@@ -65,7 +65,7 @@
 
     <!-- Initial Page Load Transition -->
     <transition name="fade">
-      <div class="loading-container" v-show="!haveData && !fetchErrorDialog">
+      <div class="loading-container" v-show="(!haveData && !fetchErrorDialog)">
```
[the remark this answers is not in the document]
cameron-freshworks: No difference! 
In fact, i was testing a spinning that required an extra condition but decided against it.
Will clean this up, thanks!

src/components/YourCompany/BusinessContactInfo.vue
cameron-freshworks: Yes and yes, absolutely will do following this PR.

I made some changes so that the read only portion of the component wasn't local data and this was a byproduct of that, i didn't notice until now!

src/components/YourCompany/BusinessContactInfo.vue
cameron-freshworks: Fair point! Will add an getter a little more explicit to what we are `get`'ing here

src/components/YourCompany/BusinessContactInfo.vue
cameron-freshworks: The asynchronous call isn't setting anything to store itself, it is making the request to update the information in Auth. 
I chose to set to store after that request in the event it fails, so we don't have data in the UI that is out of sync (behind) what we have in Auth.

src/mixins/auth-api-mixin.ts
cameron-freshworks: Thanks :)
Fixed!

src/components/YourCompany/BusinessContactInfo.vue
cameron-freshworks: Updated in next commit 👍

src/components/YourCompany/BusinessContactInfo.vue
severinbeauvais: No, the line I commented on
`this.setBusinessContact(contactInfo)`
is a store setter, right?

src/mixins/auth-api-mixin.ts
cameron-freshworks: Good point, i believe we are deconstructing down to `contacts` which itself is an array but since we are always returning the first object in that array it will never return an array under normal circumstances. 
I will look into this!

src/mixins/auth-api-mixin.ts
cameron-freshworks: The other way around! 
But gah.. you're right, i moved the GET to this mixin because i mistook it as an Auth request.. 🤦 
I will move the GET back to the Legal Api Mixin.

src/store/getters/state-getters.ts
cameron-freshworks: I actually did try unknown at some point, at which it worked. I wasn't sure of it's purpose though, so didn't settle on it. 

i've had mixed results with using global getters within other global getters, 
I can give it another try though!

src/components/YourCompany/BusinessContactInfo.vue
cameron-freshworks: Yes, `this.setBusinessContact(contactInfo)` is a store setter. 
You think this could be an issue, setting after waiting for a successful request?

src/store/getters/state-getters.ts
cameron-freshworks: oo i see how this is done now 😁 
Very cool, updated in next commit!

severinbeauvais: All getters should read from `state` right?
Example: https://github.com/bcgov/business-filings-ui/blob/84a2fbe0f8d8a3b350edba0cc7934fb63ce65d87/src/store/getters.ts#L6

tests/unit/CorrectBusinessContactInfo.spec.ts
cameron-freshworks: Shoot, just needs to be renamed to `BusinessContactInfo.spec`

tests/unit/CorrectBusinessContactInfo.spec.ts
cameron-freshworks: Will do!

src/components/YourCompany/BusinessContactInfo.vue
severinbeauvais: I think the NR guys were concerned enough about store setters that they awaited those calls.

Meanwhile, we never have and we seem to be OK.

So I dunno. Is it an issue? Is it not an issue? 🤷‍♂️

cameron-freshworks: Yea, ditto, we've never had an issue. Could be the way we've architected the setters (actions), and bound them. 
This isn't a problem i don't think.

cameron-freshworks: They `awaited` ALL their store setters? I don't recall this.. but that certainly seems fishy. We are just defining a key's value essentially, what would we need to wait for 🤔 

Perhaps, since they had a lot of sequential coding in the modules which would dictate next steps that would require those aforementioned values to be present, they had to ensure these values set before proceeded? 

Sounds like a UI not being in sync with the store kinda nightmare lol

src/store/getters/state-getters.ts
cameron-freshworks: Definitely, i'm just a goofball. Passing the state on through to the global getter works great!
Also sorted out the typing :)
Check back in next commit!

tests/unit/CorrectBusinessContactInfo.spec.ts
cameron-freshworks: > Shoot, just needs to be renamed to `BusinessContactInfo.spec`

Done in next commit!

src/components/YourCompany/BusinessContactInfo.vue
severinbeauvais: Sounds like what we've been seeing in the Sentry logs.

OK, no change neede here.

tests/unit/CorrectBusinessContactInfo.spec.ts
cameron-freshworks: Updated!

package.json
```diff
@@ -14,10 +14,11 @@
...
     "@bcrs-shared-components/detail-comment": "^1.0.4",
     "@bcrs-shared-components/enums": "^1.0.0",
     "@bcrs-shared-components/fee-summary": "^1.0.1",
-    "@bcrs-shared-components/interfaces": "^1.0.2",
+    "@bcrs-shared-components/interfaces": "^1.0.10",
```
severinbeauvais, line 21: Interfaces is up to 11 now.

src/store/getters/state-getters.ts
```diff
@@ -218,6 +218,11 @@ export const getBusinessContact = (state: StateIF): BusinessContactIF => {
   return state.stateModel.defineCompanyStep.businessContact
 }
 
+export const getSnapshotContact = (state: StateIF): BusinessContactIF => {
+  const snapShotContact: unknown = state.stateModel.originalSnapshot[5]
+  return snapShotContact as BusinessContactIF
```
severinbeauvais, line 223: 👍

severinbeauvais: Are you going to fix the other snapshot[] instances with getters?
Or next ticket?

package.json
cameron-freshworks: kk, will update when i bump the Contact version.

src/store/getters/state-getters.ts
cameron-freshworks: I think as we go. 
When the other components in Alterations are needing the specific original values for a diff, that might be a good time to add them.

severinbeauvais: ok

cameron-freshworks:
> Alterations are not available at the moment. Please check again later.

Good call, updated!

src/components/Summary/NoFeeSummary.vue
```diff
@@ -37,7 +39,7 @@
     </template>
 
     <!-- Folio Number -->
-    <template>
+    <template id="no-fee-folio">
```
severinbeauvais, line 42: What this id used for?

severinbeauvais (approved):
Looks really good. I like it :)

src/components/Summary/NoFeeSummary.vue
cameron-freshworks: Nothing yet, but will be for unit testing once the folio piece is done. Just put it in now, so it followed the same id pattern as the other 😁

cameron-freshworks:
> Looks really good. I like it :)

thanks!

andrepestana-aot:
Looks good to me too! :)

andrepestana-aot (approved):

PR #112, opened 2021-03-03
5027 display adressess in alteration formatting

andrepestana-aot:
*Issue #:* /bcgov/entity#5027

*Description of changes:*

- Fixed formatting as UX Assurance requested
- Hid Records Office addresses for COOP business type as BAs requested


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.5.7",
+  "version": "0.5.8",
```
severinbeauvais, line 3: You also need to update (and commit) the package-lock file.

cameron-freshworks (approved):
Looks goods!
Was there any unit tests for the conditions at which the records address displayed?
It could be the unit tests for Addresses were skipped when we ported this component over. 
I would have a look there and update/fix any unit tests that might be applicable :)

severinbeauvais (approved):
GTG after comment is resolved.

package.json
andrepestana-aot: ok. Done!

andrepestana-aot:
> Looks goods!
> Was there any unit tests for the conditions at which the records address displayed?
> It could be the unit tests for Addresses were skipped when we ported this component over.
> I would have a look there and update/fix any unit tests that might be applicable :)

No, I didn't add any for this.

andrepestana-aot:
> GTG after comment is resolved.

Thanks!

cameron-freshworks (commented):
Something is up with the package-lock.json, causing the install to fail.
I think* the only diff should be the version bump after an `npm i`, should have a look at that before merge 👍

package.json
severinbeauvais: @andrepestana-aot Let the person who created a conversation resolve it -- in my case it's my checklist to verify that the thing I commented on is fixed.

If you want to indicate that you're read or handled this thing, you can add a 👍  to the comment.

andrepestana-aot: ok

jachurchill (approved):

andrepestana-aot:
> Something is up with the package-lock.json, causing the install to fail.
> I think* the only diff should be the version bump after an `npm i`, should have a look at that before merge 👍

@cameron-freshworks  Yes, you are right. Thanks! Could you publish it on dev, please?

PR #114, opened 2021-03-05
Added new component for current directors as readonly. Based on the P…

jachurchill:
…eople and Roles component

*Issue #:* /bcgov/entity#5030

*Description of changes:* Added new component for current directors as readonly. Based on the People and Roles component	a467829	John Churchill <john@highwaythreesolutions.com>	Mar 5, 2021 at 9:19 AM


src/components/CurrentDirectors/index.ts
```diff
@@ -0,0 +1 @@
+export { default as CurrentDirectors } from './CurrentDirectors.vue'
```
severinbeauvais, line 1: I think, if we go down this route (folders for individual components), that we will end up with a lot of folders. There are least 2 options here:
- generalize this folder name so we can reuse it later (eg, "Directors")
- put this component in an existing, related folder (ie, "PeopleAndRoles")

Thoughts?

severinbeauvais:
I'd like to see the app version increased by 0.0.1 with every new PR.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+  private mounted (): void {}
+
+  /**
+   * Returns the current directors from the people and roles.
+   * @returns array of directors
+   */
+  private get currentDirectors () : OrgPersonIF[] {
+    const directors = this.getPeopleAndRoles
+      .filter(people => people.roles.some(role =>
+        role.roleType.toLowerCase() === 'director' && role.cessationDate === null))
```
severinbeauvais, line 103: Please use enum for role type.

severinbeauvais: Can cessation date be '' instead of null? If so, consider checking if falsey instead.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+  // Global getters
+  @Getter getPeopleAndRoles!: OrgPersonIF[]
+
+  /** Headers for the person table. */
+  readonly tableHeaders = ['Name', 'Mailing Address', 'Delivery Address', 'Effective Dates']
+
+  /**
+   * Called when component is mounted.
+   */
+  private mounted (): void {}
```
severinbeauvais, line 94: You can delete this as it's not used.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+      </div>
+
+      <!-- Instructional Text -->
+      <div class="instructional-text pt-10 pb-15 px-4">
+        To change directors, please use the Change feature in the Current Directors list on you business dashboard.
+      </div>
+
+      <div class="list-container px-4">
+        <!-- List Display Section -->
+        <div id="people-roles-list" v-if="currentDirectors.length > 0">
```
severinbeauvais, line 18: No need to change this atm, but if you have a code section, consider using the HTML5 semantic elements (eg, "section").

https://www.google.com/search?q=html5+semantic+elements

This helps with accessibility, as various tools parse the document and can identify headers/sections/footers/etc to the user.

Basically, if you see an obvious opportunity to use semantic elements instead of divs, go for it.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+
+    <v-card flat>
+      <!-- Header -->
+      <div class="header-container">
+        <v-icon color="app-dk-blue">mdi-account-multiple-plus</v-icon>
+        <label class="font-weight-bold pl-2">Current Directors</label>
+      </div>
+
+      <!-- Instructional Text -->
+      <div class="instructional-text pt-10 pb-15 px-4">
```
severinbeauvais, line 12: We have found it easier to not use bottom margin/padding for spacing purposes. Instead, the element below this can add the necessary margin/padding.

This may be more of a convention (eg, either select upper or lower element to do the spacing), but if we pick one way then it's a lot cleaner and easier to update later.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+            v-for="(orgPerson, index) in currentDirectors"
+            :key="index"
+            no-gutters
+          >
+            <!-- Name + Badge -->
+            <v-col class="text-truncate pr-5">
+              <!-- provide tooltip to display full name if name is longer than 25 chars -->
+              <v-tooltip top :disabled="formatFullName(orgPerson.officer).length < 25" color="primary">
+                <template v-slot:activator="{ on }">
+                  <span v-on="on" class="director-name">{{ formatFullName(orgPerson.officer).toLowerCase() }}</span>
```
severinbeauvais, line 38: Changing director name to lower case doesn't seem right. Do you have a spec/design that requests this?

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+            <v-col class="pr-5">
+              <p v-if="isSame(orgPerson.mailingAddress, orgPerson.deliveryAddress)"
+                class="director-detail">Same as Mailing Address
+              </p>
+              <base-address v-else class="director-detail" :address="orgPerson.deliveryAddress"/>
+            </v-col>
+
+            <!-- Appoinment Date -->
+            <v-col class="pr-5">
+              <span class="director-detail">{{ orgPerson.roles[0].appointmentDate }} to Current</span>
```
severinbeauvais, line 59: What if the person has multiple roles, each with a different appointment  date?

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+
+          <!-- List Content -->
+          <v-row
+            class="people-roles-content py-3"
+            v-for="(orgPerson, index) in currentDirectors"
+            :key="index"
+            no-gutters
+          >
+            <!-- Name + Badge -->
+            <v-col class="text-truncate pr-5">
```
severinbeauvais, line 34: How wide are each of these v-cols?

How does this look when the screen is narrowed? We don't support mobile very well atm but we do try to make our columns a bit responsive.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+      <!-- Instructional Text -->
+      <div class="instructional-text pt-10 pb-15 px-4">
+        To change directors, please use the Change feature in the Current Directors list on you business dashboard.
+      </div>
+
+      <div class="list-container px-4">
+        <!-- List Display Section -->
+        <div id="people-roles-list" v-if="currentDirectors.length > 0">
+          <!-- List Headers -->
+          <v-row class="people-roles-list-header list-item__subtitle pb-3" no-gutters>
```
severinbeauvais, line 20: I see you're adding right padding to your columns.

Try removing the "no-gutters" directive and see if that automatically adds in the desired spacing.

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: Also on line 105 below.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+  padding: 1.25rem;
+}
+
+[class^="col"] {
+  padding-top: 0;
+  padding-bottom: 0;
+}
+
+.instructional-text {
+  font-size: 16px;
```
severinbeauvais, line 127: Except for fine-tuning spacing (eg, a few px here or there), we specify sizes in rems.

https://www.google.com/search?q=why+use+rem+instead+of+px

PS 16px = 1 rem

PPS You should try to use an existing style for this instructional text. See:
- [the Vuetify built in styles](https://vuetifyjs.com/en/styles/text-and-typography/#typography)
- https://github.com/bcgov/business-edit-ui/blob/master/src/assets/styles/base.scss
- https://github.com/bcgov/business-edit-ui/blob/master/src/assets/styles/overrides.scss

severinbeauvais: This comment applies to the other font sizes in this file, and to the other classes you're using (better to not have one-time classes here esp. when existing classes probably provide the styles you need).

src/components/CurrentDirectors/index.ts
jachurchill: I'll add it to people and roles, when I started work I thought it was gonna be more involved and use multiple files like people and roles so I duplicated it. But with how simplified it is it really doesn't benefit

src/components/CurrentDirectors/CurrentDirectors.vue
jachurchill: So I did that initially but the enum gives me the value as 'Director' but on the role I get 'director' so they weren't equating. I could compare to RoleType.DIRECTOR.toLowerCase() if that would be better. Or do you have a better solution?

src/components/CurrentDirectors/CurrentDirectors.vue
jachurchill: I have css that changes it to capitalize. The stored data, at least that I was testing with was in all caps so this was to get it to match the design https://projects.invisionapp.com/share/6EYT1CDGYVN#/screens/445576065/comments which shows it as capitalized

src/components/CurrentDirectors/CurrentDirectors.vue
jachurchill: We are only showing directors, I filtered it below to only include their role as director. It's not possible to have multiple director roles is it? Not sure how this would work if that's the case

src/components/CurrentDirectors/CurrentDirectors.vue
jachurchill: I based this off the table for People and Roles so it's setup the exact same way I believe

src/components/CurrentDirectors/CurrentDirectors.vue
jachurchill: The design specifically said it wants a minimum of 20px between columns so I thought padding was a more effective way to get that

src/components/CurrentDirectors/CurrentDirectors.vue
jachurchill: Ok the design had all it's sizes in px, I'll take a look at those style though

src/components/CurrentDirectors/CurrentDirectors.vue
jachurchill: In regard to the cessation date it was null on the one I was testing but I'll change it to just check for a falsy value as I think that's more robust

src/components/CurrentDirectors/index.ts
severinbeauvais: Perfect.

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: > So I did that initially but the enum gives me the value as 'Director' but on the role I get 'director' so they weren't equating. I could compare to RoleType.DIRECTOR.toLowerCase() if that would be better. Or do you have a better solution?

How did it get set to "director" ?

Do not take an enum and lower-case it. Maybe the enum should be a lower-case value -- depends how the code uses it (ie, where type is set for the current role).

severinbeauvais: > In regard to the cessation date it was null on the one I was testing but I'll change it to just check for a falsy value as I think that's more robust

Yeah, depends how it's initialized or reset. We're not totally consistent on initializing state string to '' instead of null. See https://github.com/bcgov/business-edit-ui/blob/master/src/store/state/state-model.ts

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: So the design shows it as title caps.

The sample data you have may be all-caps from the db... but don't assume that and convert it.

If you have CSS that capitalizes it, why are you converting it to lower case here?

Bottom line: display what you get from the db/API. Otherwise you're likely to get it wrong, eg. "McDonald" don't display right if you title-case it. That might cause a user to want to fix it, but it would still display wrong.

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: Sounds good. Thanks.

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: OK I looked at the tables used in People and Roles and I don't see special v-col spacing. Please fix it.

As for the col width, you're right, it's not specified on the other page either. I dunno. If it matches the ux design then you're good.

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: I just noticed this comment is attached to the wrong line number (should be one of the lines with v-col with "pr-5".

I still think you can get the right spacing with the built-in gutters.

Anyway, this is not a deal-breaker. LMK if gutters don't work and we'll leave this as is.

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: I know :( I have asked them to specify rems many times but ...

Anyway, it's easy for us to convert px -> rem.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+            <v-col class="pr-5" v-for="(title, index) in tableHeaders" :key="index">
+              <span class="directors-title">{{ title }}</span>
+            </v-col>
+          </v-row>
+
+          <!-- List Content -->
+          <v-row
+            class="people-roles-content py-3"
+            v-for="(orgPerson, index) in currentDirectors"
+            :key="index"
```
cameron-freshworks, line 30: Not 100% but this key could create some duplicate identifiers if there is another mounted instance of `:key=index`.
An option to preserve the uniqueness would be template string a descriptor prefix to the index, like `director:${index}` or something along those lines.

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+            no-gutters
+          >
+            <!-- Name + Badge -->
+            <v-col class="text-truncate pr-5">
+              <!-- provide tooltip to display full name if name is longer than 25 chars -->
+              <v-tooltip top :disabled="formatFullName(orgPerson.officer).length < 25" color="primary">
```
cameron-freshworks, line 36: No change required here, but when the logic in the template goes beyond the most basic, its never a bad idea to create a local getter/method etc to handle these conditions.
 Especially if the logic is being used elsewhere 👍

src/components/CurrentDirectors/CurrentDirectors.vue
```diff
@@ -0,0 +1,157 @@
...
+
+.director-name {
+  font-size: 16px;
+  text-transform: capitalize;
+  color: $gray9;
+  font-weight: bold;
+}
+
+.director-detail {
+  font-size: 14px;
```
cameron-freshworks, line 154: Could be converted to .875rem

cameron-freshworks (commented):
Looks good, just had some questions/comments. 

Just curious, since this appears to be a net-new component, how far does it differ from the existing `people-and-roles`?

Only concern is 2 very similar components to maintain, and having just gone through a battle making a large component a shared component, this could be a problem if we decide to make this shared at some point.

src/components/CurrentDirectors/CurrentDirectors.vue
jachurchill: I was converting to lowercase because CSS capitalize doesn't convert the rest of the word to lower, but I see your point here, I will remove that and trust the db

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: Yes, please specify font sizes in rem.
1 rem = 16px

https://www.google.com/search?q=why+use+rem+instead+of+px

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: If, some other time, you do need to fully uppercase a word using CSS, you can use `text-transform: uppercase;`.

jachurchill:
> Looks good, just had some questions/comments.
> 
> Just curious, since this appears to be a net-new component, how far does it differ from the existing `people-and-roles`?
> 
> Only concern is 2 very similar components to maintain, and having just gone through a battle making a large component a shared component, this could be a problem if we decide to make this shared at some point.

@cameron-freshworks it's a very dumbed down version with different columns and some extra specification of what people to display. I started off duplicating the existing component but ended up scrapping almost all of that. It's definitely possible to make this using the existing component but it may complicate it unnecessarily, not sure. I'm happy to change that if you and @severinbeauvais think that's a better solution

cameron-freshworks:
> > Looks good, just had some questions/comments.
> > Just curious, since this appears to be a net-new component, how far does it differ from the existing `people-and-roles`?
> > Only concern is 2 very similar components to maintain, and having just gone through a battle making a large component a shared component, this could be a problem if we decide to make this shared at some point.
> 
> @cameron-freshworks it's a very dumbed down version with different columns and some extra specification of what people to display. I started off duplicating the existing component but ended up scrapping almost all of that. It's definitely possible to make this using the existing component but it may complicate it unnecessarily, not sure. I'm happy to change that if you and @severinbeauvais think that's a better solution

I'm ok with it as it is, you are definitely right that the components can get complex the more conditions we add to them. I guess as long as this explicitly READ-ONLY or something along those lines. 
Just don't want to end up with two components that have similar functionality, with some styling differences 😁

severinbeauvais:
> I'm ok with it as it is, you are definitely right that the components can get complex the more conditions we add to them. I guess as long as this explicitly READ-ONLY or something along those lines.
> Just don't want to end up with two components that have similar functionality, with some styling differences 😁

I'm OK either way on this one. Sounds like the simpler approach was taken and that's generally a good thing :)

src/components/PeopleAndRoles/CurrentDirectors.vue
```diff
@@ -88,21 +87,16 @@ export default class CurrentDirectors extends Mixins(CommonMixin) {
...
-
   /**
    * Returns the current directors from the people and roles.
    * @returns array of directors
    */
   private get currentDirectors () : OrgPersonIF[] {
     const directors = this.getPeopleAndRoles
       .filter(people => people.roles.some(role =>
-        role.roleType.toLowerCase() === 'director' && role.cessationDate === null))
+        role.roleType.toLowerCase() === RoleTypes.DIRECTOR.toLowerCase() && !role.cessationDate))
```
severinbeauvais, line 97: I am still not OK with this.

How did `roleType` get assigned and why is it not the same value (string) as the RoleTypes enum?

Does it come from the API, and need to be converted to the enum when the data is fetched?

jachurchill: This was the easiest way to make the comparison case insensitive. I'll take a look at the data and see if it's just the test data that is different. I didn't want to touch the enum itself because I figured that might break something

severinbeauvais: The comparison should just work like this:
`role.roleType === RoleTypes.DIRECTOR`

... at least, as long at `role.roleType` is properly declared as type RolesTypes and wasn't assigned a value that isn't of that type.

jachurchill: That was the first thing I did when I started working on it but at least for the business I've been testing with that doesn't work. The field is typed as a RoleType but the content doesn't match that. Just investigating now to see where that data is coming from to see

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: This looks like the only v-for on this page so there's no identifier conflict.

However, I agree with Cam that it's a good idea to specify a key value that is more than just the index (see Cam's example).

This is not an issue atm so I'm resolving this conversation.

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: I agree with Cam. Please keep this in mind for future code changes.

This is not an issue atm so I'm resolving this conversation.

src/components/PeopleAndRoles/CurrentDirectors.vue
severinbeauvais: Thanks for the fix.

severinbeauvais:
Please resolve conflicts and then I'll merge this PR :)

src/components/CurrentDirectors/CurrentDirectors.vue
severinbeauvais: Please fix font sizes in CurrentDirectors.vue. See lines 126 and 131.

severinbeauvais: @jachurchill ^^

severinbeauvais (approved):

PR #115, opened 2021-03-05
5103 pre existing company provisions

andrepestana-aot:
*Issue #:* /bcgov/entity#5103

*Description of changes:*

- Created Articles component
- Created Company Provisions component
- WIP - some tests coming


andrepestana-aot:
I'm still working on some unit tests.

severinbeauvais:
Please fix link to ticket in description ^^.

Is this ready for review or still a work in progress?

severinbeauvais:
I'd like to see the app version increased by 0.0.1 with every new PR.

src/components/Articles/Articles.vue
```diff
@@ -0,0 +1,78 @@
...
+export default class Articles extends Mixins(CommonMixin) {
+  // whether components have changes
+  private companyProvisionsChanges = false
+
+  @Getter getProvisionsRemoved!: boolean
+
+  // Setters
+  @Action setProvisionsRemoved!: ActionBindingIF
+
+  private updateProvisionsRemoved (provisionsRemoved: boolean): void {
```
severinbeauvais, line 43: If you touch this file again, consider removing this method, since you can call `setProvisionsRemoved` directly in the template.
If not, it's fine.

src/components/Articles/Articles.vue
```diff
@@ -0,0 +1,78 @@
...
+  /** Emits Have Changes event. */
+  @Emit('haveChanges')
+  private emitHaveChanges (haveChanges: boolean): void {}
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+.section-container {
```
severinbeauvais, line 61: Please delete any used classes.

src/components/Articles/Articles.vue
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.define-article-header {
+  display: flex;
+  background-color: $BCgovBlue5O;
+  padding: 1.25rem;
+}
+
+.define-article-title {
+  padding-left: 0.5rem;
```
severinbeauvais, line 76: I noticed inconsistency between these titles across the various components.
- they should all be bold (can use class="font-weight-bold" instead of `<strong>`)
- they should all have 0.5 rem left padding (can use class="pl-1" instead)

Could you please check the other titles and format them all similarly?

My preference is to keep specifying an id or classname for them (for testing selector purposes) but to have the bold and padding classes right in the HTML.

src/components/Articles/Articles.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+  <v-card flat id="articles">
+    <confirm-dialog ref="confirm" attach="#app" />
```
severinbeauvais, line 3: It looks like you don't use this dialog in this component. Please remove accordingly.

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,227 @@
+<template>
+  <div id="company-provisions">
+    <confirm-dialog ref="confirmCompanyProvisionsDialog" attach="#company-provisions" />
```
severinbeauvais, line 3: Is this used? If not, please delete.

src/components/Articles/Articles.vue
andrepestana-aot: All of them are used.

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,227 @@
...
+    <v-layout v-if="!isEditing">
```
severinbeauvais, line 4: v-layout and v-flex are deprecated.
Use v-container, v-row and v-col instead.

https://www.google.com/search?q=vuetify+v-layout+deprecated
https://vuetifyjs.com/en/getting-started/upgrade-guide/#grid
https://vuetifyjs.com/en/components/grids/#v-row

severinbeauvais: (Flexbox helper classes are still OK I think.)

https://vuetifyjs.com/en/styles/flex/#flex

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,227 @@
...
+        <label><strong>Pre-existing Company Provisions</strong></label>
+        <v-chip v-if="hasProvisionsRemovedPropsChanged" x-small label color="#1669BB" text-color="white">
+          CHANGED
+        </v-chip>
+      </v-flex>
+      <v-flex xs7 class="info-text" v-if="provisionsRemoved">
+        The company has resolved that none of the Pre-existing Company Provisions are to apply to this company.
+      </v-flex>
+      <v-flex xs7 class="info-text" v-else>
+        This company has Pre-existing Company Provisions
```
severinbeauvais, line 15: Since alterate sentence ends in a period, this one should too.

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,227 @@
...
+            </p>
+            <v-layout>
+                <v-checkbox
+                  id="cp-checkbox"
+                  class="mt-1"
+                  :color="isInvalid ? checkBoxColor : ''"
+                  v-model="draftProvisionsRemoved"
+                />
+                <span class="checkbox-label" :class="{ 'invalid': isInvalid }">
+                  The company has resolved that none of the Pre-existing Company Provisions are to apply to this company
```
severinbeauvais, line 75: Add period at end?

(Does the ux spec explicitly have no end period???)

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,227 @@
...
+import { CommonMixin } from '@/mixins'
+import { ConfirmDialog } from '@/components/dialogs'
+
+@Component({
+  components: {
+    ConfirmDialog
+  }
+})
+export default class CompanyProvisions extends Mixins(CommonMixin) {
+  private draftProvisionsRemoved: boolean = false
```
severinbeauvais, line 119: Redundant type declaration (can be inferred from initial value).

(Some linters complain about this. Not ours -- yet -- but it's good practise anyway.)

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,227 @@
...
+            <p class="instructions-paragraph">
+              Complete this item only if the company has resolved that none of the Pre-existing Company Provisions
+              are to apply to this company (refer to Part 17 and Table 3 of the Regulation under the Business
+              Corporations Act).
+            </p>
+            <v-layout>
+                <v-checkbox
+                  id="cp-checkbox"
+                  class="mt-1"
+                  :color="isInvalid ? checkBoxColor : ''"
```
severinbeauvais, line 71: Code smell (esp using a prop with a hard-coded hex color).

Could you make this component invalid/disabled instead, to use the built-in features?

Or, could you conditionally set a class (`:class=...`), which would specify the styles you want?

andrepestana-aot: I tried to set the style with a class but it didn't work. I checked the documentation for that component it says I should set the color prop.

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,227 @@
...
+
+  // Emitters
+  @Emit('companyProvisionsChanged')
+  private emitCompanyProvisionsChanged (provisionsremoved: boolean): void {}
+
+  @Emit('haveChanges')
+  private emitHaveChanges (haveChanges: boolean): void {}
+
+  // Watchers
+  @Watch('provisionsRemoved', { deep: true, immediate: true })
```
severinbeauvais, line 138: You don't need "deep" as this is not an object (ie, don't need to watch object props).

severinbeauvais: Ditto on line 147.

andrepestana-aot: yes, you're right. Thanks!

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  .v-btn[disabled] {
+    color: white !important;
+    background-color: #1669bb !important;
+    opacity: 0.2;
+  }
+}
+.instructions-paragraph {
+  margin-bottom: 1.875rem;
```
severinbeauvais, line 215: Our convention has been for a lower element to specify the whitespace above it, instead of an upper element to specify the whitespace below it.

Also, try to use the built-in Vuetify classes for margins and padding... it makes it much easier to fix layout right in the template above instead of having to look up custom classes all the time.

https://vuetifyjs.com/en/styles/spacing/

src/components/Articles/Articles.vue
severinbeauvais: I don't see a section-container on this page ...?

src/components/Articles/CompanyProvisions.vue
severinbeauvais: If you can't make this component invalid or disabled to achieve the style you want then OK, use the :color directive. But use the color name instead of a property to a hard-coded hex value.

Because this is a Vuetify component, it doesn't know about our app's [CSS variables](https://github.com/bcgov/business-edit-ui/blob/800fbf3cb3049cfbb739ca5e83570e1762e57313/src/assets/styles/theme.scss#L63). But it does know about the [Vuetify theme colors](https://github.com/bcgov/business-edit-ui/blob/800fbf3cb3049cfbb739ca5e83570e1762e57313/src/main.ts#L73) so use that name instead.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: We need to initialize it to be reactive: https://vuejs.org/v2/guide/reactivity.html

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: - I tried to use those built-in vuetify classes but there's no one with 30px as specified by UX. That's why I added this class. 
- If we need to set the whitespace to the lower element, should I set the bottom margin/padding to 0px for the upper element?

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: We should add the period. Will come in the next commit. Thanks.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: Will change.

src/components/Articles/Articles.vue
andrepestana-aot: Will check. Thanks.

src/components/Articles/Articles.vue
andrepestana-aot: line 10

src/components/Articles/CompanyProvisions.vue
severinbeauvais: My point is that some linters (and maybe ours eventually) will complain about redundant type declaration here (you declare it as boolean and then assign a boolean value).

It's not broken atm, but it might eventually break if you don't use simply:
`private draftProvisionsRemoved = false`

Eg, https://stackoverflow.com/questions/46799828/tslint-type-trivially-inferred-why-is-it-bad-practice-to-include-the-type-he

src/components/Articles/CompanyProvisions.vue
severinbeauvais: Yes, I know that there is no equivalent for 30px... oh well.

Yes, add the whitespace to the lower element and remove any extra whitespace from the upper element. There are a few exceptions... for example, this self-contained date block has internal bottom padding to have a gutter all around, but no bottom margin.

image

src/components/Articles/Articles.vue
severinbeauvais: Oops. You're right, I'm wrong. Thanks for your patience :)

andrepestana-aot: 👍 Thank you!

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: Got it. I thought that typescript allowed you to reassign a string for example to that variable if you don't explicitly declare the type. I checked that it's only possible using ```any``` type. Thanks!

src/components/Articles/Articles.vue
andrepestana-aot: Done.

src/components/Articles/Articles.vue
andrepestana-aot: Ok. I'm going to use a css class but it's pl-1 in this case. According to Tracey we should keep aligned to YourCompany component and I'm doing this.

src/components/Articles/Articles.vue
andrepestana-aot: Removed. Thanks!

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: Removed.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: Done.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: Done.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: Done. Thanks.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: Done.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: Done.

src/components/Articles/Articles.vue
severinbeauvais: Yes, pl-1 sounds right,

Do all titles have the same style now?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.5.8",
+  "version": "0.5.9",
```
severinbeauvais, line 3: Sorry, now we already have a .8 and .9. Please now update to .10.

src/components/Articles/Articles.vue
andrepestana-aot: Sorry I meant: it's **not** pl-1 in this case.

severinbeauvais: I would like to see all titles with the same styles (according to the ux specs). Can you confirm that all titles have the same classes? `class="font-weight-bold pl-2"` or whatever?

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,255 @@
...
+import { CommonMixin } from '@/mixins'
+
+@Component({})
+export default class CompanyProvisions extends Mixins(CommonMixin) {
+  private draftProvisionsRemoved = false
+  private isEditing = false
+  private haveChanges = false
+  private originalProvisionsRemovedValue = false
+  private isInvalid = false
+  private checkboxColor = this.$vuetify.theme.currentTheme.error
```
severinbeauvais, line 124: This is cool. I didn't know you could reference Vuetify's themes this way 👍 

But as you're not using this property, please delete it.

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,255 @@
...
+
+  // Props
+  @Prop({ default: () => { return false } })
```
severinbeauvais, line 127: This default syntax is for objects. There is a simpler syntax for basic variables. Please update if you get a chance.

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,255 @@
...
+    margin-left: 0.5rem;
+  }
+
+  .v-btn {
+    min-width: 6.5rem;
+  }
+
+  .v-btn[disabled] {
+    color: white !important;
+    background-color: #1669bb !important;
```
severinbeauvais, line 210: Please use a named colour (CSS variable) for this. See our SCSS theme file.

severinbeauvais (commented):
As this is still a Draft PR, I assume it is still in progress.

I am looking for more unit tests:
- check various initial conditions
- check what props are working as expected
- check that UI changes as expected
- check any validation rules
- check that events are emitted as expected

src/components/Articles/Articles.vue
andrepestana-aot: Most of them have a custom classes and some use font-weight-bold instead. I'm going to use font-weight-bold too.

package.json
andrepestana-aot: 👍

severinbeauvais: Or check with @jachurchill to see if he will take 10 (and you would take 11)... or just update this last before your PR is merged.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: 👍

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: 👍

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: 👍

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,246 @@
...
+        <v-col cols="7" class="pt-0 pl-0 info-text" v-else>
+          This company has Pre-existing Company Provisions.
+        </v-col>
+        <v-col mt-n2 cols="2" class="pt-0 align-right" v-if="!hasProvisionsRemovedPropsChanged">
+          <v-btn id="correct-company-provisions" text color="primary" @click="isEditing = true">
+            <v-icon small>mdi-pencil</v-icon>
+            <span>{{ editLabel }}</span>
+          </v-btn>
+        </v-col>
+        <v-col pt-0 mt-n2 cols="2" class="align-right" v-else>
```
severinbeauvais, line 23: These spacing helper classes need to be in the class attribute.

severinbeauvais: Also on line 56 below.

severinbeauvais: And line 84.

Please check to see if you missed any others.

src/components/Articles/CompanyProvisions.vue
```diff
@@ -0,0 +1,246 @@
...
+                  The company has resolved that none of the Pre-existing Company Provisions are to apply to this
+                  company.
+                </span>
+            </div>
+          </v-row>
+        </v-col>
+      </v-row>
+      <v-row>
+        <v-container pt-10>
+          <v-row xs12>
```
severinbeauvais, line 85: I don't think row elements take this attribute. I think it's only for col elements.

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: 👍

src/components/Articles/CompanyProvisions.vue
andrepestana-aot: 👍

package.json
andrepestana-aot: 👍

src/components/Articles/CompanyProvisions.vue
```diff
@@ -8,19 +8,19 @@
             CHANGED
           </v-chip>
         </v-col>
-        <v-col cols="7" class="pt-0 pl-0 info-text" v-if="provisionsRemoved">
+        <v-col id="noneOfProvisionsApplyText" cols="7" class="pt-0 pl-0 info-text" v-if="provisionsRemoved">
```
severinbeauvais, line 11: This is not critical, but this is not the ID format (style) of other components in our projects. Please use kebab-case for element IDs and classes. (We use snake-case for names that are used in the script, like refs or other properties or methods.)

PR #125, opened 2021-03-11
5178 & 6172 Share structure Integration / Resolution & Court Order Dates

cameron-freshworks:
*Issue #:* /bcgov/entity#5178
*Issue #:* /bcgov/entity#6172

*Description of changes:*

* Integrates the shared Share structure component.
* Adds summary version of Share Structure.

* Implements Resolution or Court order dates components.

*Caution:*
* There is some failing unit tests from the Date-Mixin, however i pulled master and noticed the same situation there. Thoughts on how to process here?


package.json
```diff
@@ -13,13 +13,16 @@
   },
   "dependencies": {
```
severinbeauvais, line 14: Please bump up app version.

cameron-freshworks: Will do, was waiting to see what it was currently. 
Is changing daily 😆

src/components/Articles/ResolutionDates.vue
```diff
@@ -0,0 +1,164 @@
...
+        </v-col>
+      </v-row>
+
+      <!-- Display NEW Dates -->
+      <v-row no-gutters v-if="addedDates && addedDates.length >= 1" class="my-4">
+        <v-col cols="3"></v-col>
+        <v-col cols="7">
+          <ul class="resolution-date-list info-text pl-0 mt-2">
+            <li v-for="(date, index) in addedDates"
+                :key="`newResolutionDate: ${index}`"
```
severinbeauvais, line 37: This likely works but it may look better to have keys such as `newResolutionDate-1`. Thoughts?

severinbeauvais: Also line 81?

cameron-freshworks: Yes, i think you are right. Will update!

src/components/ShareStructure/ShareStructures.vue
```diff
@@ -0,0 +1,73 @@
...
+
+    <resolution-date-dialog
+      attach="#share-structures"
+      :dialog="toggleResolutionDateDialog"
+      @emitClose="toggleResolutionDateDialog = false"
+    />
+
+    <share-structure
+      :is-edit-mode="isEditMode"
+      :isCorrection="isCorrectionView()"
```
severinbeauvais, line 12: Maybe we should make this a getter eh? (Not sure if that will work as it's declared in a mixin, but if it does work then it will be consistent and maybe a tad faster due to getters being cached.)

src/components/ShareStructure/ShareStructures.vue
```diff
@@ -0,0 +1,73 @@
...
+      :share-classes="getShareClasses"
+      :business-snapshot="getOriginalSnapshot"
+      :resolution-required="resolutionsRequired"
+      :edit-label="editLabel"
+      :edited-label="editedLabel"
+      @emitShareClasses="setShareClasses($event)"
+      @emitShareStructureChanged="setShareStructureChanged($event)"
+      @emitEditingShareStructure="setEditingShareStructure($event)"
+      @emitResolutionPrompt="toggleResolutionDateDialog = $event"
+    />
```
severinbeauvais, line 23: I love how clean this is 👍

src/components/ShareStructure/ShareStructures.vue
```diff
@@ -0,0 +1,73 @@
...
+  @Action setShareClasses!: ActionBindingIF
+  @Action setShareStructureChanged!: ActionBindingIF
+  @Action setEditingShareStructure!: ActionBindingIF
+
+  // Local Properties
+  private toggleResolutionDateDialog = false
+
+  /** Return true if changes to share structure rights will require a resolution date. */
+  get resolutionsRequired (): boolean {
+    return this.getNewResolutionDates.length === 0 && this.isAlterationView()
```
severinbeauvais, line 70: Is there any chance `this.getNewResolutionDates` isn't an array? Is it initialized as `[]`?

src/components/common/ActionChip.vue
```diff
@@ -1,21 +1,22 @@
...
   <div>
     <v-chip v-if="actionableItem.action === ActionTypes.ADDED"
-            x-small label color="#1669BB" text-color="white">ADDED</v-chip>
+            x-small label color="primary" text-color="white">ADDED</v-chip>
     <v-chip v-if="actionableItem.action === ActionTypes.EDITED"
-            x-small label color="#1669BB" text-color="white">CORRECTED</v-chip>
+            x-small label color="primary" text-color="white">{{editedLabel}}</v-chip>
     <v-chip v-if="actionableItem.action === ActionTypes.REMOVED"
-            x-small label color="#E0E0E0" text-color="grey darken-1">REMOVED</v-chip>
+            x-small label color="#E0E0E0" text-color="#212529">REMOVED</v-chip>
```
severinbeauvais, line 8: Should the v-chip in ListPeopleAndRoles be updated like this one too?

andrepestana-aot (commented):
Looks good to me but please merge changes from master. There are some little changes I did for CompanyProvisions.vue

src/mixins/filing-template-mixin.ts
```diff
@@ -42,6 +53,8 @@ export default class FilingTemplateMixin extends Vue {
   @Getter getBusinessContact!: ContactPointIF
   @Getter getAgreementType!: string
   @Getter getOriginalSnapshot: BusinessSnapshotIF[]
+  @Getter getNewResolutionDates!: string[]
+  @Getter getSnapshotShareStructure!: ShareStructureIF
```
severinbeauvais, line 57: Object or array?

src/mixins/filing-template-mixin.ts
```diff
@@ -203,7 +218,7 @@ export default class FilingTemplateMixin extends Vue {
         name: FilingTypes.ALTERATION,
         certifiedBy: this.getCertifyState.certifiedBy,
         date: this.getCurrentDate, // "absolute day" (YYYY-MM-DD in Pacific time)
-        folioNumber: this.getFolioNumber
+        folioNumber: this.getFolioNumber || ''
```
severinbeauvais, line 221: So it can't be null? Please add a comment (so someone doesn't refactor this later and break things).

src/utils/feature-flags.ts
```diff
@@ -8,8 +8,8 @@ declare const window: any
  * Uses "business-edit" project (per LD client id in config).
  */
 const defaultFlagSet: LDFlagSet = {
-  'alteration-ui-enabled': false,
-  'correction-ui-enabled': false
+  'alteration-ui-enabled': true,
+  'correction-ui-enabled': true
```
severinbeauvais, line 12: Remove this from final commit.

src/components/ShareStructure/ShareStructures.vue
cameron-freshworks: That was definitely the first choice, is there a clean way to get the route within the vuex getters? 
Time for some research! 😁

src/store/state/state-model.ts
```diff
@@ -79,6 +79,7 @@ export const stateModel: StateModelIF = {
   shareStructureStep: {
     valid: false,
     changed: false,
+    resolutionDates: [],
```
severinbeauvais, line 82: Did you see that Andre created a new object in here for "new alteration data"? Would that be a better place for this (or for the new resolution dates)?

https://github.com/bcgov/business-edit-ui/blob/4bf1bd88d855370d60edcd623cf20adf85bb616f/src/store/state/state-model.ts#L21

src/components/common/DatePicker.vue
```diff
@@ -0,0 +1,98 @@
+<template>
```
severinbeauvais, line 1: Have you thought about making this a shared component?

src/components/common/DatePicker.vue
```diff
@@ -0,0 +1,98 @@
...
+  <v-form :attach="attach" ref="form" id="date-picker-form" class="date-time-selectors">
```
severinbeauvais, line 2: It may be possible to have multiple "date picker forms" on the same page, so I wouldn't use an ID here (may be duplicate).

OTOH, the id would be a great name for the class instead :)

severinbeauvais (approved):
Very nice.

Impressive how clean the code can be with a good component/data architecture in place. I love it :)

src/components/ShareStructure/ShareStructures.vue
severinbeauvais: I would only have getters in the Vuex code that do something with `state`.

I'm wondering if we need a Getter Mixin or something for non-state getters? Or leave those in Common Mixin?

cameron-freshworks: There is a way to apply global mixins to the store i believe, 
`new Vuex.Store<any>({ mixins: { ...MixinHere } })`
I'm sure there is something more too it, but it's possible!

src/components/ShareStructure/ShareStructures.vue
cameron-freshworks: 😁

src/components/ShareStructure/ShareStructures.vue
cameron-freshworks: I don't believe so. It is initialized right in the store as `[]`. We only every add/remove the store property.

src/components/common/ActionChip.vue
cameron-freshworks: You bet! I have a commit for the Shared Repo coming to update some things.
After that, i will pull in the latest version and remove this from the local repo and use the shared component throughout.

src/components/ShareStructure/ShareStructures.vue
severinbeauvais: Good find. Is this something you want to champion in this project (and possibly NR2 later)? I leave it with you...

src/components/common/DatePicker.vue
cameron-freshworks: That was the intention :) 

After share structure, i realized making things shared is about 2Xeasier AFTER meeting the initial design requirements. 

Adjustments will always happen, but things seem to evolve on that end and it becomes tedious hopping back and forth between repo's i'm sure you know :)

src/components/common/DatePicker.vue
cameron-freshworks: hahah good point!

src/components/ShareStructure/ShareStructures.vue
severinbeauvais: I didn't see the new resolution dates in the store (but I didn't look very hard).

src/mixins/filing-template-mixin.ts
cameron-freshworks: Object

cameron-freshworks: That contains some properties and an array.

src/mixins/filing-template-mixin.ts
cameron-freshworks: hmm, yea. I found this when filing. The back end returns 'NoneType on String' error or something if null. 
A better fix here, is to initialize it i think. Will fix this.

src/mixins/filing-template-mixin.ts
severinbeauvais: Hmm, looks different than business-snapshot-interface::shareClasses.

src/store/state/state-model.ts
cameron-freshworks: ahh i see. 
I'm not sure. I'll have to look at it a bit more. 
Whats the idea? an object that contains data exclusive to Alterations? 
I think we already have that with the alteration object body.

src/mixins/filing-template-mixin.ts
cameron-freshworks: You betcha, they are different. ShareClassesIF is a sub interface. The ShareStructure is the parent object which the back end would return to us, i'm pretty sure. 

We break down that to share classes as it's a little easier to manage at the component level i think, for iterations and whatnot.

I can have a close look though, maybe something is awry!

src/store/state/state-model.ts
severinbeauvais: Yes, the idea was a store object for "current alteration data" as there wasn't one already that lent itself to what André was adding. I know that other alteration data is elsewhere in the store. The hope here was to start a new thing that other devs would migrate to (instead of creating more tech debt).

src/components/Summary/AlterationSummary.vue
```diff
@@ -197,6 +219,12 @@ export default class AlterationSummary extends Mixins(CommonMixin, DateMixin, Fi
     return (this.pleaseValidate && !this.getEffectiveDateTime.valid)
   }
 
+  /** Local getter, using a mixin method to detect changes to Share Structure. */
+  private get hasShareStructureChanged (): boolean {
+    return !this.isSame(this.getShareClasses, this.getSnapshotShareStructure?.shareClasses, ['action'])
```
severinbeauvais, line 224: I don't think the conditional chaining should be needed here. Isn't `snapshotShareStructure` initialized in the store? So it can never be null/undefined?

tests/unit/AlterationSummary.spec.ts
```diff
@@ -142,6 +144,6 @@ describe('Alteration Summary component', () => {
     const divs = wrapper.findAll('.alteration-date-time .v-card')
     expect(divs.length).toBe(2)
     expect(divs.at(1).text()).toContain('The alteration for this business will be effective as of:')
-    expect(divs.at(1).text()).toContain('Friday, March 5, 2021 at 8:30 am Pacific time')
+    expect(divs.at(1).text()).toContain('Friday, March 5, 2021 at 8:30 AM Pacific time')
```
severinbeauvais, line 147: No, it should be "am" in lower case. Maybe I missed a datetime formatter somewhere. If you find it, can you fix pls?

severinbeauvais: Ref: https://www2.gov.bc.ca/gov/content/governments/services-for-government/service-experience-digital-delivery/web-content-development-guides/web-style-guide/writing-guide/numbers#years-and-time

package.json
cameron-freshworks: > Looks good to me but please merge changes from master. There are some little changes I did for CompanyProvisions.vue

Will do, thanks the heads up Andre! 👍

cameron-freshworks:
> Looks good to me but please merge changes from master. There are some little changes I did for CompanyProvisions.vue

Will do, thanks for the heads up Andre 👍

src/mixins/filing-template-mixin.ts
cameron-freshworks: Fixed 👍

src/store/state/state-model.ts
cameron-freshworks: So if we mean all current alteration data, this will require some rework of actions, getters, mutations etc of all the components so far. Will also mean a new set of them separate from the current ones that are used for both filings. 
Also, the saving of drafts, parsing of drafts etc.

Not a bad idea, but a task that deserves its own ticket.

cameron-freshworks:
> Very nice.
> 
> Impressive how clean the code can be with a good component/data architecture in place. I love it :)

oo thanks sir :)

src/components/Summary/AlterationSummary.vue
cameron-freshworks: You are correct!

src/components/common/DatePicker.vue
cameron-freshworks: Fixed!

tests/unit/AlterationSummary.spec.ts
cameron-freshworks: K, this is my bad. I shouldn't have changed this to pass the test, will root this out!

src/components/Summary/AlterationSummary.vue
cameron-freshworks: Fixed

package.json
cameron-freshworks: Bumped to `0.5.16` 🤞 😆

src/store/state/state-model.ts
severinbeauvais: No, change nothing now (unless you are adding data that belongs in this new object).

This is "data management" that we will also have to think about in Create UI with all the new filing types.

Yes, there's tech debt in moving other data into this new object (if it makes sense). Leave that alone for now.

Point is, we're trying to avoid further tech debt :)

src/views/Alteration.vue
```diff
@@ -15,7 +15,7 @@
 
       <current-directors class="mt-10" />
 
-      <share-structure class="mt-10" />
+      <share-structures class="mt-10" />
```
severinbeauvais, line 18: Why is it plural?

tests/unit/CompanyProvisions.spec.ts
```diff
@@ -46,7 +46,7 @@ describe('company provisions', () => {
       .toBe('This company has Pre-existing Company Provisions.')
     expect(wrapper.find(changeCompanyProvisionsButton).exists()).toBe(true)
     expect(wrapper.find(defineCompanyProvisionsTitle).text())
-      .toBe('Pre-existing Company Provisions')
+      .toBe('Pre-existing  Company Provisions')
```
severinbeauvais, line 49: Ah, this is because of the br in the label.
This is why I something use spans for individual lines -- so I can use toBe() instead of toContain() -- though in this case you seem to be getting away with it 🤞

tests/unit/date-mixin.spec.ts
```diff
@@ -21,15 +21,15 @@ describe('Date Mixin', () => {
   })
 
   // FUTURE: this works locally but not in GHA; fix later
-  xit('returns correct values for createUtcDate()', () => {
+  it('returns correct values for createUtcDate()', () => {
```
severinbeauvais, line 24: So this works in GHA now? Or only certain times of the day (ie, before or after 5pm  local)?

If it works... please move the comment to line 32 instead. Thx!

cameron-freshworks: Sorry, it doesn't. It seems a few of the unit tests in this file are failing. 
I pulled master and ran them as well, and had the same results. 
Curious what course i should take here?

src/views/Alteration.vue
cameron-freshworks: 🤷  Because there is multiple share structures? hehe i guess that is open to interpretation though! 
Also, didn't want to conflict with the shared component.

tests/unit/CompanyProvisions.spec.ts
cameron-freshworks: Yea, i threw in the break to meet the designs. I know it was outside of my components but it looked nice to be uniform.
Thats a good idea though, i really like spans as well. Narrow down on those elements!

src/store/state/state-model.ts
cameron-freshworks: Ok sounds good. I'll give your RFC a deep dive asap, sounds like this is all related.

tests/unit/date-mixin.spec.ts
severinbeauvais: Change it back to `xit` for now.

severinbeauvais: (I don't actually know what's going here. My guess is the build environment doesn't have the locale we ask for, so it falls back to UTC.)

src/views/Alteration.vue
severinbeauvais: Isn't there only 1 share structure?

We can worry about this later. There's enough other stuff to do !!!

tests/unit/date-mixin.spec.ts
cameron-freshworks: I think you are right sir!
Done and done.

PR #137, opened 2021-03-22
6763 Fix FED datetime format for API submission

argush3:
Issue #: /bcgov/entity#6763

Description of changes:

- fixed FED datetime format for API submission
- added unit test

### DIFF
--- a/src/mixins/date-mixin.ts
+++ b/src/mixins/date-mixin.ts
@@ -172,6 +172,22 @@ export default class DateMixin extends Mixins(CommonMixin) {
   }
 
   /**
+   * Converts a date object to an ISO datetime string format that the API accepts (YYYY-MM-DDThh:mm:ss.sss+00:00)
+   # @example "2021-01-01 23:59:59 PST" -> "2021-01-02T07:59:59.000+00:00"
+   */
+  dateToApiIsoDateTimeString (date: Date): string {
+    if (!date) return null // safety check
+
+    // convert date object to ISO format that API accepts
+    // Note: the Z offset is not included as the API(Python)
+    // expects '+00:00' to be used for the offset
+    // eg, 2021-03-25T10:02:00.000+00:00
+    const dateStr = date.toISOString().slice(0, 23) + '+00:00'
+
+    return dateStr
+  }
+
+  /**
    * The number of days that 'date' is from today.
    * @returns -1 for yesterday
    * @returns 0 for today
--- a/.gitignore
+++ b/.gitignore
@@ -120,6 +120,7 @@ k8s/*.param
 */*/.idea/*
 /**/.idea/*
 .idea/*
+*.iml
 
 # Log files
 npm-debug.log*
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.5.22",
+  "version": "0.5.23",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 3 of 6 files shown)


severinbeauvais:
Please update the title to something shorter but informative.

Please update the description block according to the template. See other PRs as example (eg, https://github.com/bcgov/business-edit-ui/pull/138).

You may not have permissions to do so, but this PR should be assigned to YOU and appropriate reviewers should be assigned. (I will do it this time.)

src/mixins/date-mixin.ts
```diff
@@ -187,4 +208,15 @@ export default class DateMixin extends Mixins(CommonMixin) {
     const dateLocalMs = new Date(date).setHours(0, 0, 0, 0)
     return Math.round((dateLocalMs - todayLocalMs) / this.MS_IN_A_DAY)
   }
+
+  /**
+   * Padding function to ensure month, date, hour, minutes and seconds front padded with zero if input val < 10
+   * @returns '01' for 1
+   * @returns 10 for 10
+   */
+  pad (number : Number) : String {
```
severinbeauvais, line 217: The formal parameter type should be `number` (which is a type) not `Number` (which is a class).
The return type should be fixed similarly.

Finally, you don't need this as there is already a JS equivalent: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

src/mixins/date-mixin.ts
```diff
@@ -171,6 +171,27 @@ export default class DateMixin extends Mixins(CommonMixin) {
...
+    // convert date object to ISO format that API accepts
+    // eg, 2021-03-25T10:02:00.000000+00:00
+    const dateStr = date.getUTCFullYear() +
+      '-' + this.pad(date.getUTCMonth() + 1) +
+      '-' + this.pad(date.getUTCDate()) +
+      'T' + this.pad(date.getUTCHours()) +
+      ':' + this.pad(date.getUTCMinutes()) +
+      ':' + this.pad(date.getUTCSeconds()) +
+      '.' + (date.getUTCMilliseconds() / 1000).toFixed(6).slice(2, 8) +
+      '+00:00'
```
severinbeauvais, line 190: In my opinion, this is not the correct solution.

Admittedly, both the output of `date.toISOString()` (eg, 2021-01-02T07:59:59.000Z) and the Python format (eg, 2021-01-02T07:59:59.000Z) are ISO 8601, but apparently aren't compatible.

Given the examples above, it should be trivial to convert from one to the other. I would simply knock off the "Z" and append the offset string. See apiToDateAndTimeString() above and reverse the process.

argush3: so if i re-work this function(`dateToApiIsoDateTimeString`) to use `date.toISOString()` as a starting point, I would still be missing part of the milliseconds that the API expects.  What do we want to do about this?  Should we replace the Z with 000+00:00 ?  Or should I just grab the milliseconds with what I was originally using?  i.e. `(date.getUTCMilliseconds() / 1000).toFixed(6).slice(2, 8)`

severinbeauvais: As it turns out, `toISOString()` gives you the milliseconds, eg
```
> new Date().toISOString()
> "2021-03-22T20:20:56.052Z"
```

How inflexible is the Python ISO date parser? Do you have specify it all exactly? If so then, yes, you'll have to right-pad the milliseconds (and add the tz offset).

This is what I would do (but feel free to do something nicer!):
```
const dateStr = date.toISOString().slice(1,23) + '000+00:00'
```

Ref: https://www.w3schools.com/jsref/jsref_slice_array.asp

severinbeauvais: ^^ and add a code comment describing what you're doing ;)

argush3: For some reason I thought I needed the additional milliseconds.  I just tested it and the api did accept a format of `2021-03-26T13:04:00.000Z`.  I might have been just thinking we might need the millisecond precision part which is highly unlikely if I think about it.

So I'll just do as per your suggestion:

```
const dateStr = date.toISOString().slice(1,23) + '+00:00'
```

severinbeauvais (commented):
The code looks fine.

One last thing (sorry, I should have mentioned it sooner)... Please update the app's version by 0.0.1 in package.json (and then do `npm i`), and check in the new changes. This is to easily tell what version is deployed.

severinbeauvais (approved):

PR #141, opened 2021-03-24
5475 - tiny fix validation for non staff

jordiwes:
*Issue #:* /bcgov/entity#5475

*Description of changes:*

- allow for undefined email (non staff)

### DIFF
--- a/src/components/Summary/DocumentsDelivery.vue
+++ b/src/components/Summary/DocumentsDelivery.vue
@@ -86,7 +86,7 @@ export default class DocumentsDelivery extends Mixins(CommonMixin) {
 
   private validateEmailFormat (value: string): boolean {
     // allow empty as the email is optional
-    if ((value === '') || (value === undefined)) {
+    if (!value) {
       return true
     } else {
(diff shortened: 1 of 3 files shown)


src/components/Summary/DocumentsDelivery.vue
```diff
@@ -86,7 +86,7 @@ export default class DocumentsDelivery extends Mixins(CommonMixin) {
 
   private validateEmailFormat (value: string): boolean {
     // allow empty as the email is optional
-    if (value === '') {
+    if ((value === '') || (value === undefined)) {
```
severinbeauvais, line 89: What if value is `null` ?

https://developer.mozilla.org/en-US/docs/Glossary/Falsy

severinbeauvais (commented):
I think you should generalize the falsy check of the value.

Also, without updating the app version, you won't be able to tell that your fix is deployed. I'm OK either way with this. What do you think?

jordiwes:
I made the above falsy, and slid in another validation fix. Not necessary to update the version

severinbeauvais (approved):

PR #144, opened 2021-03-30
Moved staff payment to review page, updated styles and have it captur…

jachurchill:
…ed by validation

*Issue #:* /bcgov/entity#6378

*Description of changes:* Moved staff payment to review page, updated styles and have it captured by validation

### DIFF
--- a/src/components/common/StaffPayment.vue
+++ b/src/components/common/StaffPayment.vue
@@ -1,5 +1,5 @@
 <template>
-  <section id="staff-payment-section">
+  <section id="staff-payment">
     <header>
       <h2>3. Staff Payment</h2>
     </header>
@@ -88,4 +88,16 @@ export default class StaffPayment extends Vue {
 </script>
 
 <style lang="scss" scoped>
+  ::v-deep .v-input .v-label {
+    font-weight: normal;
+  }
+  ::v-deep .v-input--radio-group__input {
+    .v-radio:not(:first-child) {
+      padding-top: 2rem;
+    }
+    .v-input--checkbox {
+      padding-top: 2rem;
+    }
+  }
+
 </style>
--- a/src/store/mutations/mutations-model.ts
+++ b/src/store/mutations/mutations-model.ts
@@ -207,6 +207,7 @@ export const mutateStaffPayment = (state: StateIF, staffPayment: StaffPaymentIF)
 
 export const mutateStaffPaymentValidity = (state: StateIF, validity: boolean) => {
   state.stateModel.staffPaymentStep.valid = validity
+  state.stateModel.newAlteration.validFlags.isValidStaffPayment = validity
 }
 
 export const mutateFilingData = (state: StateIF, filingData: FilingDataIF) => {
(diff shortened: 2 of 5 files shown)


src/components/common/StaffPayment.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <section id="staff-payment-section">
+  <section id="staff-payment">
```
severinbeauvais, line 2: The previous id followed our naming convention. Why did you change it?

jordiwes: This needs to match 
export enum ReviewSummaryFlags {
  'effective-date-time',
  'court-order',
  'certify-section',
  'staff-payment',
  'document-delivery-section'
}

is probably why :)

src/views/Alteration.vue
```diff
@@ -73,6 +68,12 @@
         />
       </template>
 
+      <staff-payment
+        class="mt-10"
+        :validate="getAppValidate"
```
jordiwes, line 73: Doesn't look like the validate prop is used anywhere

src/store/mutations/mutations-model.ts
```diff
@@ -207,6 +207,7 @@ export const mutateStaffPayment = (state: StateIF, staffPayment: StaffPaymentIF)
 
 export const mutateStaffPaymentValidity = (state: StateIF, validity: boolean) => {
   state.stateModel.staffPaymentStep.valid = validity
+  state.stateModel.newAlteration.validFlags.isValidStaffPayment = validity
```
severinbeauvais, line 210: Don't change this right now... This comment is for all devs: this is a bit of a code smell...

I think we need to figure out our architecture for data that applies to different filings (like here - Corrections + Alterations). This is one approach but it results in duplicate code and won't scale well. I think a better approach would be to generalize data like this that is used in different filings. What do you guys think?

(I'm OK with this for now and we can deal with this if/when we add another filing to this project, but it would be nice to have an idea/plan what to do when we get there.)

src/components/common/StaffPayment.vue
jachurchill: yea as @jordiwes said below this was to match with the ReviewSummaryFlags, would it be better for me to change the flag instead?

src/views/Alteration.vue
severinbeauvais: IMHO - leave it (less work). Future ticket (do you have the number?) can update it if needed.

src/components/common/StaffPayment.vue
severinbeauvais: Consistent naming is always good, but I see nothing consistent in the above list of flags!

If you have to edit this code for any other reason, please rename. Otherwise just leave it (less work).

severinbeauvais:
Please update the app draft version number in package.json, then do `npm i`, then commit both package.json and the updated package-lock.json files.

src/views/Alteration.vue
jachurchill: from the ticket 

> (to note - the Top Level red bar validation shown on comp 10a2 and auto scroll can be complete in the Top Level validation ticket #https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/6546)

PR #154, opened 2021-04-01
POA draft fix / Minor design adjustments

cameron-freshworks:
*Issue #:* /bcgov/entity#6908

*Description of changes:*

* Updates the component versions for DatePicker, CourtOrderPoa & StaffPayment.

* added props to CourtOrderPoa to provide draft values.

* Minor style adjustments to Certify and Documents.

* Adjusted invalid class on Certify to display border on component box.

* Removed some extra margins for Certify and Documents.

* Reverted NameRequest filing body change, to always include but only specific values 😁  I noticed some issues on draft filings without including the baseline object here.

* Moved staffpayment into `staff` only block.


src/views/Alteration.vue
```diff
@@ -62,17 +62,19 @@
         <court-order-poa
           id="court-order"
           :validate="getAppValidate"
+          :draftCourtOrderNumber="getCourtOrderNum"
+          :hasDraftPlanOfArrangement="getPlanOfArrangement"
```
severinbeauvais, line 66: I recommend that the value be named "has...", otherwise it looks like it returns an object or something.

src/interfaces/alteration-interfaces/court-order-interface.ts
```diff
@@ -1,5 +1,6 @@
 export interface CourtOrderIF {
   fileNumber: string,
   orderDate?: string,
-  effectOfOrder?: string
+  effectOfOrder?: string,
```
severinbeauvais, line 4: What is the getter for this? It almost looks like it is `getCourtOrderNum`...

src/views/Alteration.vue
cameron-freshworks: ah right! Will fix and update on this end as well

src/interfaces/alteration-interfaces/court-order-interface.ts
cameron-freshworks: This could just be a static value string value.. though i'm not really sure the purpose.

cameron-freshworks: To also comment on the DATE, i inquired with Scott about this and he said there was no Date requirement in this component. 🤷

severinbeauvais: I see a lack of symmetry between the IF fields and the getters/setters.

What does getCourtOrderNum return then?

severinbeauvais: Ohhhhhhhhhhhhhhhh... `getCourtOrderNum` would make more sense if it was renamed `getFileNumber` (or getCourtOrderFileNumber) (or just have 1 getter for the whole object and then extract fields as needed).

src/mixins/filing-template-mixin.ts
```diff
@@ -85,6 +87,8 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
   @Action setProvisionsRemoved!: ActionBindingIF
   @Action setOriginalResolutionDates!: ActionBindingIF
   @Action setResolutionDates!: ActionBindingIF
+  @Action setCourtOrderNumber: ActionBindingIF
```
severinbeauvais, line 90: Maybe the getter and setter names could be better synchronized, too.

src/interfaces/alteration-interfaces/court-order-interface.ts
cameron-freshworks: There is a definite disconnect between what the UI designs call it and how the back end references it. 
I'm not sure which way to go here. 
ie the UI calls this the Court Order Number where the back end calls it the `fileNumber`. 
So your preference is renaming everything to `fileNumber`. 
This would involve another PR to shared components as well..

src/mixins/filing-template-mixin.ts
cameron-freshworks: 🤔 
I feel like Court order number is a consistent reference here within the UI. If we rename everything to be `fileNumber` it will be different from the UI components. 
one way or another there is a disconnect.

src/interfaces/alteration-interfaces/court-order-interface.ts
severinbeauvais: Argh.

How about adding a comment to fileNumber indicating that it's aka courtOrderNum (and maybe a similar comment for the getter/setter)?

src/mixins/filing-template-mixin.ts
severinbeauvais: OK, let's go with the lesser of 2 weevils 🐞 

(status quo)

src/interfaces/alteration-interfaces/court-order-interface.ts
cameron-freshworks: Will do!

What are your thoughts on this though, because this isn't the first time. Do we prefer there to be a bit of a translation between property names between ui / api, or would we have the translation at the point of display?

severinbeauvais: I prefer API and UI to be in sync... imagine there's a schema in place... and the template can label it whatever it wants (which is more likely to change than the data structures anyway).

It's good to not incur a bunch of tech debt, but I also don't want to create a ton of work for you :)

cameron-freshworks: ah i see what you mean! Well, i can change everything here to fileNumber. Will require another update to POA Shared component in the future to update the prop name.

src/mixins/filing-template-mixin.ts
cameron-freshworks: You've convinced me, will update all to fileNumber 😁 
In next commit!

src/App.vue
```diff
@@ -449,7 +449,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
       case SummaryActions.RESUME:
         // Save filing and return to dashboard
         await this.onClickSave()
-        this.goToDashboard(true)
+        // this.goToDashboard(true)
```
severinbeauvais, line 452: Remove this before final commit.

severinbeauvais (approved):
Ahhhh, good cleanup!

src/App.vue
cameron-freshworks: Absolutely, will do :) 
Useful for testing the filing structure and resuming drafts 😆

src/mixins/filing-template-mixin.ts
cameron-freshworks: Updated 👍

src/App.vue
cameron-freshworks: fixed!

PR #155, opened 2021-04-06
Pre-existing Comp Provisions Review

cameron-freshworks:
*Issue #:* /bcgov/entity#5177

*Description of changes:*

* Added review text when there is changes to the pre-existing company provisions.

* Fixed draft bug, where state wasn't restoring.

* Added Scroll to top effect when transitioning to Summary Mode.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -456,9 +456,12 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
         break
       case SummaryActions.CONFIRM:
         // If Summary Mode: Check validity, save and file else move into summary mode.
-        this.isSummaryMode
-          ? await this.validateApp()
-          : this.scrollToTop(document.getElementById('app')); this.setSummaryMode(true)
+        if (this.isSummaryMode) await this.validateApp()
+        else {
+          // We don't change views just interchange components, so scroll to top for better UX.
+          this.scrollToTop(document.getElementById('app'))
+          this.setSummaryMode(true)
+        }
         break
     }
   }
--- a/src/App.vue
+++ b/src/App.vue
@@ -458,7 +458,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
         // If Summary Mode: Check validity, save and file else move into summary mode.
         this.isSummaryMode
           ? await this.validateApp()
-          : this.setSummaryMode(true)
+          : this.scrollToTop(document.getElementById('app')); this.setSummaryMode(true)
         break
     }
   }
--- a/src/components/Articles/CompanyProvisions.vue
+++ b/src/components/Articles/CompanyProvisions.vue
@@ -145,7 +145,6 @@ export default class CompanyProvisions extends Mixins(CommonMixin) {
   @Watch('provisionsRemoved')
   private onProvisionsRemovedPropValueChanged (): void {
     if (!this.haveChanges) {
-      this.originalProvisionsRemovedValue = this.provisionsRemoved
       this.draftProvisionsRemoved = this.provisionsRemoved
       this.emitHaveChanges(this.haveChanges)
     }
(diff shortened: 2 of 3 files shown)


src/App.vue
```diff
@@ -458,7 +458,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
         // If Summary Mode: Check validity, save and file else move into summary mode.
         this.isSummaryMode
           ? await this.validateApp()
-          : this.setSummaryMode(true)
+          : this.scrollToTop(document.getElementById('app')); this.setSummaryMode(true)
```
severinbeauvais, line 461: Do both statements on this line get executed? It's not clear from the syntax.

I think this would be better within an actual if-then-else with curly braces.

src/components/Articles/CompanyProvisions.vue
```diff
@@ -145,7 +145,6 @@ export default class CompanyProvisions extends Mixins(CommonMixin) {
   @Watch('provisionsRemoved')
   private onProvisionsRemovedPropValueChanged (): void {
     if (!this.haveChanges) {
-      this.originalProvisionsRemovedValue = this.provisionsRemoved
```
andrepestana-aot, line 148: Will the originalProvisionsRemovedValue always start with false?

cameron-freshworks: Well it's a property local to the component, so it was always going to be whatever the prop value is. 

So when you resume a draft with a `provisionsRemoved` value of `true` it was setting the original value too ALSO `true`. It wasn't really a baseline "original" value. 

That being said, i'm not sure if a company could start an alteration with a provisionsRemoved value of `true`. 
I'll ask business on this one.

src/App.vue
cameron-freshworks: They do, because of the semi-colon. Javascript madness!

Will refactor for readability tho!

severinbeauvais (approved):

src/App.vue
cameron-freshworks: Updated in next commit!

src/components/Articles/CompanyProvisions.vue
cameron-freshworks: So yea, this was a really good question! It may involve a little more design but i believe this is handled in the back end. 

TLDR: `hasRestrictions` is a boolean that is passed in the `businessInfo`, which dictates whether we capture this boolean value ( resolved pre existing provisions ). 

If `hasRestrictions` is true, it would show the component and indicate the pre existing provisions have not been resolved. So it is safe to assume at the component level that the original value would be false if we are rendering the component. 

After filing, i would assume the back end would resolve the `hasRestrictions`.

andrepestana-aot: My concern is about editing draft alteration where that could be selected before.

I think it makes sense to 'original' in this.originalProvisionsRemovedValue because it is original value since the component was mounted as it is updated only if there was no changes.

cameron-freshworks: Oh i see. Well in a DRAFT alteration is was broken because it was setting the original value to `true` if the user selected the checkbox, then saved / resumed a draft.

cameron-freshworks: Thats what this fix is for, to not overwrite the default value ( always `false` in this case ). So it could identify a diff between the state value of `provisionsRemoved`.

andrepestana-aot: > Oh i see. Well in a DRAFT alteration is was broken because it was setting the original value to `true` if the user selected the checkbox, then saved / resumed a draft.

In this case, I think it is correct because what we call 'original' here is there original value since the user started to work on the alteration. Not original since it was previously filed.

andrepestana-aot: Ok, after we chat I got what you're saying. You're right! Thanks!

andrepestana-aot (approved):
👍

PR #156, opened 2021-04-07
UI - CourtOrder properties update

cameron-freshworks:
*Issue #:* /bcgov/entity#7038

*Description of changes:*

* This is only the Front End commits - The BE pr is here: https://github.com/bcgov/lear/pull/1136
* Will have an API commit/PR for that validator update.

* Applied CourtOrder object only when required.

* Made it optional in both interfaces and optional chain it when resuming drafts.

* Commented out unused components until we require them. I don't see them in designs, could potentially remove them altogether soon.


src/mixins/filing-template-mixin.ts
```diff
@@ -256,16 +256,18 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
       }
     }
 
-    // Save this boolean to drafts only
-    // It is a state required for UX when resuming drafts
-    if (isDraft) filing.alteration.courtOrder.hasPlanOfArrangement = this.getHasPlanOfArrangement
+    // Apply Court Order ONLY when it is required and applied
+    if (this.getHasPlanOfArrangement || this.getFileNumber) {
+      filing.alteration.courtOrder = {
+        fileNumber: this.getFileNumber,
```
severinbeauvais, line 266: The "if" condition above means File Number could be falsy. Is that OK here?

cameron-freshworks: Indeed, it is defined initially as falsy(`""`). Which is ok. 
Even if it is redefined here as `""`, we optionally set it when returning from DRAFT to avoid. 
Also to comment on that, i don't believe fallback handling is required there because we `set` it if it exists or don't if it doesn't. No harm

severinbeauvais (approved):

src/mixins/filing-template-mixin.ts
```diff
@@ -256,16 +256,18 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
+        effectOfOrder: EffectOfOrders.PLAN_OF_ARRANGEMENT,
```
argush3, line 267: so I don't believe that `effectOfOrder` should always have a value of `planOfArrangement`.  For example, if the user only provides the court order number and leaves the plan of arrangement checkbox unchecked, I believe we should be passing a null value for `effectOfOrder`

argush3 (commented):
Just the one comment about effectOfOrder.

src/mixins/filing-template-mixin.ts
cameron-freshworks: i don't know anything about this properties value, what other options there could be potentially etc. 
In the current state, this is the only value this property would ever have though 🤔 

I'm ok with modifying this later if we need to make it dynamic but until the time when we have some variance, i feel like it's ok to leave it, because i'm not even sure IF we will have that variance in this UI.

cameron-freshworks: Ah i do see what you mean though. Using the checkbox as this particular properties definer.

cameron-freshworks: i could refactor this to conditional apply the enum value, i'll see how the api handles a null value 🤞

argush3: Maybe I'm missing something here.  If you pass in `planOfArrangement` as the value for `filing.alteration.courtOrder.effectOfOrder` when only the court order number has been provided via the UI, then how would you be able to distinguish when the plan of arrangement checkbox should be unchecked?  The api will come back with the an effectOfOrder value of `planOfArrangement` when you retrieve that alteration the next time.

argush3: The api seems to accept null and an empty string for `effectOfOrder` when I tested it via postman.

cameron-freshworks: I think you mean via a draft?
I'll ping you!

cameron-freshworks: Updated as per comments 👍

argush3: took a look at the addition of the condition apply enum value.  Looks good!  👍

argush3 (approved):

PR #157, opened 2021-04-12
Share validation updates and requirement updates

cameron-freshworks:
*Issue #:* /bcgov/entity#6094

*Description of changes:*

* This is an initial commit, to contribute towards adding some validation mechanism for the components in editing mode.

* Updates to meet updated share structure requirements.

* Updated unit testing.


src/mixins/filing-template-mixin.ts
```diff
@@ -535,7 +535,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
   parseBusinessSnapshot (businessSnapshot: BusinessSnapshotIF = this.getOriginalSnapshot): void {
     // Store original snapshot
     this.setOriginalSnapshot(businessSnapshot)
-
+    console.log(businessSnapshot)
```
jordiwes, line 538: todo: remove console.log ;)

src/mixins/filing-template-mixin.ts
```diff
@@ -590,6 +590,8 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
         }
       }) || []
     )
+    // TODO: REMOVE MOCK DATA
+    businessSnapshot.shareStructure.resolutionDates = ['2012/04/06', '2014/06/16', '2015/07/05']
```
jordiwes, line 594: to remove?

src/mixins/filing-template-mixin.ts
cameron-freshworks: Thank you :)

src/enums/componentFlags.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Enum for component validator flags.
+ * enum properties MUST match component ID to scroll correctly.
+ * Order is arbitrary but follow component hierarchy.
+ * */
+export enum ComponentFlags {
+  'articles',
+}
```
jordiwes, line 8: can you elaborate what this is for?

src/mixins/filing-template-mixin.ts
cameron-freshworks: Definitely! Thank you, will do now and just re apply locally 👍

src/enums/componentFlags.ts
cameron-freshworks: Absolutely!

tests/unit/ResolutionDates.spec.ts
```diff
@@ -165,8 +211,9 @@ describe('Resolution Dates component - review mode', () => {
   let wrapperFactory: any
 
   beforeAll(() => {
+    store.state.stateModel.shareStructureStep.shareClasses = shareClasses
     wrapperFactory = (propsData: any) => {
-      return mount(ResolutionDates, { propsData: { ...propsData, isEditMode: false }, vuetify })
+      return mount(ResolutionDates, { propsData: { ...propsData, isEditMode: false }, vuetify, store })
```
jordiwes, line 216: Do we want to have a test for is resolutions dates valid

src/enums/componentFlags.ts
cameron-freshworks: o sorry, i thought you meant in a code comment 😆  There is one there!

To elaborate though, we have a mixin method `validateAndScroll` which takes this static enum. 
This enum is the ID values of components in the UI, so it can identify which component to scroll too when there is an invalid state. 

For example, we create an array of those valid flags `[false]`, and it will reference this enum so it knows which ID to scroll too. 

In this case we've only just started and have one id (flag), but will soon contain all components which have a validity check involved.

tests/unit/ResolutionDates.spec.ts
cameron-freshworks: We do, absolutely. I will expand more on the unit tests in the next commit as i'm sure there will be more to add. 
The intention here was to get the validity code into dev, because it has crossover with the larger validators ticket.

src/mixins/filing-template-mixin.ts
cameron-freshworks: removed!

src/mixins/filing-template-mixin.ts
cameron-freshworks: removed

tests/unit/ResolutionDates.spec.ts
jordiwes: 10-4

jordiwes (approved):
LGTM

src/components/Articles/ResolutionDates.vue
```diff
@@ -127,6 +153,10 @@ export default class ResolutionDates extends Mixins(DateMixin, CommonMixin) {
     return (this.previousDates?.length > 0)
   }
 
+  get currentDateString (): string {
+    return this.dateToDateString(new Date())
```
severinbeauvais, line 157: Are you sure the user's browser's date is set to Pacific timezone and is not correct (not incorrect or spoofed)?

Please use `getCurrentDate` instead. Please see how this is set to understand how to work around the possible date issues ^^^

cameron-freshworks: ah good point. Will do, thanks!

src/App.vue
```diff
@@ -603,13 +604,26 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
 
+  /** Perform high level component validations before proceeding to summary. */
+  private async validateComponents (): Promise<void> {
+    // Verify component validations
+    await this.setValidResolutionDate(this.getIsResolutionDatesValid)
+
+    // evaluate valid flags. Scroll to invalid components or continue to review.
+    if (this.validateAndScroll(this.getValidComponentFlags, ComponentFlags)) {
+      // We don't change views just interchange components, so scroll to top for better UX.
+      this.scrollToTop(document.getElementById('app'))
```
severinbeauvais, line 615: I just had a look at `scrollToTop`. It's actually asynchronous, which makes sense because we the scroll transition takes some ticks to complete. However, it's not properly written to await.

Could you please fix `scrollToTop` and then await it everywhere it's used?

(ie, we want the scroll to complete before we set summary mode to true)

src/App.vue
```diff
@@ -603,13 +604,26 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
+  }
+
   /** Perform high level validations before filing. */
   private async validateApp (): Promise<void> {
     // Prompt app validations
     this.setAppValidate(true)
 
     // evaluate valid flags. Scroll to invalid components or file alteration.
-    if (this.validateAndScroll(this.getAlterationValidFlags)) await this.onClickSave(false)
+    if (this.validateAndScroll(this.getAlterationValidFlags, ReviewSummaryFlags)) await this.onClickSave(false)
```
severinbeauvais, line 626: Just a small nit: when the code is more than trivial, I prefer to see the usual braces style.
```
if (condition) {
  do something
}
```

src/App.vue
cameron-freshworks: For sure, will do!

src/App.vue
cameron-freshworks: No prob at all!
sounds good, will update

src/components/dialogs/ResolutionDateDialog.vue
```diff
@@ -62,6 +66,10 @@ export default class ResolutionDateDialog extends Vue {
     return this.date ? '' : this.errorMsg
   }
 
+  get currentDateString (): string {
+    return this.dateToDateString(new Date())
```
severinbeauvais, line 70: Again, use the current date that the app maintains, not `new Date()`.

src/mixins/common-mixin.ts
```diff
@@ -40,14 +40,15 @@ export default class CommonMixin extends Vue {
...
+   * @param elementFlags Static element Ids to identify the component to scroll too.
    */
-  validateAndScroll (validatorFlags: ValidFlagsIF): boolean {
+  validateAndScroll (validatorFlags: object, elementFlags: object): boolean {
     // Create an array of the flag values
     const validFlagArray = Object.keys(validatorFlags).map(key => validatorFlags[key])
 
     // Find the first corresponding flag id that is invalid
-    const component = document.getElementById(ReviewSummaryFlags[validFlagArray.indexOf(false)])
+    const component = document.getElementById(elementFlags[validFlagArray.indexOf(false)])
```
severinbeauvais, line 51: Is line 55 async? Since it does a scroll (not a jump), I'm thinking it takes some time and must be async. And we want it to complete before continuing execution?

src/store/getters/state-getters.ts
```diff
@@ -387,6 +393,14 @@ export const hasContactInfoChange = (state: StateIF): boolean => {
     (businessContact.extension !== getSnapshotContact(state).extension)
 }
 
+/** Check for changes between current contact and original contact. */
+export const hasShareStructureChanges = (state: StateIF): boolean => {
+  const originalShareClasses = cloneDeep(state.stateModel.originalSnapshot?.shareStructure.shareClasses)
+  const currentShareClasses = cloneDeep(state.stateModel.shareStructureStep.shareClasses)
```
severinbeauvais, line 399: Why do you need to clone it to compare it?

src/store/getters/state-getters.ts
```diff
@@ -411,3 +425,34 @@ export const getFileNumber = (state: StateIF): string => {
...
   return state.stateModel.newAlteration.courtOrder.hasPlanOfArrangement
 }
+
+/** Get boolean indicating if the share structure contains any special rights of restrictions. */
+export const getHasRightsOrRestrictions = (state: StateIF): any => {
+  const shareClasses = state.stateModel.shareStructureStep.shareClasses
+  const hasRightsOrRestrictionsArr = []
+
+  // Isolate the hasRightsOrRestrictions boolean values
+  shareClasses.forEach(shareClass => {
```
severinbeauvais, line 435: Could you use `some` instead? Theoretically faster (and slicker ;) ).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

vue.config.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
     proxy: {
       // this is needed to prevent a CORS error when running locally
       '/local-keycloak-config-url/*': {
-        target: 'https://business-edit-dev.apps.silver.devops.gov.bc.ca/businesses/edit/config/kc/',
+        target: 'https://business-create-dev.apps.silver.devops.gov.bc.ca/businesses/edit/config/kc/',
```
severinbeauvais, line 34: Yikes, good fix :)

src/store/getters/state-getters.ts
cameron-freshworks: I don't actually, being a bit over cautious here. 
The root issue of some issues in the past was references with in share structure overriding something i didn't want to override.. 
perhaps it made me over cautious.. i can remove :)

src/store/getters/state-getters.ts
cameron-freshworks: I'll check it out.. no promises though 😁 lol

vue.config.js
cameron-freshworks: haha yes, for now! It might change tomorrow though 😆

src/store/getters/state-getters.ts
severinbeauvais: I think these can be removed, since I believe `isEqual()` only compares the objects without changing them.

src/store/getters/state-getters.ts
severinbeauvais: In the old days, we'd `break` out of "for" loops to stop searching when we found a match. Am I (carbon) dating myself, lol?

vue.config.js
severinbeauvais: Actually you're right... 🙄... Patrick and I were talking about using external (vanity) URLs for Create and Edit UIs.

Nevertheless, thanks for fixing my copy-paste bug 😛

src/store/getters/state-getters.ts
cameron-freshworks: hahah you've been exposed! jk jk

src/store/getters/state-getters.ts
cameron-freshworks: Absolutely. I've been condition to fear object comparisons...

src/components/Articles/ResolutionDates.vue
cameron-freshworks: Updated in next commit!

src/App.vue
cameron-freshworks: Updated in next commit!

src/App.vue
cameron-freshworks: Updated!

src/components/dialogs/ResolutionDateDialog.vue
cameron-freshworks: Fixed 👍

src/mixins/common-mixin.ts
cameron-freshworks: Updated 👍

src/store/getters/state-getters.ts
cameron-freshworks: Removed!

src/store/getters/state-getters.ts
cameron-freshworks: Updated!

tests/unit/ResolutionDates.spec.ts
```diff
@@ -205,6 +206,72 @@ describe('Resolution Dates component - edit mode', () => {
...
+    rows = wrapper.findAll('.row')
+    expect(rows.length).toBe(2)
+
+    // verify date picker form
+    expect(wrapper.find('.date-picker-form')).toBeDefined()
+
+    // inject a date
+    vm.onDateEmitted('2021-03-17')
+    store.state.stateModel.shareStructureStep.resolutionDates = ['2021-03-17']
+    await Vue.nextTick()
```
severinbeauvais, line 254: Great test. I bet you had fun with this one! :)

cameron-freshworks: haha always do. Tests not only test the code but our resolve.

cameron-freshworks:
> PS even shorter :)
> 
> ```
> return shareClasses.some(sc => sc.hasRightsOrRestrictions)
> ```
> 
> But... do you also have to check the series for every share class? (You might need an outer loop after all, sorry.)

oh nice, ok i will minimize. 

I don't believe we do actually :) I put in a comment above explaining why.

severinbeauvais (approved):
Looks good!

tests/unit/ResolutionDates.spec.ts
cameron-freshworks: Updated!

PR #169, opened 2021-04-20
Previous Resolution Dates & Validation UX improvements

cameron-freshworks:
*Issue #:* /bcgov/entity#7065

*Description of changes:*

* Fixes previous resolutions date not being present. Required a new request as there was a separate endpoint for resolution dates.
* Updates parsing and handling of previous resolution dates.
* Share structure validation updates, fixed some minor bugs.
* Updated Resolution Date validation styling as per UX feedback.
* Minor syntax clean up.
* Updated testing.


src/enums/reviewSummaryFlags.ts
```diff
@@ -5,8 +5,8 @@
  * */
 export enum ReviewSummaryFlags {
   'effective-date-time',
-  'court-order',
+  'document-delivery-section',
```
severinbeauvais, line 8: Could you update the comment on line 4 to say that order should follow the component layout, please?

src/mixins/filing-template-mixin.ts
```diff
@@ -496,7 +496,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     const shareStructure = filing.alteration.shareStructure
     this.setShareClasses(shareStructure.shareClasses)
     this.setResolutionDates(shareStructure.resolutionDates)
-    this.setOriginalResolutionDates(this.getSnapshotShareStructure?.resolutionDates || null)
+    this.setOriginalResolutionDates(businessSnapshot.resolutions)
```
severinbeauvais, line 499: Should this fall back to "null" if it's undefined, or is it never undefined?

src/mixins/filing-template-mixin.ts
```diff
@@ -596,8 +596,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
 
     // Store share classes and original resolution dates
     this.setShareClasses(cloneDeep(businessSnapshot.shareStructure.shareClasses))
-    this.setResolutionDates([]) // clear new resolution dates
```
severinbeauvais, line 599: Don't we still need to clear any new res dates the user entered?

src/mixins/legal-api-mixin.ts
```diff
@@ -270,6 +270,27 @@ export default class LegalApiMixin extends Mixins(FilingTemplateMixin) {
       })
   }
 
+  /**
+   * Fetch the resolutions of the current business.
+   * @returns a promise to return the data
+   */
+  async fetchResolutions (): Promise<any> {
```
severinbeauvais, line 277: Can this return an IF type?

jordiwes (approved):
Looks good!

src/enums/reviewSummaryFlags.ts
cameron-freshworks: Absolutely

src/store/getters/state-getters.ts
```diff
@@ -440,8 +456,21 @@ export const hasContactInfoChanged = (state: StateIF): boolean => {
 
 /** Whether share structure data has changed. */
 export const hasShareStructureChanged = (state: StateIF): boolean => {
-  const currentShareClasses = { ...omit(state.stateModel.shareStructureStep.shareClasses, ['action']) }
-  const originalShareClasses = { ...omit(state.stateModel.originalSnapshot?.shareStructure.shareClasses, ['action']) }
+  let currentShareClasses = getShareClasses(state)
+  let originalShareClasses = state.stateModel.originalSnapshot?.shareStructure.shareClasses
+
+  // Null action properties can be assigned to the ShareClasses when cancelling edits
+  // This is fail safe to ensure null actions are not included in the comparison
```
severinbeauvais, line 463: Why not just use `omit()` as before?

severinbeauvais: (I'm asking, not telling.)

src/store/mutations/mutations-model.ts
```diff
@@ -259,8 +260,8 @@ export const mutateProvisionsRemoved = (state: StateIF, provisionsRemoved: boole
   state.stateModel.newAlteration.provisionsRemoved = provisionsRemoved
 }
 
-export const mutateOriginalResolutionDates = (state: StateIF, resolutionDates: string[]) => {
-  state.stateModel.originalAlteration.alteration.shareStructure.resolutionDates = resolutionDates
+export const mutateOriginalResolutionDates = (state: StateIF, resolutionDates: ResolutionsIF[]) => {
+  state.stateModel.shareStructureStep.previousResolutionDates = resolutionDates
```
severinbeauvais, line 264: Is there any chance that `resolutionDates` is null?
If so, should provide fallback to [], or use conditional chaining everywhere it's used to prevent de-referencing a null array. (eg, list[0]?.field)

src/mixins/filing-template-mixin.ts
cameron-freshworks: Well it could be an empty [] but i don't think that's a problem right? 
If the request fails, we never make it into the failing i'm pretty sure.

tests/unit/ResolutionDates.spec.ts
```diff
@@ -92,7 +101,7 @@ describe('Resolution Dates component - edit mode', () => {
   })
 
   it('displays previous dates', async () => {
-    const wrapper = wrapperFactory({ previousDates: arrayOfDates })
+    const wrapper = wrapperFactory({ previousDates: previousDates })
```
severinbeauvais, line 104: Can be abbreviated to `{ previousDates }` if you like.

src/mixins/filing-template-mixin.ts
cameron-freshworks: I don't think so, if we want to save them for drafts, we would want to preserve whatever they've added so far no?

src/mixins/legal-api-mixin.ts
cameron-freshworks: oo yea :) Will do, cood catch!

src/store/getters/state-getters.ts
severinbeauvais: PS Very helpful code comment - thanks!

cameron-freshworks: My previous code wasn't working :( It was actually omitting `action` on the array level and not within the individual objects. 
Also, we don't actually want to omit all the actions because a `removed` for instance, we'd still want to see until the point of filing for the purposes of display/restore etc..

It's been quite complicated this part of the code 😅

src/mixins/filing-template-mixin.ts
severinbeauvais: Empty array is fine. I'm wondering if we would ever do `resolutions[0].date` (which would fail if resolutions is null.

As long as the API returns [] when there are no resolutions (re: fetchResolutions()) then it should be fine.

(What I'm checking: initial state, whatever we might assign to it, and how we might use it.)

TLDR: confirm API and if OK then nothing needs to be done here.

src/store/mutations/mutations-model.ts
cameron-freshworks: In this case, this is defined in the store as an empty [].

tests/unit/ResolutionDates.spec.ts
cameron-freshworks: oo nice trick. Will do!

src/mixins/filing-template-mixin.ts
severinbeauvais: This is `parseBusinessSnapshot` right? I thought it was called to reset/delete the current data? In that case, we would want to clear whatever the user had already entered.

src/store/getters/state-getters.ts
severinbeauvais: Yes, it looks complicated, and I wish it didn't.

src/store/mutations/mutations-model.ts
severinbeauvais: Yes, but do we ever assign anything else to it, like null?

src/mixins/filing-template-mixin.ts
cameron-freshworks: ah yes, this is the snapshot.
This is setting the original resolutions to store here. We don't need to set the addedDates to [] as the store already does that.

severinbeauvais: This method is also called to "delete all", so it _should_ delete whatever res dates (and anything else) the user entered.

Here: https://github.com/bcgov/business-edit-ui/blob/b80e6097a40744943840d24bf5adbfdfdb002b5e/src/App.vue#L501

severinbeauvais:
Have you tested that resolution dates are saved in a draft and correctly restored?

src/mixins/filing-template-mixin.ts
cameron-freshworks: Can Confirm 👍 

Screen Shot 2021-04-20 at 7 04 53 AM

src/enums/reviewSummaryFlags.ts
cameron-freshworks: Fixed in next commit

src/mixins/filing-template-mixin.ts
cameron-freshworks: 🤔 I don't quite understand. We can't delete previous existing Resolution Dates?
Maybe i'm misunderstanding. 

Are you saying we should preserve setting the new resolutions to `[]` when we parse the snapshot?
Only concern is we might delete dates from a draft, but i can look into it!

src/mixins/legal-api-mixin.ts
cameron-freshworks: fixed in next commit!

src/store/mutations/mutations-model.ts
cameron-freshworks: Should only ever be an empty array. I don't believe we would assign it `null`. 
Obviously i can't guarantee  nothing could/would ever go wrong but i do know how the component handles these dates. 
It either adds or subtracts strings(dates) from this array but never wipes the [] afaik.

cameron-freshworks:
> Have you tested that resolution dates are saved in a draft and correctly restored?

Indeed! Works as expected :)

src/mixins/filing-template-mixin.ts
severinbeauvais: I am saying that when we "delete all", which calls parseBusinessSnapshot(), we need to delete any data the user has entered, which in this case is the new resolutions dates.

cameron-freshworks: Would it make more sense to delete the user entered data outside of `parseBusinessSnapshot()` since we also use that when starting a new Alteration and resuming a Draft Alteration?

severinbeauvais: Yes, I think it would make more sense, but there's other data in parseBusinessSnapshot() that is also "cleared" and would have to be moved.
🤷

cameron-freshworks: kk, i'll have a look into both scenarios asap :)

tests/unit/ResolutionDates.spec.ts
cameron-freshworks: Updated in next commit 👍

src/mixins/filing-template-mixin.ts
cameron-freshworks: Added this back in as it doesn't seem to conflict with the drafts 👍 

I was having issues

severinbeauvais (approved):
Thanks for the updates. Looks good!

cameron-freshworks:
> Thanks for the updates. Looks good!

np! Thanks for the great review as always 👍

PR #171, opened 2021-04-21
Updated a few styles and changed the number in the header

jachurchill:
*Issue #:* /bcgov/entity#6378

*Description of changes:* Updated a few styles and changed the number in the header

### DIFF
--- a/src/components/common/StaffPayment.vue
+++ b/src/components/common/StaffPayment.vue
@@ -1,7 +1,7 @@
 <template>
   <section id="staff-payment">
     <header>
-      <h2>3. Staff Payment</h2>
+      <h2>4. Staff Payment</h2>
     </header>
 
     <staff-payment-component
@@ -88,6 +88,7 @@ export default class StaffPayment extends Vue {
 </script>
 
 <style lang="scss" scoped>
+  @import '@/assets/styles/theme.scss';
   ::v-deep .v-input .v-label {
     font-weight: normal;
   }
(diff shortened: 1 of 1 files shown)


src/components/common/StaffPayment.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <section id="staff-payment">
     <header>
-      <h2>3. Staff Payment</h2>
+      <h2>4. Staff Payment</h2>
```
cameron-freshworks, line 4: oh weird that it is picking up a diff here, i believe this change was already in.

cameron-freshworks (approved):

src/components/common/StaffPayment.vue
severinbeauvais: Ditto.

severinbeauvais (commented):
1. Looks like this branch needs to be rebased.
2. Should update the app version.
3. Since John has left now, Cam or I will need to pick this up.
4. Since this change is small, Cam or I can probably include this as part of other work.

Thoughts?

cameron-freshworks:
> 1. Looks like this branch needs to be rebased.
> 2. Should update the app version.
> 3. Since John has left now, Cam or I will need to pick this up.
> 4. Since this change is small, Cam or I can probably include this as part of other work.
> 
> Thoughts?

Sorry, late on this reply. Will be picking this up as well, updating the shared components versions and whatnot :)

cameron-freshworks:
Closed this PR. Work was incorporated into pr: https://github.com/bcgov/business-edit-ui/pull/173

PR #174, opened 2021-04-22
6949 - Update text and ($100.00) Fee 

andrepestana-aot:
*Issue #:* /bcgov/entity#6949

*Description of changes:*

- Updated text in Alteration page and AlterationSummary
- Added tests


src/components/Summary/AlterationSummary.vue
```diff
@@ -140,10 +140,13 @@
           </v-col>
 
           <v-col cols="9" class="inner-col-2">
-            <p class="info-text">Select the date and time of alteration of your business. You may select a date and
-              time up to 10 days in the future (note: there is an <strong>additional fee of $100.00</strong> to enter
-              an alteration date and time in the future). Unless a business has special requirements, most businesses
-              select an immediate Alteration Date and Time.
+            <p
+              id="effective-date-time-instructions"
+              class="info-text">
```
cameron-freshworks, line 145: These could probably fit on 1 line. We usually (i think*) go to multi line only when it doesn't fit.

src/components/Summary/AlterationSummary.vue
```diff
@@ -140,10 +140,13 @@
...
+              Select the date and time of alteration of your business. You may select a date and time up to 10 days in
+              the future (note: there is an <strong>additional fee of ${{futureEffectiveFees.toFixed(2)}}</strong> to
```
cameron-freshworks, line 147: Could this number ever be returned beyond 2 decimals?

src/components/Summary/AlterationSummary.vue
```diff
@@ -232,6 +238,23 @@ export default class AlterationSummary extends Mixins(
   /** Prop to perform validation. */
   @Prop() readonly validate: boolean
 
+  /** The future effective fee for Alteration. */
+  futureEffectiveFees: number = 0
+
+  async mounted () {
+    // fetches the fee prices to display in the text
+    const feePrices = await this.fetchFeePrices()
+    this.futureEffectiveFees = feePrices.futureEffectiveFees
```
cameron-freshworks, line 247: if `feePrices` request fails `feePrices.futureEffectiveFees` could return a type error. You might want to optional chain here `feePrices?.futureEffectiveFees` or set a fallback value to prevent `futureEffectiveFees` being defined as `null` or `undefined` with could bubble up into the UI.

src/components/Summary/AlterationSummary.vue
```diff
@@ -232,6 +238,23 @@ export default class AlterationSummary extends Mixins(
...
+  }
+
+  /** Fetches the Fee prices to display in the text */
+  private async fetchFeePrices (): Promise<FeesIF> {
+    const result = await Promise.resolve(this.fetchFilingFees(FilingCodes.ALTERATION,
+      this.getEntityType, true))
```
cameron-freshworks, line 253: Minor structural thing here. 
I'm not sure about other reviewers but if they didn't all fit in one line, it's nice to break the parameters into their own lines respectively. 
ie `Promise.resolve(
this.fetchFilingFees(FilingCodes.ALTERATION,this.getEntityType, true)
)`

src/components/Summary/AlterationSummary.vue
```diff
@@ -232,6 +238,23 @@ export default class AlterationSummary extends Mixins(
...
+    if (!('filingFees' in result)) throw new Error('Failed to fetch fees prices')
```
cameron-freshworks, line 254: Is this the same as `!result.filingFees` or `!result['filingFees]`?

src/components/Summary/AlterationSummary.vue
```diff
@@ -267,6 +290,11 @@ export default class AlterationSummary extends Mixins(
     await this.scrollToTop(document.getElementById('app'))
   }
 
+  // sum of alteration fees
+  get alterationFees (): number {
+    return this.getCurrentFees.filingFees + this.getCurrentFees.futureEffectiveFees
```
cameron-freshworks, line 295: Not a bad idea to protect yourself here if `getCurrentFees` is null or undefined. 
ie optional chain `this.getCurrentFees?` 
On that note, if this also is feeding a value to the UI for display, a fallback value might be required.

src/mixins/pay-api-mixin.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Mixin that provides integration with the Auth API.
+ */
+@Component({})
+export default class PayApiMixin extends Vue {
+  /**
+   * Fetches filing fees.
+   * @returns a promise to return the fees for a filing
+   */
+  async fetchFilingFees (filingCode: FilingCodes, entityType: CorpTypeCd, isFutureEffective: boolean): Promise<any> {
```
cameron-freshworks, line 19: return could be `Promise<FeesIF>` here no? 

Also, just curious. Is this information not something we get through the FeeSummary calculator or is that totally isolated to that component?

cameron-freshworks (commented):
Looks pretty good! Just had a few questions/comments :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.6.0",
+  "version": "0.6.1",
```
severinbeauvais, line 3: After you change this, you need to do `npm i` so the package-lock file is also updated (or you can cheat and update it manually).

src/components/Summary/AlterationSummary.vue
severinbeauvais: Yes, or we try to follow the style of the surrounding code, to make the code look nicer and thus be easier to review :)

severinbeauvais: Note that paragraph elements are block-level, so this
```
<p>blah</p>
```
displays the same as
```
<p>
   blah
</p>
```

... Which is to say, follow whatever style looks best in this scenario.

src/components/Summary/AlterationSummary.vue
severinbeauvais: I think you don't need `Promise.resolve` here (or elsewhere you do this).

`Await` takes care of resolving the promise.

PS If you need to catch the promise being rejected, it's like this:
`const foo = await bar().catch(() => doSomething())`

src/components/Summary/AlterationSummary.vue
severinbeauvais: Just to add a bit: if `result` object doesn't have a `filingFees` property, then `result.filingFees` is undefined (which is falsy).

src/components/Summary/AlterationSummary.vue
severinbeauvais: Right.
You possibly throw an error in `fetchFeePrices()` so you need to catch it here or `mounted` will fail and possibly the whole component won't render.
And it's possibly that `feePrices` doesn't contain the property `futureEffectiveFees` so you need to check it and have a reasonable fallback.

src/interfaces/payment-interfaces/taxes-interface.ts
```diff
@@ -0,0 +1,5 @@
+/** Interface for Fees Taxes from Payment API. */
+export interface TaxesIF {
```
severinbeauvais, line 2: If you wanted, since taxes are intimately tied to fees, you could declare this interface in the same file as fees. Up to you.

src/interfaces/payment-interfaces/fees-interface.ts
```diff
@@ -0,0 +1,14 @@
+/** Interface for Fees from Payment API. */
+import { TaxesIF } from '@/interfaces'
+
+export interface FeesIF {
```
severinbeauvais, line 4: If some of these are optionally returned by the network call, you could specify them as optional in the interface like this:
`futureEffectiveFees?: number`
LMK if any of these is optional. If not then I will resolve this comment.

src/mixins/pay-api-mixin.ts
```diff
@@ -0,0 +1,36 @@
+// Libraries
+import { Component, Vue } from 'vue-property-decorator'
+import { axios } from '@/utils'
+
+// Shared Interfaces
+import { AlterationFilingIF, EffectiveDateTimeIF } from '@/interfaces'
```
severinbeauvais, line 6: Remove imports not used in this file.

src/components/Summary/AlterationSummary.vue
```diff
@@ -232,6 +238,23 @@ export default class AlterationSummary extends Mixins(
   /** Prop to perform validation. */
   @Prop() readonly validate: boolean
 
+  /** The future effective fee for Alteration. */
+  futureEffectiveFees: number = 0
+
+  async mounted () {
+    // fetches the fee prices to display in the text
+    const feePrices = await this.fetchFeePrices()
```
severinbeauvais, line 246: Why do you need this, since you already did it in Alteration.vue right? Can you just read the values that you saved to the store?

src/components/Summary/AlterationSummary.vue
andrepestana-aot: ok.

src/components/Summary/AlterationSummary.vue
andrepestana-aot: No, it is returned with 1 decimal actually. It's just to make it uniform. And in the future it may come different from payment-api.

src/mixins/pay-api-mixin.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (isFutureEffective) {
+      url += '?futureEffective=true'
+    }
+    return axios.get(url)
+      .then(response => {
+        const fees = response?.data
+        if (!fees.filingFees) {
+          throw new Error('Invalid API response')
+        }
+        return fees
```
severinbeauvais, line 30: I think it would be better to return `fees.filingFees` here, so that you don't need extra methods to extract it later (see my other comment on this).

src/components/Summary/AlterationSummary.vue
andrepestana-aot: Yes, good point indeed. But then I should catch it and emit a "fetchError" for the parent component deal with it (Alteration page in this case)?

src/mixins/pay-api-mixin.ts
severinbeauvais: The FeeSummary component makes the same call to get its information, but I think it would be awkward to scrape that info from it. Better if we don't assume anything about that component.

src/components/Summary/AlterationSummary.vue
andrepestana-aot: I'm not saving the feePrices in the store, only the currentFees for the filing. Do you think I should save it to the store as well?

src/components/Summary/AlterationSummary.vue
severinbeauvais: That's a good question. I had to think about this...

1. We don't handle fee errors in any other case (eg, in Fee Summary), so I think we don't need to think too much about it here.

2. If we fail to get the fees, I don't think we should block this alteration filing altogether. I think we can probably display some reasonable fallback text instead. Ask Scott/Tracey about this, or come up with something yourself and pass it by us. This should be a rare scenario so (1) don't crash, and (2) try to show something reasonable, even if it's only "unknown" for the fee numbers.

src/components/Summary/AlterationSummary.vue
severinbeauvais: The problem I see is too many calls that do (almost) the same thing.

You could possibly just make 1 call, in Alteration.vue, and then extract/sum the individual fees you care about. Unfortunately that's making an assumption about adding the individual fees... but I think I saw somewhere in here where you already do that, so it's not any worse :)

src/components/Summary/AlterationSummary.vue
severinbeauvais: This is OK with me. There are other places in the code that use `toFixed(2)` for amounts.

package.json
andrepestana-aot: ok

src/components/Summary/AlterationSummary.vue
andrepestana-aot: ok. I added the feePrices to the store. Thanks.

src/components/Summary/AlterationSummary.vue
andrepestana-aot: ok. I changed to what Tracey mentioned in the ticket - if we don't get the data, the text will be displayed without the values but without breaking the text.

src/components/Summary/AlterationSummary.vue
andrepestana-aot: Ok. Yeah, I changed that. Thanks.

src/components/Summary/AlterationSummary.vue
andrepestana-aot: 👍

src/components/Summary/AlterationSummary.vue
andrepestana-aot: Yeah, you're right. I changed it to check truthy for `this.getCurrentFees?.filingFees`. Otherwise it doesn't make sense to display anything for alterationFees.

src/interfaces/payment-interfaces/fees-interface.ts
andrepestana-aot: Ok. I did it based on the schema:
https://github.com/bcgov/sbc-pay/blob/development/pay-api/src/pay_api/schemas/schemas/fees.json

src/interfaces/payment-interfaces/taxes-interface.ts
andrepestana-aot: 👍

src/mixins/pay-api-mixin.ts
andrepestana-aot: 👍

src/mixins/pay-api-mixin.ts
andrepestana-aot: 👍

src/views/Alteration.vue
```diff
@@ -256,6 +274,22 @@ export default class Alteration extends Mixins(CommonMixin, LegalApiMixin, Filin
     Vue.nextTick(() => this.setHaveUnsavedChanges(false))
   }
 
+  /** Fetches current fees for the Filing */
+  private async fetchCurrentFees (): Promise<FeesIF> {
+    const result = await Promise.resolve(this.fetchFilingFees(FilingCodes.ALTERATION,
+      this.getEntityType, this.getEffectiveDateTime.isFutureEffective))
+    if (!('filingFees' in result)) throw new Error('Failed to fetch current fees')
```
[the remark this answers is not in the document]
andrepestana-aot: ok. Thanks!

severinbeauvais:
> Yes, it is as we check for it. Do you suggest any other approach here?

Yes, I believe your code is safe. But I like consistency, so I would either init state to an empty object, or fall back to zero values here. Search the code for "empty address" for an example of the latter. Up to you, but currently you init to values and then fall back to an empty object, which is asymmetrical :)

andrepestana-aot:
> > Yes, it is as we check for it. Do you suggest any other approach here?
> 
> Yes, I believe your code is safe. But I like consistency, so I would either init state to an empty object, or fall back to zero values here. Search the code for "empty address" for an example of the latter. Up to you, but currently you init to values and then fall back to an empty object, which is asymmetrical :)



> > Yes, it is as we check for it. Do you suggest any other approach here?
> 
> Yes, I believe your code is safe. But I like consistency, so I would either init state to an empty object, or fall back to zero values here. Search the code for "empty address" for an example of the latter. Up to you, but currently you init to values and then fall back to an empty object, which is asymmetrical :)

Yes, ok. I did it. I thought about doing that I thought you guys wouldn't like it. Thanks.

cameron-freshworks (approved):
Looks good, thanks for the updates 👍

src/store/state/state-model.ts
```diff
@@ -217,5 +217,33 @@ export const stateModel: StateModelIF = {
     shareStructure: false,
     incorporationAgreement: false
   },
-  summaryMode: false
+  summaryMode: false,
+  currentFees: {
```
severinbeauvais, line 221: There is no reason to initialize this to zeros any more. I recommend just initializing this (and feePrices below) to null, since you now check for that when you "get" values from it. (Plus it makes things consistent.)

src/views/Alteration.vue
```diff
@@ -245,6 +272,14 @@ export default class Alteration extends Mixins(CommonMixin, LegalApiMixin, Filin
         priority: false
       })
 
+      // update the current fees for the Filing
+      this.setCurrentFees(await this.fetchFilingFees(
+        FilingCodes.ALTERATION, this.getEntityType, this.getEffectiveDateTime.isFutureEffective
+      ) || {})
+
+      // fetches the fee prices to display in the text
+      this.setFeePrices(await this.fetchFilingFees(FilingCodes.ALTERATION, this.getEntityType, true) || {})
```
severinbeauvais, line 281: I'm not sure that this works as you expect...
I believe `fetchFilingFees` can throw an error, which you don't catch here, so it will bubble up to the main method and cause alteration loading to fail.
When an awaited method throws an error, it does not return falsy (which is what this code seems to expect).
You can catch the error like this (note the `catch`, which returns {}):
```
const filingFees = await this.fetchFilingFees(FilingCodes.ALTERATION, this.getEntityType, true).catch(() => {})
this.setFeePrices(filingFees)
```

severinbeauvais:
> Yes, it is as we check for it. Do you suggest any other approach here?

This is OK.

severinbeauvais:
> > > Yes, it is as we check for it. Do you suggest any other approach here?
> > 
> > 
> > Yes, I believe your code is safe. But I like consistency, so I would either init state to an empty object, or fall back to zero values here. Search the code for "empty address" for an example of the latter. Up to you, but currently you init to values and then fall back to an empty object, which is asymmetrical :)
> 
> Yes, ok. I did it. I thought about doing that but I thought you guys wouldn't like it. Thanks.

You did it in one commit but the final code doesn't include it.

I don't care either which you choose --- might as well just go with {} as it's the easiest -- as long as the code is consistent, which makes it easier to review/understand/maintain.

Thanks.

src/views/Alteration.vue
andrepestana-aot: ok. I see.. you're right. Thanks!

andrepestana-aot:
> > > > Yes, it is as we check for it. Do you suggest any other approach here?
> > > 
> > > 
> > > Yes, I believe your code is safe. But I like consistency, so I would either init state to an empty object, or fall back to zero values here. Search the code for "empty address" for an example of the latter. Up to you, but currently you init to values and then fall back to an empty object, which is asymmetrical :)
> > 
> > 
> > Yes, ok. I did it. I thought about doing that but I thought you guys wouldn't like it. Thanks.
> 
> You did it in one commit but the final code doesn't include it.
> 
> I don't care either which you choose --- might as well just go with {} as it's the easiest -- as long as the code is consistent, which makes it easier to review/understand/maintain.
> 
> Thanks.

Yes, as I defined it as FeesIF it can't be `{}`. It's consistent now.

src/store/state/state-model.ts
andrepestana-aot: 👍

src/views/Alteration.vue
severinbeauvais: PS You can test this by blocking the endpoint in Chrome dev tools. Or you can edit fetchFilingFees to try to call an endpoint that doesn't exist (eg /andre). And then see what the UI does...

src/components/Summary/AlterationSummary.vue
```diff
@@ -272,6 +277,21 @@ export default class AlterationSummary extends Mixins(
...
+  get alterationFees (): string {
+    if (this.getCurrentFees?.filingFees) {
+      return `($${(this.getCurrentFees.filingFees + this.getCurrentFees?.futureEffectiveFees).toFixed(2)} Fee)`
+    }
+    return ''
+  }
+
+  get futureEffectiveFeePrice (): string {
+    if (this.getFeePrices?.futureEffectiveFees) {
+      return `of $${this.getFeePrices.futureEffectiveFees.toFixed(2)}`
```
severinbeauvais, line 290: Does getFeePrices always return an object? Or sometimes null?

src/views/Alteration.vue
```diff
@@ -193,6 +206,35 @@ export default class Alteration extends Mixins(CommonMixin, LegalApiMixin, Filin
...
+  }
+
+  private get futureEffectiveFeesPrice (): string {
+    if (this.getFeePrices?.futureEffectiveFees) {
+      return `$${this.getFeePrices.futureEffectiveFees.toFixed(2)}`
+    }
+    return ''
+  }
+
+  private emptyFees: FeesIF = {
```
severinbeauvais, line 223: Move this to where the Fees IF is declared. Then anyone using the IF also has the empty IF it can use :)

Also it probably makes more sense for all of these to be '' or 0 (depending on their type).

src/components/Summary/AlterationSummary.vue
andrepestana-aot: Yes, it always returns an object but I prefer being safe here, as someone can change it in the future.

src/store/state/state-model.ts
severinbeauvais: Oh, what I actually meant was to use `currentFees: null` but I see you've gone the other way (which is better!).

---

So, here, please assign the empty IF, eg:
```
currentFees: cloneDeep(emptyFees),
feePrices: cloneDeep(emptyFees),
```

(You'll have to `import { cloneDeep } from 'lodash'`.)

(You have to clone the object otherwise boht variables point to the same instance of it.)

src/components/Summary/AlterationSummary.vue
severinbeauvais: OK, yes, I see the "if" now :)

I guess you don't need the "?" on line 283 for the same reason then.

src/components/Summary/AlterationSummary.vue
```diff
@@ -272,6 +277,21 @@ export default class AlterationSummary extends Mixins(
     await this.scrollToTop(document.getElementById('app'))
   }
 
+  // sum of alteration fees
+  get alterationFees (): string {
+    if (this.getCurrentFees?.filingFees) {
+      return `($${(this.getCurrentFees.filingFees + this.getCurrentFees?.futureEffectiveFees).toFixed(2)} Fee)`
```
severinbeauvais, line 283: What happens if filingFees is 0 ?
Or if futureEffectiveFees is 0 or undefined or null ? (Can this ever happen?)

src/components/Summary/AlterationSummary.vue
andrepestana-aot: ok.

src/views/Alteration.vue
andrepestana-aot: ok, moved. About the null values, I think it's better to keep it as null for the `emptyFees` because I know that it's not coming from pay-api (the prices may be zero in some future scenario).

severinbeauvais: OK

src/components/Summary/AlterationSummary.vue
andrepestana-aot: As now they can be null or number only, I'm checking against null so I know that the values where populated with values from the pay-api and not from `emptyFees`.

src/components/Summary/AlterationSummary.vue
severinbeauvais: What happens if futureEffectiveFees is 0 here?

src/views/Alteration.vue
andrepestana-aot: Yes, I did the test with Chrome Dev Tools and I could check you're really right. I didn't know we could do that in dev tools. Thanks for the tip!

src/store/state/state-model.ts
```diff
@@ -217,5 +217,7 @@ export const stateModel: StateModelIF = {
     shareStructure: false,
     incorporationAgreement: false
   },
-  summaryMode: false
+  summaryMode: false,
+  currentFees: emptyFees,
```
severinbeauvais, line 221: I gave you an example with `clone()` and told you why you should use it. Why didn't you use it?

src/views/Alteration.vue
```diff
@@ -245,6 +273,19 @@ export default class Alteration extends Mixins(CommonMixin, LegalApiMixin, Filin
         priority: false
       })
 
+      // update the current fees for the Filing
+      this.setCurrentFees(
+        await this.fetchFilingFees(
+          FilingCodes.ALTERATION, this.getEntityType, this.getEffectiveDateTime.isFutureEffective
+        ).catch(() => emptyFees) || emptyFees
```
severinbeauvais, line 280: What does `|| emptyFees` do here?
Same on line 286.

tests/unit/Alteration.spec.ts
```diff
@@ -284,6 +325,52 @@ describe('Alteration component', () => {
...
+    await flushPromises()
+    expect(store.state.stateModel.feePrices.filingFees).toBe(100)
+    expect(store.state.stateModel.feePrices.futureEffectiveFees).toBe(100)
+  })
+
+  it('display the fee prices properly', async () => {
+    await wrapper.setProps({ appReady: true })
+    await flushPromises()
+    store.state.stateModel.summaryMode = true
+    store.state.stateModel.tombstone.filingType = FilingTypes.ALTERATION
```
severinbeauvais, line 344: Don't use enums in the tests. Use the actual value. This makes the tests more fragile, so they will fail if a developer updates the enum without updating the tests.

src/interfaces/payment-interfaces/fees-interface.ts
```diff
@@ -0,0 +1,31 @@
...
+    filingTypeCode: string
+    futureEffectiveFees?: number
+    priorityFees?: number
+    processingFees?: number
+    serviceFees?: number
+    tax?: TaxesIF
+    total?: number
+}
+
+const emptyFees = {
```
severinbeauvais, line 16: You could type this :)

src/components/Summary/AlterationSummary.vue
andrepestana-aot: ops... update here as well. Thanks

severinbeauvais: fixed

src/store/state/state-model.ts
andrepestana-aot: Yes, done now. Sorry I was still revising the comments.

src/views/Alteration.vue
andrepestana-aot: 👍

tests/unit/Alteration.spec.ts
andrepestana-aot: ok.

src/interfaces/payment-interfaces/fees-interface.ts
andrepestana-aot: 👍

severinbeauvais (approved):

PR #175, opened 2021-04-22
Prevent ShareClass Removal

cameron-freshworks:
*Issue #:* /bcgov/entity#5693

*Description of changes:*

* Implements validation logic for minimum ShareClasses

* Adds unit testing.

*Note:*

* Applies to both Alterations and Corrections. No additional design/rules were commented on in the ticket for Corrections.

### DIFF
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -552,7 +552,7 @@ export const showFeeSummary = (state: StateIF): boolean => {
   return (haveFilingChange && !isEqual(getFilingData(state), defaultFilingData))
 }
 
-/** Return a boolean indicating if the minimum share classes requirements are not met. */
+/** A boolean indicating if the minimum share classes requirements are not met. */
 export const invalidMinimumShareClass = (state: StateIF): boolean => {
   const shareClasses = state.stateModel.shareStructureStep.shareClasses
 
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -1,4 +1,4 @@
-import { AccountTypes, CorpTypeCd, FilingCodes, FilingTypes } from '@/enums'
+import { AccountTypes, ActionTypes, CorpTypeCd, FilingCodes, FilingTypes } from '@/enums'
 import {
   IncorporationFilingIF,
   NameRequestDetailsIF,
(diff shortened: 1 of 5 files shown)


andrepestana-aot (approved):
LGTM 👍

src/store/getters/state-getters.ts
```diff
@@ -551,3 +551,14 @@ export const showFeeSummary = (state: StateIF): boolean => {
   )
   return (haveFilingChange && !isEqual(getFilingData(state), defaultFilingData))
 }
+
+/** Return a boolean indicating if the minimum share classes requirements are met. */
+export const invalidMinimumShareClass = (state: StateIF): boolean => {
```
severinbeauvais, line 556: Sorry to be picky but this doesn't return anything. It's a getter, so it "is" a particular value.

Also, the comment is the opposite of what it is 😛

src/store/getters/state-getters.ts
```diff
@@ -551,3 +551,14 @@ export const showFeeSummary = (state: StateIF): boolean => {
...
+  const shareClasses = state.stateModel.shareStructureStep.shareClasses
+
+  // Filter out REMOVED class actions
+  const currentShareClasses = shareClasses.filter(x => x.action !== ActionTypes.REMOVED)
+    .map((x) => { const { action, ...rest } = x; return rest })
```
severinbeauvais, line 561: This is correct but maybe unnecessary... can you remove the `map`, since you only care about the number of them, not actually what they are?
(less work done = faster code)

severinbeauvais (commented):
Good job on the unit tests! 👍

src/store/getters/state-getters.ts
cameron-freshworks: I don't follow? All the getters are a type of value? Whether it is returning a value or `is` a particular value. It still has a type no?

For sure, will fix the comment :)

src/store/getters/state-getters.ts
cameron-freshworks: I think you are right, most likely could! Will verify 👍

cameron-freshworks: Indeed, works good.
Good change and good catch, thanks!

src/store/getters/state-getters.ts
cameron-freshworks: Updated 👍

src/store/getters/state-getters.ts
cameron-freshworks: Updated

src/store/getters/state-getters.ts
cameron-freshworks: Ah.. you were referring the Comment `Returns` 😆 

Fixed!

severinbeauvais: Yup, yup, thanks.

severinbeauvais (approved):

PR #178, opened 2021-04-28
6724 - Alteration UI - Added component validation to CompanyProvisions and ResolutionDates

andrepestana-aot:
*Issue #:* /bcgov/entity#6274

*Description of changes:*

- Added component validation to CompanyProvisions and ResolutionDates
- Added tests

**This is part of 6274 regarding the component validations when the user clicks on "Review and Certify" only.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.6.2",
+  "version": "0.6.3",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.6.3",
+  "version": "0.6.4",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/src/components/Articles/Articles.vue
+++ b/src/components/Articles/Articles.vue
@@ -14,7 +14,7 @@
         :provisionsRemoved="getProvisionsRemoved"
         @companyProvisionsChanged="setProvisionsRemoved($event)"
         @haveChanges="emitHaveChanges($event)"
-        @isEditingCompanyProvisions="setEditingCompanyProvisions($event)"
+        @onEditingCompanyProvisions="setEditingCompanyProvisions($event)"
       />
     </div>
 
@@ -27,6 +27,7 @@
         :isEditMode="true"
         :hasRightsOrRestrictions="getHasRightsOrRestrictions"
         @addRemoveDate="setResolutionDates($event)"
+        @onAddingResolutionDate="setIsAddingResolutionDate($event)"
       />
     </div>
   </v-card>
@@ -54,6 +55,7 @@ export default class Articles extends Mixins(CommonMixin) {
   private companyProvisionsChanges: boolean
 
   private isEditingCompanyProvisions = false
+  private isAddingResolutionDate = false
 
   // Global getters
   @Getter getBusinessInformation!: BusinessInformationIF
@@ -80,12 +82,16 @@ export default class Articles extends Mixins(CommonMixin) {
     this.setValidComponent({ key: 'isValidCompanyProvisions', value: !editing })
   }
 
+  setIsAddingResolutionDate (addingResolutionDate: boolean) {
+    this.isAddingResolutionDate = addingResolutionDate
+  }
+
   private get invalidCompanyProvisions (): boolean {
     return this.getComponentValidate && this.isEditingCompanyProvisions
   }
 
   private get invalidResolutionDates (): boolean {
-    return this.getComponentValidate && !this.getIsResolutionDatesValid
+    return this.getComponentValidate && (this.isAddingResolutionDate || !this.getIsResolutionDatesValid)
   }
 }
 </script>
--- a/src/components/Articles/Articles.vue
+++ b/src/components/Articles/Articles.vue
@@ -5,16 +5,22 @@
       <label class="define-article-title">Articles</label>
     </div>
 
-    <div class="section-container" v-if="getBusinessInformation.hasRestrictions">
+    <div
+      class="section-container"
+      v-if="getBusinessInformation.hasRestrictions"
+      :class="{'invalid-section': invalidCompanyProvisions}">
       <company-provisions
         class="sub-section"
         :provisionsRemoved="getProvisionsRemoved"
         @companyProvisionsChanged="setProvisionsRemoved($event)"
         @haveChanges="emitHaveChanges($event)"
+        @isEditingCompanyProvisions="setEditingCompanyProvisions($event)"
       />
     </div>
 
-    <div class="section-container" :class="{'invalid-section': !getIsResolutionDatesValid}">
+    <div
+      class="section-container"
+      :class="{'invalid-section': invalidResolutionDates}">
       <resolution-dates
         :addedDates="getNewResolutionDates"
         :previousDates="getPreviousResolutionDates"
@@ -47,6 +53,8 @@ export default class Articles extends Mixins(CommonMixin) {
   // Whether components have changes
   private companyProvisionsChanges: boolean
 
+  private isEditingCompanyProvisions = false
+
   // Global getters
   @Getter getBusinessInformation!: BusinessInformationIF
   @Getter getNewResolutionDates!: string []
@@ -54,14 +62,31 @@ export default class Articles extends Mixins(CommonMixin) {
   @Getter getPreviousResolutionDates!: string[]
   @Getter getHasRightsOrRestrictions!: boolean
   @Getter getIsResolutionDatesValid!: boolean
+  @Getter getIsCompanyProvisionsValid!: boolean
+  @Getter getComponentValidate!: boolean
+  @Getter getValidComponentFlags!: boolean
 
   // Global actions
   @Action setProvisionsRemoved!: ActionBindingIF
   @Action setResolutionDates!: ActionBindingIF
+  @Action setValidComponent!: ActionBindingIF
 
   /** Emits Have Changes event. */
   @Emit('haveChanges')
   emitHaveChanges (haveChanges: boolean): void {}
+
+  setEditingCompanyProvisions (editing: boolean) {
+    this.isEditingCompanyProvisions = editing
+    this.setValidComponent({ key: 'isValidCompanyProvisions', value: !editing })
+  }
+
+  private get invalidCompanyProvisions (): boolean {
+    return this.getComponentValidate && this.isEditingCompanyProvisions
+  }
+
+  private get invalidResolutionDates (): boolean {
+    return this.getComponentValidate && !this.getIsResolutionDatesValid
+  }
 }
 </script>
 
--- a/src/components/Articles/Articles.vue
+++ b/src/components/Articles/Articles.vue
@@ -12,9 +12,9 @@
       <company-provisions
         class="sub-section"
         :provisionsRemoved="getProvisionsRemoved"
-        @companyProvisionsChanged="setProvisionsRemoved($event)"
+        @isChanged="setProvisionsRemoved($event)"
         @haveChanges="emitHaveChanges($event)"
-        @onEditingCompanyProvisions="setEditingCompanyProvisions($event)"
+        @isEditing="setEditingCompanyProvisions($event)"
       />
     </div>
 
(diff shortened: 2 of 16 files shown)


src/components/Articles/Articles.vue
```diff
@@ -47,21 +53,40 @@ export default class Articles extends Mixins(CommonMixin) {
   // Whether components have changes
   private companyProvisionsChanges: boolean
 
+  private isEditingCompanyProvisions = false
```
severinbeauvais, line 56: This is the same name as the company provisions component event. While this may not be causing a conflict, it may result in confusion.
How about renaming the event `isEditMode` (same as some of the other components)?

andrepestana-aot: no, I meant to check if the component was left without finish editing. I just followed what we have in YourCompany component: `isEditingNames`, `isEditingType` etc.

src/components/Articles/ResolutionDates.vue
```diff
@@ -193,10 +195,26 @@ export default class ResolutionDates extends Mixins(CommonMixin) {
...
+  @Watch('isAdding')
+  private updateResolutionDatesValidity (): void {
+    if (this.isAdding) {
+      this.setValidComponent({ key: 'isValidResolutionDate', value: false })
+    } else {
+      this.setValidComponent({ key: 'isValidResolutionDate', value: !this.newResolutionDateRequired })
+    }
+  }
+
+  private get newResolutionDateRequired (): boolean {
```
severinbeauvais, line 209: If it makes sense to you, move this (and maybe line 216) to the state-getters module (which is where some other such checks exist).
I'm not saying to change this now... I would say we don't have a complete design for such methods/getters atm so I leave it up to you.

severinbeauvais: Oh, I now see you moved some of this out of state-getters. I'm sure you had a good reason. Ignore comment above and carry on :)

andrepestana-aot: I don't like the idea of adding a lot of logic to state-getters... actually it was in a state-getter and I refactored it to here.

andrepestana-aot: And, yes, I had a good reason. We need to discuss.

tests/unit/ResolutionDates.spec.ts
```diff
@@ -271,6 +271,25 @@ describe('Resolution Dates component - edit mode', () => {
...
+    const rows = wrapper.findAll('.row')
+    expect(rows.length).toBe(1)
+
+    const cols = rows.at(0).findAll('.col')
+    expect(cols.length).toBe(3)
+
+    const button = cols.at(2).find('.add-btn')
+    button.trigger('click')
+    await Vue.nextTick()
+    expect(store.state.stateModel.newAlteration.validComponents.isValidResolutionDate).toBe(false)
```
severinbeauvais, line 288: Great test!

src/components/Articles/Articles.vue
cameron-freshworks: +1 to Severins comment. 
In this case it makes sense to rename the component event to something more generic, as we KNOW that event belongs to the component, where the local property needs to be specific.

severinbeauvais: The local variable name is fine, but I think we should have consistent event names.

src/components/Articles/ResolutionDates.vue
cameron-freshworks: I disagree, i think it makes sense for this getter to be global, as you are now pulling shareStructure information into this component.

I think using getters to compute some value is fine as long as it's specific and clear in its intent. 
My preference is to keep the components as simple as possible (i know it's tough sometimes), but moving this logic into the component only achieves 1 thing imo and that is making in accessible to anywhere else.

severinbeauvais: As I said, we don't have a complete design for this :)

What do you (y'all) think of a mixin for methods with store logic?

andrepestana-aot: > I disagree, i think it makes sense for this getter to be global, as you are now pulling shareStructure information into this component.
Yes, I agree that this is not ideal dealing with shareStructure here.

> My preference is to keep the components as simple as possible (i know it's tough sometimes), but moving this logic into the component only achieves 1 thing imo and that is making in accessible to anywhere else.
I agree on keeping the components as simple as possible. I think we can find another approach other than using state-getters because when you do it, you are overriding the real state (the state will show is `isValidResolutionDate === true` but the getter may result in false for example)

Creating a mixin is a better alternative imo.

severinbeauvais: > I think using getters to compute some value is fine as long as it's specific and clear in its intent.
My preference is to keep the components as simple as possible (i know it's tough sometimes), but moving this logic into the component only achieves 1 thing imo and that is making in accessible to anywhere else.

I strongly agree with this.

However, atm it's a guideline and not a strong design.

cameron-freshworks (commented):
I've noticed a few things when running locally. 

The validation is still allowing you to proceed through to Review mode even if your Resolution Date is invalid. 

Also, the border isn't appearing on the invalid resolution date section.

Screen Shot 2021-04-28 at 10 35 11 AM

Screen Shot 2021-04-28 at 10 35 17 AM

src/components/Articles/ResolutionDates.vue
cameron-freshworks: I don't fully understand how a getter is overriding real* state, where the computed value is merely a representation of what the store values are. 

Mixin's are great for sure, but are purposed for used/extended by components. 
I think we've used getters to compute some values in the past because the easy access to store, aka `state`.

There is always a better way for sure, i'm 100% open to trying anything!

severinbeauvais: Andre, why don't you try something that makes sense to you (looks/feels like good design), and we can review it and hopefully follow along with our later work?

andrepestana-aot:
> I've noticed a few things when running locally.
> 
> The validation is allowing you to proceed through to Review mode even if your Resolution Date is invalid.

How do I test the Review mode?

> Also, the border isn't appearing on the invalid resolution date section.
> 
> The invalid text under the Review and certify isn't appearing either.

I see what you're saying but this is before you click on "Review and Certify". When you click on "Review and Certify" it will trigger the component validations, and then, the border and the message will be shown. I thought that this would be the behaviour expected.

What's the behaviour you expect in this case?

src/components/Articles/Articles.vue
andrepestana-aot: >The local variable name is fine, but I think we should have consistent event names.

Do you guys think I should change the name? I followed what we have for other components like in YourCompany.

severinbeauvais: Yes, please change the event name (not the variable name).

cameron-freshworks:
> > I've noticed a few things when running locally.
> > The validation is allowing you to proceed through to Review mode even if your Resolution Date is invalid.
> 
> How do I test the Review mode?
> 
> > Also, the border isn't appearing on the invalid resolution date section.
> > The invalid text under the Review and certify isn't appearing either.
> 
> I see what you're saying but this is before you click on "Review and Certify". When you click on "Review and Certify" it will trigger the component validations, and then, the border and the message will be shown. I thought that this would be the behaviour expected.
> 
> What's the behaviour you expect in this case? Right away after the shareStructure is added it should show the red border? The message is only shown after clicking on the button.

Review mode is the second screen, once you've made changes you can click Review and Confirm, and if all is valid you can see the review/summary mode. Currently you are able to proceed even if the resolution date section is invalid. 

You are right that you won't see the text under the Review and Certify button until clicked, but in this case it isn't showing even after clicking. 

With Resolution Date you are able to be invalid before clicking review and certify, for example if you Add a share class with Special rights and restrictions and add a DATE, but then remove the date, you will be in an invalid state, which should prevent proceeding. 

Expected behaviour is the Resolution Date section should be invalid if you remove the date required for any special rights or restrictions.

andrepestana-aot:
@cameron-freshworks Thanks for pointing out the validation issue!

I decided to keep the logic in the state-getter for now. We can think of a better approach later. I also fixed issues with the scrolling. I checked all the scrolling and there's still something to be improved for shareStructure because all the scrolling is going to the start of the red border but as shareStructure has a dynamic number of divs, for this one the scroll takes the user to the top of the section.

src/components/Articles/Articles.vue
```diff
@@ -5,22 +5,29 @@
...
+    <div
+      class="section-container"
+      v-if="getBusinessInformation.hasRestrictions"
+      :class="{'invalid-section': invalidCompanyProvisions}">
       <company-provisions
         class="sub-section"
         :provisionsRemoved="getProvisionsRemoved"
         @companyProvisionsChanged="setProvisionsRemoved($event)"
         @haveChanges="emitHaveChanges($event)"
+        @onEditingCompanyProvisions="setEditingCompanyProvisions($event)"
```
severinbeauvais, line 17: I was hoping the event would simply be called "isEditing" (and same for all components).

(Admittedly I would prefer the event on line 15 to be called "isChanged". But nvm that.)

The "on" naming is usually for event handlers, eg,
`@isEditing="onEditingDoSomething($event)"`

andrepestana-aot: ok. I can change to `isEditing`. 👍

src/components/Articles/Articles.vue
```diff
@@ -5,22 +5,29 @@
...
+    <div
+      class="section-container"
+      :class="{'invalid-section': invalidResolutionDates}">
       <resolution-dates
         :addedDates="getNewResolutionDates"
         :previousDates="getPreviousResolutionDates"
         :isEditMode="true"
         :hasRightsOrRestrictions="getHasRightsOrRestrictions"
         @addRemoveDate="setResolutionDates($event)"
+        @onAddingResolutionDate="setIsAddingResolutionDate($event)"
```
severinbeauvais, line 30: Again, this is really an event telling us if the component is in editing mode, right? So I would use the same verb as above: `@isEditing="..."`

What I'm doing here is taking a step back and looking at things... the Articles component doesn't care about the details of Resolution Dates component. It only cares that Resolution Dates is in edit mode. So that's all I would keep track of. And if we can make the component event handers consistent then that will be helpful for someone to generally handle the code. This project is almost big enough now that I don't understand everything in it, so keep my "immediate concerns" in scope and not thinking about what sub-components do is very helpful.

src/components/Articles/Articles.vue
severinbeauvais: Thanks. That will be cleaner.

Also, other events can be "isValid", "isChanged" (or "hasChanges" or "hasChanged")... yeah we're still working on naming :)

Re: hardest problems in computer science - https://martinfowler.com/bliki/TwoHardThings.html

src/components/Articles/Articles.vue
andrepestana-aot: Yes, I understand what you're saying but I need to know if the subcomponent isEditing to set the invalid state for the resolution-dates component. editMode is another thing related to check if the component is "editable".

severinbeauvais: isAdding ?

andrepestana-aot: yes, I changed onAddingResolutionDate to isEditing to keep it consistent to the other components.

severinbeauvais (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.6.2",
+  "version": "0.6.3",
```
cameron-freshworks, line 3: i believe `main` is now at `6.3` so this might require another bump up

cameron-freshworks:
> @cameron-freshworks Thanks for pointing out the validation issue!
> 
> I decided to keep the logic in the state-getter for now. We can think of a better approach later. I also fixed issues with the scrolling. I checked all the scrolling and there's still something to be improved for shareStructure because all the scrolling is going to the start of the red border but as shareStructure has a dynamic number of divs, for this one the scroll takes the user to the top of the section.

No prob! I ran this locally as well and it all seems to work great 👍 Good job here, it is a tricky bit :)

cameron-freshworks (approved):
Looks good, great job 👍 

Just check the version # before merge, might need to bump it up and `npm i` to update your package-lock.

package.json
andrepestana-aot: 👍

andrepestana-aot:
@cameron-freshworks Please merge if everything is ok. Thanks!

PR #182, opened 2021-04-29
6724 - EffectiveDateTime to enable all the fields when selecting FED

andrepestana-aot:
*Issue #:* /bcgov/entity#6724

*Description of changes:*

- Changed EffectiveDateTime to enable all the fields when selecting FED;
- Updated tests

**This PR is part of 6724 regarding the checkbox: "Can we have the users able to select date/time/am in any order or is there a reason we force users to select the date field first? Would prefer all the fields are "active" when you click the FED radio button"**

### DIFF
--- a/src/components/common/EffectiveDateTime.vue
+++ b/src/components/common/EffectiveDateTime.vue
@@ -266,7 +266,7 @@ export default class EffectiveDateTime extends Mixins(DateMixin) {
     // wait for form to update itself before checking validity
     await Vue.nextTick()
 
-    const isDateValid = this.$refs.datePickerRef.$refs.form['validate']()
+    const isDateValid = this.$refs.datePickerRef.validateForm()
     const isTimeValid = this.$refs.form.validate()
     if (isDateValid && isTimeValid) {
       const year = +this.dateText.slice(0, 4)
@@ -363,7 +363,7 @@ export default class EffectiveDateTime extends Mixins(DateMixin) {
       this.selectHour = []
       this.selectMinute = []
       this.selectPeriod = PeriodTypes.AM
-      this.$refs.datePickerRef.$refs.form['reset']()
+      this.$refs.datePickerRef.clearDate()
     }
 
     // update the parent
@@ -389,7 +389,7 @@ export default class EffectiveDateTime extends Mixins(DateMixin) {
 
     // wait for form to update itself before checking validity
     await Vue.nextTick()
-    const isDateValid = this.$refs.datePickerRef.$refs.form['validate']()
+    const isDateValid = this.$refs.datePickerRef.validateForm()
     const isTimeValid = this.$refs.form.validate()
     return (!!this.effectiveDateType &&
       isDateValid && isTimeValid &&
--- a/src/components/common/EffectiveDateTime.vue
+++ b/src/components/common/EffectiveDateTime.vue
@@ -7,6 +7,7 @@
 
     <v-form ref="form" class="date-time-selectors">
       <date-picker
+        ref="datePickerRef"
         title="Date"
         nudge-right="40"
         :inputRules="dateRules"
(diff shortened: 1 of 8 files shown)


src/components/common/EffectiveDateTime.vue
```diff
@@ -261,7 +266,9 @@ export default class EffectiveDateTime extends Mixins(DateMixin) {
     // wait for form to update itself before checking validity
     await Vue.nextTick()
 
-    if (this.$refs.form.validate()) {
+    const isDateValid = this.$refs.datePickerRef.$refs.form['validate']()
```
severinbeauvais, line 269: What do you think about exposing new methods `validate` and `reset` on the DatePicker component instead?

tests/unit/EffectiveDateTime.spec.ts
```diff
@@ -122,10 +122,9 @@ describe('Effective Date Time component', () => {
     const radioIsFutureEffective = radioInput.at(1)
 
     await radioIsFutureEffective.trigger('click')
-    wrapper.vm.dateText = new Date().toISOString().split('T')[0]
```
severinbeauvais, line 125: Since you've removed this, is there another test in this file that verified a correct date (or no input, or whatever this previously tested)?

src/components/common/EffectiveDateTime.vue
andrepestana-aot: It has the same effect. This is way is faster and keep the component leaner.

severinbeauvais: Yes but we lose typing. And it looks hacky.

I'd rather not try to save a little time today and hope to fix it some day... In my experience, we will be just as busy "some day" as now and it may never get fixed.

tests/unit/EffectiveDateTime.spec.ts
andrepestana-aot: There was another place that set dateText but didn't make any difference for the test validity. I removed it as well.

src/components/common/EffectiveDateTime.vue
andrepestana-aot: ok. Will do it.

tests/unit/EffectiveDateTime.spec.ts
severinbeauvais: Does this component now need a new test to verify a valid date?

andrepestana-aot: It has other tests that test for the valid date.

severinbeauvais: ok

src/components/common/EffectiveDateTime.vue
andrepestana-aot: @severinbeauvais I've been looking into it but I couldn't find a way to expose the component's methods other than hacky ways to do it. What do you suggest?

severinbeauvais: Is it possible to call `this.$refs.myComponent.myMethod()`?

If not, leave it alone for now. Good luck :)

andrepestana-aot: I tested here and checked that if we change the method's visibility to 'public' it lets you to use it as you are referring to:
https://github.com/bcgov/bcrs-shared-components/blob/master/src/components/DatePicker/DatePicker.vue#L87

Then you can call:
`this.$refs.datePickerRef.clearDate()`

severinbeauvais (approved):
GTG after your final changes.

cameron-freshworks (approved):
g2g after updating the component version 👍 
Let me know if you need a hand publishing changes to the shared component.

src/interfaces/utils-interfaces/form-type.ts
```diff
@@ -4,3 +4,5 @@ export interface FormType extends HTMLFormElement {
   validate(): boolean
   resetValidation(): void
 }
+// To keep compatibility with the shared components that use FormIF and not FormType
+export interface FormIF extends FormType {}
```
severinbeauvais, line 8: What would it take to change all "FormType" references to "FormIF" instead in this project?

andrepestana-aot: ok. I'll do it.

PR #185, opened 2021-05-05
Staff Payment and Office Address Bugs

cameron-freshworks:
*Issue #:* /bcgov/entity#7416

*Description of changes:*

* Include Staff payment in filing and draft parsing.

* Added Staff payment to alteration filing build.

* Created method to parse staff payment into store for both alterations and corrections.

* Fixed Office addresses bug where it wouldn't initialize the data after changing views, would only update the component when the Store values changed, which only happened once in Alterations.

* verified resolution dates are working as expected.

* Minor unit test updates to include the additional event.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.6.7",
+  "version": "0.6.8",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.6.6",
+  "version": "0.6.7",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/src/components/YourCompany/OfficeAddresses.vue
+++ b/src/components/YourCompany/OfficeAddresses.vue
@@ -354,6 +354,11 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
   /** Model value for "same as (records) mailing address" checkbox. */
   private inheritRecMailingAddress: boolean = true
 
+  /** Use 'mounted' lifecycle hook to ensure component is kept up to date after changing views. */
+  mounted (): void {
+    this.updateAddresses()
+  }
+
   /** The office addresses from the original IA. NB: may be {} */
   private get originalOfficeAddresses (): IncorporationAddressIf {
     return (this.getOriginalIA.incorporationApplication.offices as IncorporationAddressIf)
@@ -624,7 +629,7 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
    * sets local properties and emits state events.
    */
   @Watch('getOfficeAddresses', { deep: true })
-  private updateAddresses (addresses: IncorporationAddressIf): void {
+  private updateAddresses (): void {
     // set local properties from store
     this.setLocalProperties()
 
--- a/src/components/YourCompany/OfficeAddresses.vue
+++ b/src/components/YourCompany/OfficeAddresses.vue
@@ -620,7 +620,7 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
   }
 
   /**
-   * When stored office addresses change (ie, when data is loaded/updated/reset),
+   * When the component is mounted or stored office addresses change (ie, when data is loaded/updated/reset),
    * sets local properties and emits state events.
    */
   @Watch('getOfficeAddresses', { deep: true, immediate: true })
@@ -629,6 +629,7 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
     this.setLocalProperties()
 
     // update external state
+    // Will emit on component mount and subsequent changes
     this.emitHaveChanges()
   }
 
--- a/src/components/YourCompany/OfficeAddresses.vue
+++ b/src/components/YourCompany/OfficeAddresses.vue
@@ -354,11 +354,6 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
   /** Model value for "same as (records) mailing address" checkbox. */
   private inheritRecMailingAddress: boolean = true
 
-  /** Use 'mounted' lifecycle hook to ensure component is kept up to date after changing views. */
-  mounted (): void {
-    this.updateAddresses()
-  }
-
   /** The office addresses from the original IA. NB: may be {} */
   private get originalOfficeAddresses (): IncorporationAddressIf {
     return (this.getOriginalIA.incorporationApplication.offices as IncorporationAddressIf)
@@ -628,7 +623,7 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
    * When stored office addresses change (ie, when data is loaded/updated/reset),
    * sets local properties and emits state events.
    */
-  @Watch('getOfficeAddresses', { deep: true })
+  @Watch('getOfficeAddresses', { deep: true, immediate: true })
   private updateAddresses (): void {
     // set local properties from store
     this.setLocalProperties()
--- a/src/components/YourCompany/OfficeAddresses.vue
+++ b/src/components/YourCompany/OfficeAddresses.vue
@@ -622,6 +622,9 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
   /**
    * When the component is mounted or stored office addresses change (ie, when data is loaded/updated/reset),
    * sets local properties and emits state events.
+   *
+   * Called on mount to init data in the event the component is destroyed and there is no changes to trigger the Watcher
+   * This happens in Alterations when navigating between views and where office addresses are not editable.
    */
   @Watch('getOfficeAddresses', { deep: true, immediate: true })
   private updateAddresses (): void {
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -283,6 +283,29 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     // Include name request info when applicable
     if (this.hasNewNr || this.hasBusinessNameChanged) filing.alteration.nameRequest = { ...this.getNameRequest }
 
+    // Populate Staff Payment according to payment option
+    switch (this.getStaffPayment.option) {
+      case StaffPaymentOptions.FAS:
+        filing.header.routingSlipNumber = this.getStaffPayment.routingSlipNumber
+        filing.header.priority = this.getStaffPayment.isPriority
+        break
+
+      case StaffPaymentOptions.BCOL:
+        filing.header.bcolAccountNumber = this.getStaffPayment.bcolAccountNumber
+        filing.header.datNumber = this.getStaffPayment.datNumber
+        filing.header.folioNumber = this.getStaffPayment.folioNumber // this overrides original folio number
+        filing.header.priority = this.getStaffPayment.isPriority
+        break
+
+      case StaffPaymentOptions.NO_FEE:
+        filing.header.waiveFees = true
+        filing.header.priority = false
+        break
+
+      case StaffPaymentOptions.NONE: // should never happen
+        break
+    }
+
     return filing
   }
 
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.6.6",
+  "version": "0.6.7",
```
severinbeauvais, line 3: This will be 0.6.8 now that Andre's PR is merged.

src/components/YourCompany/OfficeAddresses.vue
```diff
@@ -624,7 +629,7 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
    * sets local properties and emits state events.
    */
   @Watch('getOfficeAddresses', { deep: true })
```
severinbeauvais, line 631: Why not use `immediate: true`  to fire this method up front?

src/mixins/filing-template-mixin.ts
```diff
@@ -283,6 +283,29 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     // Include name request info when applicable
     if (this.hasNewNr || this.hasBusinessNameChanged) filing.alteration.nameRequest = { ...this.getNameRequest }
 
+    // Populate Staff Payment according to payment option
```
severinbeauvais, line 286: This looks identical to the code for corrections... could this use a new helper method?

src/mixins/filing-template-mixin.ts
```diff
@@ -611,4 +601,48 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     }
     this.setBusinessContact(contactPoint)
   }
+
+  /** Parse Staff Payment data into store.
+   * @param filing The alteration or correction filing to parse.
+   */
+  storeStaffPayment (filing: AlterationFilingIF | CorrectionFilingIF): void {
```
severinbeauvais, line 608: Private.

src/components/YourCompany/OfficeAddresses.vue
cameron-freshworks: It's fired up front already. The issue is during the re-mount ie changing to review and back to company changes. 
Would immediate fire it on a re-mount?

package.json
cameron-freshworks: oh nice, thanks for the heads up. Will update!

tests/unit/OfficeAddresses.spec.ts
```diff
@@ -957,7 +957,7 @@ describe('actions and events', () => {
 
     // verify initial events
     const haveChanges = wrapper.emitted('haveChanges')
-    expect(haveChanges.length).toBe(1)
+    expect(haveChanges.length).toBe(2)
```
severinbeauvais, line 960: Why are there 2 initial events -- what are they?

src/mixins/filing-template-mixin.ts
cameron-freshworks: It is! Yes you bet, i did the same for the parse, should have done this for the build 😁

src/mixins/filing-template-mixin.ts
cameron-freshworks: indeed, will do!

tests/unit/OfficeAddresses.spec.ts
cameron-freshworks: Mounted lifecycle and the Watcher

src/components/YourCompany/OfficeAddresses.vue
severinbeauvais: There is no such thing as a remount. The component is actually being re-loaded/re-rendered.

(This is part of why NR was nasty -- the components are always (mostly) rendered, and conditionally displayed. In that case you have to catch the display event and init there.)

severinbeauvais: Short answer: yes, when the component is reloaded, all immediate watchers should fire.

What is firing this up front already? Why doesn't that do what it's supposed to with the data?

tests/unit/OfficeAddresses.spec.ts
severinbeauvais: If the watcher fires when `getOfficeAddresses` changes, why do you need to call it from mounted() ?

src/components/YourCompany/OfficeAddresses.vue
cameron-freshworks: > There is no such thing as a remount. The component is actually being re-loaded/re-rendered.
> 
> (This is part of why NR was nasty -- the components are always (mostly) rendered, and conditionally displayed. In that case you have to catch the display event and init there.)

I'm not sure what you want to call it, re-mounting seemed most accuratre. The component itself is mounted every time you change to Review and back to the components.

It fires up front because the store values change with regards to the office addresses, every time a filing is loaded. Why it works initially but doesn't when it 'mounts again'. Because there is no changes to the store value at that point.

tests/unit/OfficeAddresses.spec.ts
severinbeauvais: (Not that I've never done it before, but it's a code smell when a component emits the same event multiple times on init / when nothing has happened yet.)

src/components/YourCompany/OfficeAddresses.vue
severinbeauvais: > I'm not sure what you want to call it, re-mounting seemed most accuratre. The component itself is mounted every time you change to Review and back to the components.

Haha, I agree with you there. Let's call it "remounted" then.

My point is that this component is destroyed between mounted calls. It's not just dormant and re-initialized.

severinbeauvais: What if this component wasn't rendered (mounted!) until the data is loaded...?

cameron-freshworks: Totally, the issue was purely that the initialization was based on a watcher (office address store data) which only changes once in an Alteration. Not an issue in corrections because the single view, the component is never destroyed.

I think why this isn't working like our other components, is the value fed into the Base Address Components isn't exactly reactive, it is computed values which the component is internally doing, and in this case the trigger for that is a change.

I'll try `immediate` though. Would much prefer that as well!

tests/unit/OfficeAddresses.spec.ts
cameron-freshworks: I agree, see the comment above 😁 I can elaborate more though, if need be,

src/components/YourCompany/OfficeAddresses.vue
cameron-freshworks: immediate works great, thanks! 

I didn't know this option did this, assumed it tied into `created` hook, this is really good to know 👍

package.json
cameron-freshworks: Updated

tests/unit/OfficeAddresses.spec.ts
andrepestana-aot: Yeah, just to give my 2 cents: maybe checking in the watch if the value has really changed may resolve it if it's because of the mount that triggers it. I'm not sure but there may be any option to discard it.

src/mixins/filing-template-mixin.ts
cameron-freshworks: Updated

src/mixins/filing-template-mixin.ts
cameron-freshworks: updated

tests/unit/OfficeAddresses.spec.ts
cameron-freshworks: With the inclusion of `immediate` the changes still stand as it will call the method an additional time.
When the component is mounted and when the watcher is triggered, ie any changes to the office addresses.

severinbeauvais: A small comment where the component emits the event, to describe the init sequence (and thus why there are multiple events), would be great! ❤️

cameron-freshworks: Will do 👍

src/mixins/filing-template-mixin.ts
```diff
@@ -611,4 +561,78 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
+      case StaffPaymentOptions.NO_FEE:
+        filing.header.waiveFees = true
+        filing.header.priority = false
+        break
+
+      case StaffPaymentOptions.NONE: // should never happen
+        break
+    }
+
+    return filing
```
severinbeauvais, line 592: Theoretically this isn't needed, because the code above changes the original object. (If you go this route, you'd have to update lines 170 and 267.)

severinbeauvais: The object is passed by reference I think (like a pointer to an object in C).

severinbeauvais: Or maybe not.
https://stackoverflow.com/questions/13104494/does-javascript-pass-by-reference

severinbeauvais: https://www.geeksforgeeks.org/pass-by-value-and-pass-by-reference-in-javascript/

I'm going to assume the above are correct and resolve this conversation with no change required ;)

src/components/YourCompany/OfficeAddresses.vue
```diff
@@ -620,15 +620,16 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
   }
 
   /**
-   * When stored office addresses change (ie, when data is loaded/updated/reset),
+   * When the component is mounted or stored office addresses change (ie, when data is loaded/updated/reset),
    * sets local properties and emits state events.
```
severinbeauvais, line 624: Can you also explain why this component emits this event twice on init?

src/mixins/filing-template-mixin.ts
cameron-freshworks: Hmm.. i thought it would be. 
If it isn't it would be cleaner and preferable. I will try!

severinbeauvais (approved):
Ready to merge when you are.

src/components/YourCompany/OfficeAddresses.vue
cameron-freshworks: Doesn't this already explain that? 
How would you word it?

andrepestana-aot (approved):
👍 Looks good to me!

src/components/YourCompany/OfficeAddresses.vue
cameron-freshworks: It only emits once on init (mounted due to immediate) and it emits again once the data changes (in our case, when the filing is loaded, and any subsequent changes). 

hehe i can write that.. though it seems long winded. I need your 'comment' expertise here!

src/mixins/filing-template-mixin.ts
cameron-freshworks: Yea.. works good! I was worried the filing would return without the changes.
Good clean up, thanks!

src/components/YourCompany/OfficeAddresses.vue
severinbeauvais: I'm not looking for "what" it's doing (ie, immediate + watching) but "why".

> It fires up front because the store values change with regards to the office addresses, every time a filing is loaded. Why it works initially but doesn't when it 'mounts again'. Because there is no changes to the store value at that point.

This sort of explained it for me but I will soon forget it, so hoping to get this into a code comment.

I guess the gist is that this component loads/inits/mounts before the data is loaded, and also after. Or something about the reactivity not set up to work as intended. I dunno.

cameron-freshworks: I'll try my best to do so without getting too wordy :)

severinbeauvais: > It only emits once on init (mounted due to immediate) and it emits again once the data changes (in our case, when the filing is loaded, and any subsequent changes).

Yeah, this.

But now I'm wondering why this component is displayed before it had data 😛  (IMHO it shouldn't.) But that can be reworked another day.

src/mixins/filing-template-mixin.ts
severinbeauvais: OK so now I'm confused why it works 🤷

src/components/YourCompany/OfficeAddresses.vue
cameron-freshworks: It's an internal issue in the component i think.
The component computes the values which it then feeds into the Base Address Sub component. This was only being called when the Store values changed. 
I think if the Base Addres sub components were utilizing the store getter directly it wouldn't have been an issue, but the Parent component is doing a local computation of sorts.

src/mixins/filing-template-mixin.ts
cameron-freshworks: Like you said, the javascript reference :) 

Tbh i dunno either, it does mutate the original values though.

severinbeauvais: Maybe this?
https://dev.to/bytebodger/a-gotcha-of-javascript-s-pass-by-reference-1afm

severinbeauvais: TLDR

PR #207, opened 2021-09-13
Alteration Structural Changes

cameron-freshworks:
*Issue #:* /bcgov/entity#8934

*Description of changes:*

* Allow a draft alteration to be loaded blank.
* Only Include changed sections into Alteration filing.

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -393,7 +393,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     this.setBusinessSnapshot(businessSnapshot)
 
     // Restore current entity type
-    this.setEntityType(filing.alteration.business.legalType)
+    this.setEntityType(filing.alteration.business?.legalType || businessSnapshot.businessInfo.legalType)
 
     // Restore business information
     this.setBusinessInformation({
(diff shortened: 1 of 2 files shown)


src/mixins/filing-template-mixin.ts
```diff
@@ -218,21 +222,11 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
-        },
-        nameTranslations: nameTranslations,
-        shareStructure: {
-          resolutionDates: this.getNewResolutionDates,
-          shareClasses
-        },
+        provisionsRemoved: this.getProvisionsRemoved,
         contactPoint: {
           email: this.getBusinessContact.email,
           phone: this.getBusinessContact.phone,
```
andrepestana-aot, line 232: It's not related to this change but can you please test if the `contactPoint.extension` can be null here? I changed it in the create-ui. Maybe the same change can/should be done here.

andrepestana-aot (approved):
Looks good! Just a comment about the `contactPoint.extension` because the legal-api may validate it when it is null. According to the schema it is option and should be number only. Thanks!

src/mixins/filing-template-mixin.ts
cameron-freshworks: Good call, i will have a look!
I bumped into the same thing in the ShareStructure but noticed you had already fixed it in the Shared Component, Gj! 👍

vysakh-menon-aot (approved):

src/mixins/filing-template-mixin.ts
cameron-freshworks: Yessir, no issues here. It will pass as undefined if there is no Extension.

cameron-freshworks: Tried without a phone number as well, since it's pulled from the account and could potentially be null. All good though 👍

andrepestana-aot: Thanks!

PR #236, opened 2022-02-18
Edit UI framework/configuration updates

cameron-freshworks:
*Issue #:* /bcgov/entity#10613

*Description of changes:*
* Updated configurations for Corrections/Alterations/Change filings
* Updated configurations for the existing supported entities + SP/GP
* Added template view for Change filings
* Updated existing reusable components with configurations for basic display in Change Filings
* Basic filing/snapshot templating for Firms (SCHEMA and Business content still pending, so this is just to initialize)
* Extensive directory restructure to separate filing/entity types


src/components/Correction/index.ts
```diff
@@ -0,0 +1 @@
+export * from './IncorporationAgreement'
```
severinbeauvais, line 1: FWIW, I went the other way here and got rid of most index.ts for components. In Create UI this got rid of circular / nested imports and some weirdness.

But this change may be fine here. Up to you as things evolve how you want to go. (It's easy to change to one way or the other.)

severinbeauvais: The reason I reorganized things in Create UI was to make it easy(er) to tell what components a filing type uses, and also what filing types a component is used by.

src/components/common/YourCompany/SummaryDefineCompany.vue
```diff
@@ -60,8 +60,8 @@ import { Component, Mixins, Prop } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 
 // Components
-import { FolioInformation, BusinessContactInfo, OfficeAddresses } from '@/components/YourCompany'
-import { CorrectNameOptions } from '@/components/YourCompany/CompanyName'
+import { FolioInformation, BusinessContactInfo, OfficeAddresses } from '@/components/common/YourCompany/index'
```
severinbeauvais, line 63: I don't think you need to specify `index`. I also saw this elsewhere (search for "/index") if you choose to clean this up.

src/interfaces/change-firm-interfaces/change-firm-interface.ts
```diff
@@ -0,0 +1,8 @@
+import { FilingBusinessIF, FilingHeaderIF } from '@/interfaces'
+
+/** Interface for data object UI sends to API. */
+export interface ChangeFirmIF {
+  header: FilingHeaderIF
+  business: FilingBusinessIF
+  changeFirm: any
```
severinbeauvais, line 7: No type here? Or union type (A | B) ?

src/interfaces/store-interfaces/state-model-interface.ts
```diff
@@ -26,6 +26,7 @@ export interface StateModelIF {
   originalIA: IncorporationFilingIF
   originalAlteration: AlterationFilingIF
   businessSnapshot: BusinessSnapshotIF
+  firmSnapshot: any
```
severinbeauvais, line 29: Type? (If you have a good reason why not then please add a short comment explaining.)

src/resources/index.ts
```diff
@@ -1,4 +1,21 @@
...
+]
+
+export const ChangeFirmResources: Array<ResourceIF> = [
+  SoleProprietorshipResource,
+  GeneralPartnershipResource
+]
+
+export const CorrectionResources: Array<any> = [
+  BenefitCompanyStatementResource
+]
```
severinbeauvais, line 21: Interesting way to do this. I like it!

src/components/Correction/index.ts
cameron-freshworks: Sounds good, will have a closer look at this! I see what you mean

src/utils/feature-flags.ts
```diff
@@ -9,7 +9,8 @@ declare const window: any
  */
 const defaultFlagSet: LDFlagSet = {
   'alteration-ui-enabled': false,
-  'correction-ui-enabled': false
+  'correction-ui-enabled': false,
+  'change-ui-enabled': false
```
severinbeauvais, line 13: Are any of the old flags now obsolete? (Would we ever want to disable alterations or corrections in Prod?)

src/interfaces/change-firm-interfaces/change-firm-interface.ts
cameron-freshworks: Not yet, there is a few places that i left the Type undefined as we didn't have a model/schema for Sp/Gps quite yet at the time of writing this, so i opted to leave it until we did know. 

In the process of creating some Sp/Gp records to help model out these entity types that don't/didn't currently exist.

src/interfaces/store-interfaces/state-model-interface.ts
cameron-freshworks: oh yea, same comment as above. I don't know exactly what the sp/gp model will look like, was going to update once we had some business records and a schema for the filing 😁

src/utils/feature-flags.ts
cameron-freshworks: Good question, i doubt it! 
Lets have a closer look at this but my first thought is that we can remove the two older flags 👍

severinbeauvais: On second thought, these 3 flags now control the 3 types of filings that this UI supports. Let's leave these in a little longer (in case we break anything or whatever).

severinbeauvais: (Kind of like a maintenance page that would pre-empt an entire UI, these flags would allow us to disable filing types individually. It feels like something we might want to leave in long-term.)

cameron-freshworks: Good call, certainly wouldn't hurt!

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -5,7 +5,8 @@
     <template v-if="!isEditing">
       <v-layout row id="summary-registered-address" class="mx-0">
         <v-flex xs3>
-          <label>Registered Office</label>
+          <label v-if="isChangeFiling">Business Addresses</label>
+          <label v-else>Registered Office</label>
```
severinbeauvais, line 9: I'm thinking this should be a resource.

lekshmimallika-aot (approved):

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: For sure, can do.
Seemed like a pretty small thing, so i wasn't sure at the time.

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -404,7 +406,7 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
    * Sets local address data and "inherit" flags from store.
    */
   private setLocalProperties (): void {
-    if (this.getOfficeAddresses.registeredOffice) {
+    if (this.getOfficeAddresses?.registeredOffice) {
```
severinbeauvais, line 409: This should not be necessary, since `state.stateModel.officeAddresses` is initialized as `{}`.

But this raises the question, should `officeAddresses` be initialized to null or an object with empty fields instead? It depends how it's used. It may not matter here. FWIW, both are used in Create UI:

https://github.com/bcgov/business-create-ui/blob/7404fcbbccccced5736ad316294daa86b80153a9/src/store/state/state-model.ts#L33

https://github.com/bcgov/business-create-ui/blob/7404fcbbccccced5736ad316294daa86b80153a9/src/store/state/state-model.ts#L168

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -7,16 +7,18 @@
 
     <div class="define-company-header">
       <v-icon color="appDkBlue">mdi-domain</v-icon>
-      <label class="define-company-title">Your Company</label>
+      <label class="define-company-title">Your {{ getResource.entityReference }}</label>
```
severinbeauvais, line 10: I would include "Your" in the resource, too.

(In case UX decides to change this to "My Company" -- haha -- or a number of other possibilities.)

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -7,16 +7,18 @@
...
     </div>
 
     <div id="company-name-section" class="section-container" :class="{'invalid-section': invalidNameSection}">
       <!-- Company Name -->
       <v-row no-gutters class="mt-4">
         <v-col cols="3">
-          <label :class="{'error-text': invalidNameSection}"><strong>Company Name</strong></label>
+          <label :class="{'error-text': invalidNameSection}">
+            <strong>{{ getResource.entityReference }} Name</strong>
```
severinbeauvais, line 18: Oh, you use it here, too.

Well, how about a new resource string for this?

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: That 'never' happens 😆

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: For sure!

src/store/getters/state-getters.ts
```diff
@@ -49,6 +48,17 @@ export const isAlterationFiling = (state: StateIF): boolean => {
...
+/** Whether the current filing is a change filing. */
+export const isChangeFiling = (state: StateIF): boolean => {
+  return (state.stateModel.tombstone.filingType === FilingTypes.CHANGE_FIRM)
+}
+
+/** Whether the current filing is an Edit filing. */
+export const isEditFiling = (state: StateIF): boolean => {
+  return (state.stateModel.tombstone.filingType === FilingTypes.CHANGE_FIRM) ||
+    (state.stateModel.tombstone.filingType === FilingTypes.ALTERATION)
+}
```
severinbeauvais, line 60: Naming things is hard.

I originally wrote that I didn't truly understand the difference between a "change" filing and an "edit" filing. And after looking closer, I see that I was even more confused than I thought. I see now that an edit filing is a superset of a change filing. My spidey senses are tingling!

I recommend that you replace isEditFiling with `isChangeFiling || isAlterationFiling`. This is more verbose but also more clear.

cameron-freshworks: Can do but it also brings into question the directory structure..

src/enums/nameRequestTypes.ts
```diff
@@ -1,7 +1,20 @@
...
+/** Enum to sync NR types with corp types. */
+export enum NameRequestEntityTypes {
+  BC = CorpTypeCd.BC_COMPANY,
+  CR = CorpTypeCd.BC_CORPORATION, // SPECIAL NAMEREQUEST-ONLY ENTITY TYPE
+  UL = CorpTypeCd.BC_UNLIMITED, // SPECIAL NAMEREQUEST-ONLY ENTITY TYPE
+  CC = CorpTypeCd.BC_CCC,
+  CP = CorpTypeCd.COOP,
+  FR = CorpTypeCd.NR_SOLE_PROP,
+  GP = CorpTypeCd.PARTNERSHIP
+}
```
severinbeauvais, line 20: This is now a copy of what's in Create UI (NameRequestTypes). Let's remember to move this to shared components repo some time.

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: Or maybe a getter, if the rule for company vs business is clear?

src/store/getters/state-getters.ts
severinbeauvais: I think we'd be better with more verbosity (clarity) instead of optimization here. Honestly it's a gray area and more of a "feel" than anything else.

Or rename this getter `isChangeAlterationFiling` ??? (IMO this is ugly but clear. Prefer you don't do this 😁 )

cameron-freshworks: Oh for sure, will refactor this.
We don't need to centralize this computed value. `isChangeFiling || isAlterationFiling` is fine!

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -199,8 +206,37 @@
...
+      <v-divider class="mx-4" />
+
+      <div class="section-container">
+        <v-row no-gutters>
+          <v-col cols="3">
+            <label><strong>Nature of Business</strong></label>
+          </v-col>
+
+          <v-col cols="9">
+            <div class="info-text">Cookie and cracker manufacturing</div>
```
severinbeauvais, line 232: Mmm, my favourite!

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -405,12 +444,17 @@ export default class YourCompany extends Mixins(
     return 'Unknown'
   }
 
+  /** The start date for a firm. */
+  private get businessStartDate (): string {
+    return 'Unknown'
```
severinbeauvais, line 449: Unless this changes, we know which field this is:
https://github.com/bcgov/business-schemas/blob/6b0e8f0734a656aee10444b1ada2ed16e2368d5d/src/registry_schemas/schemas/registration.json#L38

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -405,12 +444,17 @@ export default class YourCompany extends Mixins(
...
+  }
+
   /** Compare names. */
   private get isNewName () {
     const correctedName = this.getApprovedName
     const currentName = this.isCorrectionFiling
-      ? this.getOriginalIA.incorporationApplication.nameRequest.legalName
-      : this.getBusinessSnapshot.businessInfo.legalName
+      ? this.getOriginalIA?.incorporationApplication?.nameRequest.legalName
+      : this.getBusinessSnapshot?.businessInfo?.legalName
```
severinbeauvais, line 457: Why might getOriginalIA or getBusinessSnapshot be falsey?

The risk with using too much conditional chaining is that it hides cases where something might incorrectly be null/undefined. (Like a startup timing issue.)

severinbeauvais: Also -- and you might be watching this already -- I think the business object and NR object are changing a bit, such that some fields have moved.

For example (for registrations at least), legal type is now in the NR object: https://github.com/bcgov/business-create-ui/blob/7404fcbbccccced5736ad316294daa86b80153a9/src/mixins/filing-template-mixin.ts#L333

I _think_ Vysakh_ championed this. But I'm not sure about anything (sorry).

src/resources/Edit/Alteration/cooperative.ts
```diff
@@ -0,0 +1,19 @@
...
+  displayName: GetCorpFullDescription(CorpTypeCd.COOP),
+  nameRequestType: NameRequestEntityTypes.CP,
+  filingData: {
+    entityType: CorpTypeCd.COOP,
+    filingTypeCode: FilingCodes.ALTERATION,
+    priority: false
+  },
+  certifyClause: 'Note: It is an offence to make a false or misleading statement in respect ' +
+    'of a material fact in a record submitted to the Corporate Registry for filing. ' +
+    'See section 427 of the Business Corporations Act.'
```
severinbeauvais, line 18: Double check the act here. Might be different for coops.

src/views/Alteration.vue
```diff
@@ -19,10 +19,6 @@
         <ShareStructures class="mt-10" />
 
         <Articles class="mt-10" />
-
-        <AgreementType class="mt-10" v-if="false" />
-
-        <Detail class="mt-10" v-if="false" />
```
severinbeauvais, line 25: Where did these go?

src/views/Alteration.vue
```diff
@@ -256,6 +248,16 @@ export default class Alteration extends Mixins(
     return ''
   }
 
+  /** The entity specific resource file for an Alteration filing. */
+  private get alterationResources (): any {
+    const resources = AlterationResources.find(x => x.entityType === this.getEntityType)
+    if (!resources) {
+      // go to catch()
+      throw new Error(`Invalid Alteration resources entity type = ${this.getEntityType}`)
```
severinbeauvais, line 256: This seems to work OK because this getter is only use by code (not by template).

Personally, I'd be tempted to just return the result from `find` and let the caller deal with possible null/undefined response. That would move the error-checking logic entirely into the caller, which would be more clear. Currently some of that logic is hidden here -- the caller would probably not know that this getter could throw an exception. (In some cases we explicitly say that some code might throw an error, just to show that error handling is distributed between caller and callee.)

src/views/Change.vue
```diff
@@ -0,0 +1,294 @@
...
+    }
+    return ''
+  }
+
+  /** The entity specific resource file for a change filing. */
+  private get changeFirmResources (): any {
+    const resources = ChangeFirmResources.find(x => x.entityType === this.getEntityType)
+    if (!resources) {
+      // go to catch()
+      throw new Error(`Invalid Change resources entity type = ${this.getEntityType}`)
```
severinbeauvais, line 181: Same comment as above,

src/views/Change.vue
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+
+  /** Emits Fetch Error event. */
+  @Emit('fetchError')
+  private emitFetchError (message: string = ''): void { }
+
+  /** Emits Have Data event. */
+  @Emit('haveData')
+  private emitHaveData (haveData: Boolean = true): void { }
```
severinbeauvais, line 284: `Boolean` is the class. `boolean` is the type. This should be the latter.

severinbeauvais (approved):
OK, I've reviewed everything.

None of my comments are show-stopping, if you need to merge this ASAP, but would like to see them addressed/closed one way or the other.

cameron-freshworks:
> OK, I've reviewed everything.
> 
> None of my comments are show-stopping, if you need to merge this ASAP, but would like to see them addressed/closed one way or the other.

Thank you! Will make these updates in the morning, absolutely 👍

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: Updated 👍

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: Updated 👍

cameron-freshworks: Actually, change of heart here. I believe it's simpler to leave the reference as a single resource value and apply the variations where used. In this case 'Your`. 
This allows the resource value a little more reusability and flexibility.

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: Same as above, i've left this resource value as the single reference. 'company' or 'business' etc are the 1 commonality between the designs requirements where these values are used, with minor adjustments between implementations. 
`Your Company` and `Your Business` for one example, `Company Name` and `Business Name` for another example.
I think it makes sense to make the common value and resource that alone. 
Going to leave as is 👍

src/store/getters/state-getters.ts
cameron-freshworks: Updated!

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: As per our original design, we would try to have minimal logic in the components to handle filing/entity type variations (so the logic doesn't get complicated). In this case the logic is straight-forward at the moment so I'm ok with this. (Can always be easily changed later.)

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: ok

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: For sure, but not how/where it's retrieved yet. I'm assuming general `business` info will include this in it's response for a sp/gp but this schema references a Registration filing structure not exactly what we are looking at in this instance.

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: For sure, i agree. This was more or less to produce the base of the new flow without throwing a bunch of type errors. 
This needs to be refactor as we work on the individual components and adapt them to the new entity types. 

Will try and preserve the `getBusinessSnapshot` so that it can be either firm or company, so there would both could never be `falsey`.

src/resources/Edit/Alteration/cooperative.ts
cameron-freshworks: haha i was thinking the same. However, we don't have Alterations available for Coops to date, so in the event we implement them this can be updated!

src/views/Alteration.vue
cameron-freshworks: Removed them as they were not being used.

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: Either way, removed the optional chaining so i don't forget later 👍

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: ok for now

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: ok for now

src/resources/Edit/Alteration/cooperative.ts
severinbeauvais: How about dropping a "FUTURE" comment in here so we remember to update it?

src/views/Alteration.vue
severinbeauvais: OK!

Oh, haha, I see the "v-if" now :)

src/views/Alteration.vue
cameron-freshworks: For sure, i see what you mean. 
I'm not sure how to refactor. 
Seems verbose to reference and check the getter before it's used it imo but i'm open to other ideas!

src/views/Change.vue
cameron-freshworks: typo, good catch

cameron-freshworks: Updated!

src/resources/Edit/Alteration/cooperative.ts
cameron-freshworks: Will do, updated!

src/views/Alteration.vue
severinbeauvais: Have the getter just return the data you're looking for (line 253).

Move the check on lines 254-257 to the caller. Then all the checks and try-catch are in the same place.

cameron-freshworks: Updated, the resources actions were already wrapped in a try catch, so added a check throw the error if required. 
Same for `Change.vue`

src/views/Change.vue
cameron-freshworks: Updated!

src/components/common/YourCompany/SummaryDefineCompany.vue
severinbeauvais: I see you fixed one of these, but I think there are more in the code.

severinbeauvais (approved):
I don't need to review this again. When you're happy with it, I'm happy with it.

src/components/common/YourCompany/SummaryDefineCompany.vue
cameron-freshworks: I suspect these are a result of directory renames, as i didn't/wouldn't put `/index` in the imports 😛 
Will have a search.

cameron-freshworks: Definitely, found a few more. Updated 👍

PR #238, opened 2022-02-25
Entity Type Update + Snapshot and Draft Refactors 

cameron-freshworks:
*Issue #:* /bcgov/entity#10620

*Description of changes:*
* Entity snapshot refactors
* Unit test additions/fixes
* Includes Business Type + Update tooltips
* Implemented Placeholder for Change Summary
* Refactors to handle save and resume for Change Draft Filings
* UX Updates for My Business Section
* Implements updated Fee Codes


src/components/Edit/ChangeSummary.vue
```diff
@@ -0,0 +1,155 @@
...
+  // Global getters
+  @Getter getApprovedName!: string
+  @Getter getBusinessNumber!: string
+  @Getter isBusySaving!: boolean
+
+  // Change flag getters
+  @Getter hasBusinessNameChanged!: boolean
+
+  // Global actions
+  @Action setSummaryMode!: ActionBindingIF
```
severinbeauvais, line 81: Depending on how "low level" this component is, it might be better for it to NOT use the store (use props and events instead).

I would say...
1. top level - the main view - should not know too much detail abouts its components but can use the store, talk with App.vue, fetch data
2. middle level - a simple component on the main page/view - doesn't need to know about top level stuff, provides interface between its low-level sub-components and store
2b. middle level - if middle component is complex, it may contain middle level components of its own
3. low level - does the detailed work/logic - ideally should have minimal dependencies, for example should not get/set from/to store; this low-level is an easy candidate for reuse/sharing

src/components/Edit/ChangeSummary.vue
```diff
@@ -0,0 +1,155 @@
...
+    // We don't change views just interchange components, so scroll to top for better UX.
+    await this.scrollToTop(document.getElementById('app'))
+  }
+
+  onDeleteClicked (): void {
+    this.$root.$emit('delete-all')
+  }
+
+  @Emit('haveChanges')
+  emitHaveChanges (): void {}
```
severinbeauvais, line 104: Not used?

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -231,16 +246,15 @@ export default class ChangeBusinessType extends Mixins(CommonMixin, EnumMixin) {
   // Global getters
   @Getter getApprovedName!: string
   @Getter getEntityType!: CorpTypeCd
-  @Getter isConflictingLegalType!: boolean
-  @Getter getBusinessSnapshot!: BusinessSnapshotIF
-
-  // Alteration flag getters
+  @Getter getEntitySnapshot!: EntitySnapshotIF
+  @Getter getResource!: ResourceIF
```
severinbeauvais, line 250: Since this component uses global getter (esp resource) then I'd say this is a middle-level component, even though it's in a sub-sub-folder and uses no lower components. And that's fine; the architecture can be a bit flexible :)

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -178,6 +178,8 @@
       </v-row>
     </div>
 
+    <v-divider v-if="isChangeFiling" class="mx-4 my-1" />
```
severinbeauvais, line 181: Or alteration filing?

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -269,7 +281,7 @@
 
     <!-- Folio Information -->
     <template v-if="isPremiumAccount">
-      <v-divider class="mx-4" />
+      <v-divider class="mx-4 my-1" />
```
severinbeauvais, line 284: Wow, almost 300 lines of template. In another life, we should create sub-components for various sections in here.

src/interfaces/resource-interfaces/resource-interface.ts
```diff
@@ -13,5 +13,6 @@ export interface ResourceIF {
     filingTypeCode: FilingCodes,
     priority?: boolean
   },
+  typeChangeInfo?: string
```
severinbeauvais, line 16: Would it make sense to create a "change" object to nest this under? Maybe in the future when there is more than 1 unique change property?

src/mixins/filing-template-mixin.ts
```diff
@@ -306,23 +304,46 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
-  private prepareNameTranslations () : NameTranslationIF[] {
-    // Filter out and modify name translation to match schema
-    return this.getNameTranslations?.filter(x => x.action !== ActionTypes.REMOVED)
-      .map(x => {
-        let nameTranslation = {
-          name: x.name
-        }
-        if (x.action !== ActionTypes.ADDED) {
-          nameTranslation['id'] = x.id
+  buildChangeFiling (isDraft: boolean): ChangeFirmIF {
```
severinbeauvais, line 311: You can get rid of these Codecov comments by writing a unit test that executes this code. Example:
https://github.com/bcgov/business-create-ui/blob/9c73415ce3472859096ed199e28dab84d14a33b6/tests/unit/filing-template-mixin.spec.ts#L32

severinbeauvais (approved):
Really good looking code. Nice refactors/generalizations. No concerns.

lekshmimallika-aot (approved):

src/components/Edit/ChangeSummary.vue
cameron-freshworks: Updated! 

Was a left over from the template i used for this placeholder component. 
I'm sure it will be used in the near future, but definitely not used at the moment.

src/components/common/YourCompany/ChangeBusinessType.vue
cameron-freshworks: Good point, this is worth a closer look tho!

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: Oddly enough this divider is exclusive to Change Filings. 
I asked directly about this because i missed it in the initial design and was part of UX feedback. 
But yea, unique design for this filing type 😝

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: It has certainly grown! 
Will look a little closer at improving this though, if not now will certainly keep it in mind as we proceed.

src/interfaces/resource-interfaces/resource-interface.ts
cameron-freshworks: Updated!

src/components/Edit/ChangeSummary.vue
cameron-freshworks: While i do agree this component in particular is a reflection of store data for the most part, so i think if perhaps any layer to is to have access to store, it makes sense here.
Either we bind the getters here or the alternative is bind them on the parent, pass them into the component and create the prop bindings etc. 
The latter just feels like more code, with more potential points of failure. I'm all for it, just hard to see the value in this case in particular.

src/mixins/filing-template-mixin.ts
cameron-freshworks: This was really cool, implemented the start of this in Edit UI, thank you!

cameron-freshworks:
Codecov is consistently reported on untouched code.. so i'm not sure 🤷

src/components/Edit/ChangeSummary.vue
severinbeauvais: Since this component's parent is a view, I agree that it's fine that THIS component handle all the store stuff.

src/mixins/filing-template-mixin.ts
severinbeauvais: Looks good 👍

PR #241, opened 2022-03-02
Contact Info Refactors and Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#10687

*Description of changes:*
* Implements updated design/ux locally and from Shared Component
* Refactors and cleans up handling of contact info
* Updated unit testing


src/mixins/auth-api-mixin.ts
```diff
@@ -55,7 +55,11 @@ export default class AuthApiMixin extends Vue {
       .then(response => {
         if (response?.data) {
           return {
-            contacts: response.data.contacts,
+            contact: {
+              email: response.data.contacts[0].email,
+              phone: response.data.contacts[0].phone,
+              extension: response.data.contacts[0].phoneExtension
+            },
```
cameron-freshworks, line 62: Where we used this data, it was de-structured down to the first contact item in the array.
This serves two purposes
- to clean up that de-structuring and only do it in one place
- to map the `phoneExtension` property to `extension` property. 

Lear (and locally here) we referred to `extension` property but there is a disconnect from auths `phoneExtension` property name.
I'm not sure when the values name in auth changed or if ever had, but the display of the extension value was broken due to this.

src/mixins/common-mixin.ts
```diff
@@ -93,20 +93,27 @@ export default class CommonMixin extends Vue {
...
     if (this.isCorrectionFiling) return 'Corrected'
     if (this.isAlterationFiling || this.isChangeFiling) return 'Changed'
     return 'Edited'
   }
 
+  /** The appropriate edits saved label for corrections, alterations or change filings. */
+  get editSavedLabel (): string {
+    if (this.isCorrectionFiling) return 'Corrections Saved'
+    if (this.isAlterationFiling || this.isChangeFiling) return 'Changes Saved'
+    return 'Edits Saved'
```
cameron-freshworks, line 114: A new display value for changes that take effect immediately i think

src/mixins/filing-template-mixin.ts
```diff
@@ -493,6 +493,11 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
         }
       }) || []
     )
+    // Store business contact info
+    this.setBusinessContact({
+      ...entitySnapshot.authInfo.contact
+    })
+
```
cameron-freshworks, line 500: As with alterations and change filings, this information will always be up to date. 
Regardless if its a new change filing or from a draft change filing, once the changes are saved, they take effect immediately.

src/components/common/YourCompany/BusinessContactInfo.vue
```diff
@@ -5,8 +5,9 @@
     :originalBusinessContact="originalContact"
     :hasBusinessContactInfoChange="hasBusinessContactInfoChange"
     :editLabel="editLabel"
-    :editedLabel="editedLabel"
+    :editedLabel="editSavedLabel"
```
severinbeauvais, line 8: Where does the new value come from?

severinbeauvais: Oh, Common Mixin. OK.

src/mixins/common-mixin.ts
severinbeauvais: It's not obvious when getters come from a mixin like this. Would it make sense to move these to the store getters file (or have a new getter file)? We already do this with some getters like isRoleStaff. Just thinking out loud. If this bothers noone else then nvm.

src/mixins/filing-template-mixin.ts
```diff
@@ -493,6 +493,11 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
         }
       }) || []
     )
+    // Store business contact info
+    this.setBusinessContact({
+      ...entitySnapshot.authInfo.contact
```
severinbeauvais, line 498: Why destructure into another object? To create a copy? Probably not needed as I assume the store creates its own copy.

(I see this is also done on line 298. Probably not needed there either.)

(There are definitely times when we want to create a copy/clone of an object, otherwise we can end up pointing to the same object as someone else, which can cause some difficult bugs to find.)

severinbeauvais: Also lines 592-594 and 669-671 ?

lekshmimallika-aot (approved):

src/store/state/state-model.ts
```diff
@@ -177,7 +177,7 @@ export const stateModel: StateModelIF = {
       contactPoint: {
         email: '',
         phone: '',
-        extension: ','
```
severinbeauvais, line 180: ???

Um, yeah.

severinbeauvais (approved):
Nice little refactors 👍 

I also like the new tests that verify the contacts -> contact change.

src/mixins/filing-template-mixin.ts
cameron-freshworks: Good point! 
I think this was spread previously because we we applying `confirmEmail` at the same time. Since this is happening at the component level and we don't care about the confirmEmail beyond components validations we could probably just set it! 
Will try this, thank you

src/mixins/common-mixin.ts
cameron-freshworks: My thoughts were the same as yours initially, it does feel a bit like a phantom computed value, especially if the component is extending a few mixins. 
That being said, it does cut down on the property decorators in the component itself and this is also the case for several of the computed labels/tags etc in the common mixin. 

So i guess i'm impartial, i'm fine with this because it's less code in our components and finding the computed values themselves are just a  Command (⌘) + click away. 🤷

src/store/state/state-model.ts
cameron-freshworks: haha yea, i had to look into this, i thought perhaps it was intentional at first.

cameron-freshworks:
> Nice little refactors 👍
> 
> I also like the new tests that verify the contacts -> contact change.

Thank you!

cameron-freshworks:
There was just some requirements updated in the ticket, so will be making another commit! 
Might involve another change to the shared component.. stay tuned.

src/mixins/common-mixin.ts
severinbeauvais: It doesn't bug me enough right now to change it. Maybe later 😁

src/mixins/filing-template-mixin.ts
cameron-freshworks: Updated!

cameron-freshworks:
Updated with PR comments and shared component version number which include added requirements

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "0.7.1",
+  "version": "0.7.2",
```
severinbeauvais, line 3: In one of your PRs, I recommend updating this to 2.x.x (maybe 2.7.2 to retain some connection with past work?).

Why v2? Because v1 has already been live for a while, and you're adding a new filing type.

Open to other ideas.

cameron-freshworks: I bumped it up to `0.7.x` with this new filing type, but i agree. 
Perhaps when we release Change Firm filings, we can then set the version to `2.0.0`.

src/assets/styles/base.scss
```diff
@@ -267,6 +267,19 @@ ul.list {
...
+.header-icon {
+  size: 25px !important;
+  color: $app-dk-blue !important;
+  vertical-align: top !important;
+}
+
+.info-icon {
+  size: 22px !important;
+  color: $app-blue !important;
+  vertical-align: top !important;
```
severinbeauvais, line 280: Is "important" needed for all properties? I think it's better to reserve the use of "important" to components that need to override it -- so base properties should minimize the use of "important".
(It is still possible to override these properties if needed in the future by being more specific with the CSS selector, or perhaps by "latest match wins" logic.)

severinbeauvais: FWIW, I have not seen any guidance or best practises on CSS in base classes vs components, so my opinion is based on my experience only.

src/components/Edit/ChangeSummary.vue
```diff
@@ -4,33 +4,9 @@
...
         </v-col>
-
-        <!-- Actions -->
-        <v-col cols="3" class="mt-n2">
-          <div class="actions mr-4">
-            <v-btn
-              text color="primary"
-              id="btn-change"
-              :disabled="isBusySaving"
-              @click="onChangeClicked()"
```
severinbeauvais, line 18: Can you now delete onChangeClicked() and onDeleteClicked() ?

src/components/common/YourCompany/BusinessContactInfo.vue
```diff
@@ -5,9 +5,11 @@
...
     :disableActions="isCorrectionFiling"
+    :disableActionTooltip="isChangeFiling"
     :invalidSection="invalidSection"
+    :optionalPhone="true"
```
severinbeauvais, line 12: Did you try without the `:` ? Theoretically it should work since this is not a dynamic binding.

src/assets/styles/base.scss
```diff
@@ -267,6 +267,19 @@ ul.list {
   color: $primary-blue;
 }
 
+// General icon styling
```
severinbeauvais, line 270: Nice refactor / cleanup!

src/components/Edit/ChangeSummary.vue
cameron-freshworks: Indeed!

This is what happens when i get cater to ongoing UX feedback, get caught in a spiral of small updates outside the scope of my tickets 😅 

Will remove!

package.json
severinbeauvais: When to change a version number is a good question...

In my opinion/experience, we should change it when we start a new feature. Let's say we go with 2.0.0 here. As the feature progresses, we will go to 2.0.1 or maybe 2.1.0. Eventually the production-ready version of the feature might be, say, 2.1.10. Then we repeat for future features. But what if we need to back-patch a bug (using a bugfix branch)? Then it makes sense that the fixed version for that feature is 2.1.11. This allows the "final version" of a feature to always be 2.something.

IMO, it makes less sense to say "2.0.0 is the production-ready version for this feature". First, what about changes and fixes? Second, the next feature will be in 2.x.x instead of its own version.

Again, my opinion, open to other ideas.

src/assets/styles/base.scss
severinbeauvais: Leave it as is. I trust your experience and intuition here.

src/components/common/YourCompany/BusinessContactInfo.vue
cameron-freshworks: I did! 
It works without the binding but i'm not sure the reason to write it differently than the other props.

package.json
cameron-freshworks: Ok, i'm pretty impartial.
Updated to 2.0.0!

src/components/common/YourCompany/BusinessContactInfo.vue
cameron-freshworks: Sorry, it doesn't work how i would expect it too. Writing `optionalPhone` does not work, however assigning it a value does work `optionalPhone="true"`. 
That being said, you could declare it as `optionalPhone="fooBar"` and it will still work 😆

severinbeauvais: I'm slightly leaning towards no colon just to show this isn't a bound prop, but your point is fair and I'm OK with the current code.

(Maybe -- though it may be optimized out by the compiler or Vue engine -- just maybe, this saves some reactive code from being attached to this prop, which isn't needed since this prop will never change. Back in the stone age, this was a valid manual optimization to save CPU cycles. Today this matters little. FWIW)

cameron-freshworks: For sure, thats a fair take. I do appreciate this insight!

cameron-freshworks: Updated this. 
Initially made it true because it's not editable in a correction but updated to be more explicit when we allow optional phone.

severinbeauvais: > Sorry, it doesn't work how i would expect it too. Writing `optionalPhone` does not work, however assigning it a value does work `optionalPhone="true"`. That being said, you could declare it as `optionalPhone="fooBar"` and it will still work 😆

Yeah, I would expect just this prop's presence (no value) to work, as with other Vue props. I haven't figured that out yet.

"fooBar" probably works because it's truthy.

src/components/Edit/ChangeSummary.vue
severinbeauvais: And also onDeleteClicked() ?

cameron-freshworks: Was a weird one, my editor said it was still in use!
removed (until i add i back in..)

src/components/Edit/ChangeSummary.vue
```diff
@@ -64,10 +64,6 @@ export default class ChangeSummary extends Mixins(
 
     return `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.`
   }
-
-  onDeleteClicked (): void {
-    this.$root.$emit('delete-all')
```
severinbeauvais, line 100: Is this event still used by another component?

cameron-freshworks: Not in this component, no. 
But Alteration Summary has the same event which is called in some unit testing for that Summary component. Perhaps that was caused it, since it wasn't a private method.

PR #242, opened 2022-03-04
Contact Info Ux

cameron-freshworks:
*Issue #:* /bcgov/entity#10687

*Description of changes:*
* Implements latest version of contact info
* Layout updates to meet latest requirements for general layout

### DIFF
--- a/src/assets/styles/base.scss
+++ b/src/assets/styles/base.scss
@@ -269,13 +269,13 @@ ul.list {
 
 // General icon styling
 .header-icon {
-  size: 25px !important;
+  font-size: 25px !important;
   color: $app-dk-blue !important;
   vertical-align: top !important;
 }
 
 .info-icon {
-  size: 22px !important;
+  font-size: 22px !important;
   color: $app-blue !important;
   vertical-align: top !important;
 }
--- a/src/assets/styles/base.scss
+++ b/src/assets/styles/base.scss
@@ -269,13 +269,13 @@ ul.list {
 
 // General icon styling
 .header-icon {
-  font-size: 25px !important;
+  font-size: $px-25 !important;
   color: $app-dk-blue !important;
   vertical-align: top !important;
 }
 
 .info-icon {
-  font-size: 22px !important;
+  font-size: $px-22 !important;
   color: $app-blue !important;
   vertical-align: top !important;
 }
--- a/src/assets/styles/layout.scss
+++ b/src/assets/styles/layout.scss
@@ -53,7 +53,7 @@
 }
 
 .section-container {
-  padding: 1rem 1rem;
+  padding: 1.25rem 1.25rem;
 }
 
 .sub-section {
--- a/src/assets/styles/theme.scss
+++ b/src/assets/styles/theme.scss
@@ -61,4 +61,16 @@ $BCgovFontColorInverted: #ffffff;
 $app-blue: #1669bb; // same as the Vuetify theme primary
 $app-dk-blue: #38598a; // same as the Vuetify theme appDkBlue
 $app-red: #d3272c; // same as the Vuetify theme error
-$app-green: #1a9031, // same as the Vuetify theme success
+$app-green: #1a9031; // same as the Vuetify theme success
+
+// Sizes for Fonts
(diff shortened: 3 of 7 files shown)


src/components/common/YourCompany/YourCompany.vue
```diff
@@ -1,12 +1,12 @@
 <template>
-  <v-card flat id="your-company">
+  <v-card flat id="your-company" class="pb-2">
```
cameron-freshworks, line 2: Requirements call for more padding at the bottom of this card.

src/assets/styles/layout.scss
```diff
@@ -53,7 +53,7 @@
 }
 
 .section-container {
-  padding: 1rem 1rem;
+  padding: 1.25rem 1.25rem;
```
cameron-freshworks, line 56: Requirements call for more padding on these subsections. 
Verified in Alterations, Change and Correction filings

lekshmimallika-aot (approved):

src/assets/styles/base.scss
```diff
@@ -269,13 +269,13 @@ ul.list {
 
 // General icon styling
 .header-icon {
-  size: 25px !important;
+  font-size: 25px !important;
```
severinbeauvais, line 272: Do we have font size variables that can be used?

severinbeauvais (approved):

src/assets/styles/base.scss
cameron-freshworks: Will look 👀

cameron-freshworks: We don't but i'll add them to the theme, apply them here :)

cameron-freshworks: Updated, lemme know if that is what you meant

src/assets/styles/theme.scss
```diff
@@ -61,4 +61,16 @@ $BCgovFontColorInverted: #ffffff;
...
+$px-11: 0.6875rem;
+$px-12: 0.7500rem;
+$px-13: 0.8125rem;
+$px-14: 0.8750rem;
+$px-15: 0.9375rem;
+$px-16: 1.0000rem;
+
+// Sizes for Icons
+$px-22: 1.375rem;
+$px-25: 1.5625rem;
```
cameron-freshworks, line 76: Called these out explicitly as icons although we could add to the above list. 
Just can't recall the last time we have font this large outside of headers.

Good with either or

severinbeauvais: I'd code the icons with their directly (ie, in rem) and reserve these variables for fonts only.

src/assets/styles/base.scss
severinbeauvais: Perfect.

I know it might seem silly to name a variable after its value like this, but if we use those only for font sizes then it's easy to do a global search to see where they're used (compared with hard values). And it will be easy to perhaps prune our font sizes down to 4 instead of 6 (or more in other UIs), when UI team goes through and asks for cleanup.

severinbeauvais: Aside: Here's an example of something we should clean up some time:

https://github.com/bcgov/business-create-ui/blob/main/src/assets/styles/theme.scss#L67

cameron-freshworks:
Need to do something with this CodeCov report. 
Consistently picks up all the code. Is this a configuration of sorts?

severinbeauvais:
> Need to do something with this CodeCov report. Consistently picks up all the code. Is this a configuration of sorts?

This seems to be the configuration for it, but I don't know how to tune it. Maybe Patrick does?

https://github.com/bcgov/business-edit-ui/blob/main/.github/workflows/ci.yml#L42

cameron-freshworks:
> > Need to do something with this CodeCov report. Consistently picks up all the code. Is this a configuration of sorts?
> 
> This seems to be the configuration for it, but I don't know how to tune it. Maybe Patrick does?
> 
> https://github.com/bcgov/business-edit-ui/blob/main/.github/workflows/ci.yml#L42

Thanks! Yea, i'll touch base with Patrick, perhaps there is something we can update here.

PR #244, opened 2022-03-08
Edit UI - Address Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#10686

*Description of changes:*
* Implements Updated Office Address - refactors for reusability between corrections and change filings. ReadOnly for Alterations
* Update OfficeAddress to supported vue syntax  
* Additional OfficeAddress editing template for Change filings
* Globalized Address and address change getters
* Store refactors
* Unit test updates
* Minor Updates to Document Delivery for placeholder


lekshmimallika-aot (approved):
Looks good

vysakh-menon-aot (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "2.0.1",
+  "version": "2.0.2",
```
severinbeauvais, line 3: I recommend you update sbc-common-components to v42 while you're here. (The change is to retain the accountid in the URL, which was previously automatically removed by SbcHeader.)

src/components/Edit/ChangeSummary.vue
```diff
@@ -26,17 +26,26 @@
         </v-row>
       </div>
     </template>
+
+    <template>
```
severinbeauvais, line 30: You don't need the template. Above it was used a nice way to wrap both the divider and div in a v-if.

I recommend a block comment as well (useful when you use code folding to see the big picture).

package.json
cameron-freshworks: Good call, thanks for the heads up.

src/components/Edit/DocumentsDelivery.vue
```diff
@@ -1,12 +1,12 @@
...
+    <h2>{{sectionNumber}} Documents Delivery</h2>
+    <div class="document-info pt-4 pb-4">Copies of the {{ getFilingName.toLowerCase() }} documents will be sent
       to the following email address listed below.</div>
     <div :class="{ 'invalid-section': documentDeliveryInvalid }">
-      <v-card flat class="pt-4 pr-8">
+      <v-card flat class="section-container">
         <v-container>
-          <v-row class="pl-4">
+          <v-row>
             <v-col cols="3" class="px-0">
```
severinbeauvais, line 10: You might want to add `cols="12"` and `md="3"` to the cols to promote wrapping below size "md".

src/components/Edit/ChangeSummary.vue
cameron-freshworks: Sorry, missed my conditional here on the template. 
Will include a divider as well. 
Update inc

src/components/Edit/DocumentsDelivery.vue
cameron-freshworks: If we are aiming for responsive we will need to update an awful lot. 
Were these additional requirements in Registrations?

severinbeauvais: No, they were not requirements.

I updated various rows/cols in Create UI because I noticed they were not acting similar on the review page (and it's good practise 😁 ).

You can ignore this conversation.

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -35,76 +37,78 @@
           />
           <div v-else-if="isEmpty(deliveryAddress)" class="info-text">(Not entered)</div>
           <div v-else class="info-text">Same as Mailing Address</div>
-        </v-flex>
+        </v-col>
+
+        <template v-if="!isSummaryView">
```
severinbeauvais, line 42: I see a flag `isEditing` and this flag `isSummaryView`. Are they mutually exclusive? Do they interact (what are the possible values)?

It might be easier to understand with a sort of state machine like I did for Nature of Business. Then you know exactly which condition (state) the code applies to.

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -35,76 +37,78 @@
...
-        </v-flex>
-      </v-layout>
+            </div>
+          </v-col>
+        </template>
+      </v-row>
 
-      <v-layout row id="summary-records-address" class="mt-4 mx-0" v-if="entityFilter(CorpTypeCd.BENEFIT_COMPANY)">
-        <v-flex xs3>
+      <v-row v-if="entityFilter(CorpTypeCd.BENEFIT_COMPANY)" id="summary-records-address" class="mt-4 mx-0" no-gutters>
```
severinbeauvais, line 102: I replaced some `entityFilter(CorpTypeCd.BENEFIT_COMPANY)` with `isBComp` for cleaner code. What do you think here?

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -783,4 +850,12 @@ ul {
   color: $gray7;
   font-weight: normal;
 }
+
+::v-deep {
```
severinbeauvais, line 854: Please add a comment saying what this is for (since it's once-removed from whatever it's styling).

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
```diff
@@ -1,21 +1,22 @@
...
 /** Entity snapshot object. */
 export interface EntitySnapshotIF {
   businessInfo: BusinessInformationIF
   authInfo: AuthInformationIF
   orgPersons: GetOrgPersonIF[]
+  addresses: AddressesIF
 
   // Properties exclusive to Companies
-  incorporationAddress?: IncorporationAddressIf
+  // incorporationAddress?: AddressesIF
```
severinbeauvais, line 14: Obsolete or for future use? Also line 21.

src/interfaces/filing-interfaces/index.ts
```diff
@@ -5,3 +5,6 @@ export * from './filingData-interface'
 export * from './filingHeader-interface'
 export * from './incorporationApplication-interface'
 export * from './incorporationFiling-interface'
+
+// Proxy export for Shared Components which rely on IncorporationAddressIf
+export { AddressesIF as IncorporationAddressIf } from '@/interfaces/stepper-interfaces/YourCompany/address-interface'
```
severinbeauvais, line 10: Sad but OK. (We _should_ go and update the shared components, but it's a breaking change.)

This is more of an alias than a proxy :)

src/interfaces/stepper-interfaces/YourCompany/address-interface.ts
```diff
@@ -18,8 +18,8 @@ export interface BaseAddressObjIF {
   deliveryAddress?: AddressIF
 }
 
-/** Interface to define the incorporation addresses. */
-export interface IncorporationAddressIf {
+/** Interface to define the addresses. */
+export interface AddressesIF {
```
severinbeauvais, line 22: This name is probably OK for now but this is not a univeral address interface. For example, [registration business addresses](https://github.com/bcgov/business-create-ui/blob/main/src/interfaces/Registration/business-address-interface.ts) don't look like this.

If a better name comes to you, consider changing it 😁

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: They are their own states. They don't interact directly no. They are booleans. 
One is an indicator if the component is in its form state, the other to indicate a summary state. 

I'm not sure i want to add complexity, as they are both rather simple flags. I can have a look at what you are referring too though!

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: Definitely! Will update

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
cameron-freshworks: Indeed, will clean this up!

src/store/state/state-model.ts
```diff
@@ -92,7 +92,7 @@ export const stateModel: StateModelIF = {
   documentDelivery: {
     documentOptionalEmail: ''
   },
-  officeAddresses: {},
+  officeAddresses: null,
```
severinbeauvais, line 95: The old value supported de-referencing but null doesn't. As long as this is always initialized somehow, this is fine (and preferable). 👍

src/mixins/filing-template-mixin.ts
```diff
@@ -340,6 +338,14 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
       }
     }
 
+    // Apply business address changes to filing
+    if (this.officeAddressesChanged) {
+      filing.changeOfRegistration.businessAddress = {
+        mailingAddress: this.getOfficeAddresses.registeredOffice.mailingAddress,
+        deliveryAddress: this.getOfficeAddresses.registeredOffice.deliveryAddress
+      }
```
severinbeauvais, line 346: Oh, you're using AddressesIF to store the business address into the registeredOffice object?

src/views/Change.vue
```diff
@@ -35,6 +35,13 @@
           :validate="getAppValidate"
         />
 
+        <DocumentsDelivery
+          class="mt-10"
+          sectionNumber="1."
```
severinbeauvais, line 40: Can we use the CSS counter to automate section numbers?

src/interfaces/filing-interfaces/index.ts
cameron-freshworks: hahah i'm not sure! Alias sounds deceitful

tests/unit/OfficeAddresses.spec.ts
```diff
@@ -1111,17 +1055,14 @@ describe('actions and events', () => {
...
+    expect(moreBtn.find('span').exists()).toBe(true)
+    moreBtn.trigger('click')
+    await Vue.nextTick()
+
+    const correctBtn = wrapper.find('#btn-more-actions-edit')
+    correctBtn.trigger('click')
+    await Vue.nextTick()
 
     // make a change (and verify)
     await wrapper.setData({ mailingAddress: getAddressX(6) })
```
severinbeauvais, line 1068: Nice functional tests 👍 

(More than just checking static component elements.)

src/interfaces/stepper-interfaces/YourCompany/address-interface.ts
cameron-freshworks: Alas, naming is hard. 
It does sound pretty universal but within the context of this app we have defined it at least. 
I would have preferred another format (ie a new interface to match an updated addresses responses from the api) but it sounds like there will be no changes on that end.

src/mixins/filing-template-mixin.ts
cameron-freshworks: You bet, this was the api restrictions i referred to above. 
The api is/will be returning these addresses wrapped regardless of EntityType at the moment. 
As our Firm schema refers to these addresses wrapped as businessAddress we need to handle that.

src/views/Change.vue
cameron-freshworks: I looked at this, i'm not sure if it can be established on the parent and inherited into the sub components but can give it a try.

src/components/common/YourCompany/OfficeAddresses.vue
severinbeauvais: Thanks for explanation. I didn't realize you could have a summary view while editing.

If you come up with a way to comment or self-document this to reduce future confusion, great. Otherwise, OK for now I guess.

src/interfaces/filing-interfaces/index.ts
severinbeauvais: This is how I understand it:
Proxy: something on behalf of something else.
Alias: another name for something (*).

(*) Only deceitful in legal circles 😆

package.json
cameron-freshworks: Updated

src/components/Edit/ChangeSummary.vue
cameron-freshworks: Updated

src/interfaces/alteration-interfaces/business-snapshot-interface.ts
cameron-freshworks: Updated

src/interfaces/stepper-interfaces/YourCompany/address-interface.ts
severinbeauvais: Yes, naming is hard. (especially when you're old like me)

Actually, I think the API data (schema) is different for business addresses and other addresses. (Registered/Records office vs none)

nvm for now

src/interfaces/filing-interfaces/index.ts
cameron-freshworks: hahaha enjoying this comment!

ok, i had to post these here. Out of context and english has a thousand meanings for every word.
Updated to alias tho :) 

Screen Shot 2022-03-08 at 1 59 05 PM
Screen Shot 2022-03-08 at 1 58 48 PM

src/mixins/filing-template-mixin.ts
severinbeauvais: Got it.

I went the other way and created [registration business addresses](https://github.com/bcgov/business-create-ui/blob/main/src/interfaces/store-interfaces/state-interfaces/registration-state-interface.ts#L29).

Closing this conversation...

src/interfaces/filing-interfaces/index.ts
cameron-freshworks: Updated

src/views/Change.vue
severinbeauvais: Great. Don't spend a ton of time on it: it was like this when you got here 😁

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: Comment added. To override the default font styling of the base address component, form input fields. 
Updated

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: ok for sure, will have a look. I think a comment where defined would help then!

cameron-freshworks: Updated prop comment, although the comment on `isEditing` seems pretty good.
To 'summarize' (no pun intended) `isSummaryView` is a 3rd state, where the component is in its static form but WITHOUT its actions to initialize the editing mode. 

This is an alternative to an additional summary component to display the information obviously.

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: updated

src/components/common/YourCompany/OfficeAddresses.vue
severinbeauvais: So UI/UX _doesn't_ want the delivery instructions to be italicized here?

(Are you using the latest BaseAddress component? I recently added some classes to it.)

cameron-freshworks: That styling still exists here, currently on the latest cc version. Will rm -rf nodeMods and check again. 
There is local styling here to italicize, could be that it is no longer required. Will 👁️

cameron-freshworks: i'm not picking up the italics font unless applied manually, locally. Even without any of the overrides in play. Is the component supposed to be applying italic for the delivery instructions?

severinbeauvais: Oops, I forgot that I changed that externally (not in BaseComponent)...

https://github.com/bcgov/business-create-ui/blob/main/src/components/Registration/BusinessAddresses.vue#L370

^^ If UI/UX asks you to do the same.

cameron-freshworks: perfect! Updated
fixed what was here.

PR #249, opened 2022-03-16
Address and App Validation Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#10686

*Description of changes:*
* Updated functionality/rules/schema for Addresses
* Updated Addresses UX
* Updated App/Component Validations
* fixed unit tests


vysakh-menon-aot (approved):

src/components/Edit/NatureOfBusiness.vue
```diff
@@ -25,32 +27,49 @@ import { NatureOfBusiness as NatureOfBusinessShared } from '@bcrs-shared-compone
...
 
 ::v-deep {
+  .invalidSection .col-sm-3:first-child label {
+    color: $app-red !important;
+  }
+
+  .v-input .v-label {
+    font-weight: normal;
+  }
+
```
severinbeauvais, line 72: Would you mind making these changes to the shared component so we can all benefit from them? (we=me, haha)

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -857,5 +896,9 @@ ul {
     font-size: $px-16;
     font-weight: normal;
   }
+
+  .v-input input {
+    color: $gray9
+  }
```
severinbeauvais, line 902: All these styles should be right in the BaseAddress component.

Please create a new ticket if you don't have time in your current ticket to update this.

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -175,12 +178,13 @@
...
             hide-details
             v-model="inheritMailingAddress"
+            :disabled="disableSameDeliveryAddress"
             @change="setDeliveryAddressToMailingAddress()"
           />
         </v-col>
       </v-row>
 
-      <template v-if="!inheritMailingAddress">
+      <template v-if="!inheritMailingAddress || disableSameDeliveryAddress">
```
severinbeauvais, line 187: Won't this affect other change filings?

I mean, is the rule behind this the same for all change filings (ie, BC + CA) ?

src/interfaces/utils-interfaces/address-schema-interface.ts
```diff
@@ -0,0 +1,9 @@
+export interface AddressSchemaIF {
+  streetAddress: {}
+  streetAddressAdditional: {}
+  addressCity: {}
+  addressCountry: {}
+  addressRegion: {}
+  postalCode: {}
+  deliveryInstructions: {}
+}
```
severinbeauvais, line 9: In Create UI, this file is in a new folder (schemas).

Will you add validation rules here, as in the file below?

https://github.com/bcgov/business-create-ui/blob/main/src/schemas/office-address-schema.ts

src/schemas/office-address-schema.ts
```diff
@@ -32,3 +32,5 @@ export const OfficeAddressSchema = {
     maxLength: maxLength(80)
   }
 }
+
+export const RegistrationDeliveryAddressSchema = OfficeAddressSchema
```
severinbeauvais, line 36: Would it make sense to move this alias closer to where it's used (as in the component itself)?

If `RegistrationDeliveryAddressSchema` is used in multiple places then, imho, leave this here.

(Ditto for alias in person-address-schema.ts.)

src/components/Edit/NatureOfBusiness.vue
```diff
@@ -1,11 +1,13 @@
 <template>
-  <div id="nature-of-business-info">
+  <div id="nature-of-business-shared">
```
severinbeauvais, line 2: This is actually not the shared component. I've been trying to have the ID the same as the filename.

I realize "nature-of-business" ID is already taken by the shared component. I don't have a suggestion on what to call this. I avoided this altogether by not wrapping this in a div, hence no ID 😁 

https://github.com/bcgov/business-create-ui/blob/main/src/components/Registration/NatureOfBusiness.vue

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -238,8 +238,10 @@
       <v-divider class="mx-4 my-1" />
 
       <!-- Nature of Business -->
-      <div class="section-container">
-        <NatureOfBusiness />
+      <div id='nature-of-business' class="section-container" :class="{'invalid-section': invalidNatureOfBusiness}">
+        <NatureOfBusiness
+          :invalidSection="invalidNatureOfBusiness"
+        />
```
severinbeauvais, line 244: This can be simplified, eg:

https://github.com/bcgov/business-create-ui/blob/main/src/views/Registration/RegistrationDefineBusiness.vue#L18

Also please use double quotes in the template.

severinbeauvais: 1. Ditto for office addresses below.

2. You can specify `class="section-container"` on the component as well.

Up to you on whatever looks cleaner.

severinbeauvais (commented):
Looks really really good!

src/components/Edit/NatureOfBusiness.vue
cameron-freshworks: You bet, will do!

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: Created!

https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11547

src/components/Edit/NatureOfBusiness.vue
cameron-freshworks: Created!

cameron-freshworks: https://github.com/bcgov/bcrs-shared-components/pull/119

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: Commented on all these but git didn't seem to apply them 😩 

Alterations are unable to edit addresses (this code block)
Corrections uses another sub template from below, since the design layout there is quite different.

src/interfaces/utils-interfaces/address-schema-interface.ts
cameron-freshworks: There is a schema folder in this UI as well which currently houses these :)

src/schemas/office-address-schema.ts
cameron-freshworks: Updated
Removed the alias in the schemas themselves and just applied them at the component 👍

src/components/Edit/NatureOfBusiness.vue
cameron-freshworks: You are right! 

Just removed the wrapper, it wasn't required.
We use the id on the view level for our scroll-to validation.

cameron-freshworks:
Heads up, it appears the git actions api is still having some difficulties.
A snapshot of the unit tests from local.

Screen Shot 2022-03-17 at 10 11 34 AM

src/components/Edit/NatureOfBusiness.vue
cameron-freshworks: I'm not sure what happens to the class binding on the component itself but it wouldn't recognize it without a root node.
I can look into it some more, but my conditional selector wouldn't work without.

cameron-freshworks: Will 👀

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -414,12 +431,15 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
...
   @Action setEditingOfficeAddresses!: ActionBindingIF
+  @Action setValidComponent!: ActionBindingIF
 
   // Declarations for template
   readonly isEmpty = isEmpty
   readonly AddressTypes = AddressTypes
   readonly CorpTypeCd = CorpTypeCd
   readonly addressSchema = OfficeAddressSchema
+  readonly RegistrationMailingAddressSchema = PersonAddressSchema
+  readonly RegistrationDeliveryAddressSchema = OfficeAddressSchema
```
severinbeauvais, line 442: I like this aliasing here.

Now I see that _some time_ (future). we should rename PersonAddressSchema and OfficeAddressSchema -- for example, the person one isn't being use for a person so...

Future.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -259,8 +262,9 @@
     <v-divider class="mx-4 my-1" />
 
     <!-- Office addresses -->
-    <div class="section-container">
+    <div id="business-addresses" class="section-container" :class="{'invalid-section': invalidAddressSection}">
       <OfficeAddresses
+        :invalidSection="invalidAddressSection"
```
severinbeauvais, line 267: I think you can combine the div and OfficeAddresses here also.

src/interfaces/stepper-interfaces/YourCompany/address-interface.ts
```diff
@@ -1,5 +1,6 @@
 /** Interface to define a base address. */
 export interface AddressIF {
+  id?: number
```
severinbeauvais, line 3: I am aware of multiple definitions of AddressIF (eg, also in Create UI, shared components, and probably Filings UI).

It would be _great_ to have a single definition. Thoughts?

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: Looks good, yes. I think the ID in the page is probably overwriting the ID in the shared component. Probably still OK though.

src/components/common/YourCompany/OfficeAddresses.vue
cameron-freshworks: Agreed!

severinbeauvais: Haha, I think I made that comment once already.

It's a double-edged sword. It can be used to self-document what you're doing (like here). So I like this. But it seems weird to use a "person" schema" for a "registration mailing address". Ideally a reviewer (or future developer, or me after I forget) would be able to guess what the schema contains based on its name.

Still future 😁

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: OK, nvm.

cameron-freshworks: Looking into this though still, it seems like the same issue with the Nature of Business wrapper. 
Somehow rearranges the selectors, will try to find out how!

severinbeauvais: I think what Vue is doing is copying the shared component into the `<NatureOfBusiness>` element. So properties are merged.

cameron-freshworks: Interesting!

src/components/Edit/NatureOfBusiness.vue
cameron-freshworks: Ok, like the wrapper below. The id's were in conflict with one overriding the other. 
Removed the override ID's (which were just used in validations) and relied on the component template Id's. 
Removed the wrappers, looks good!

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: You were right! Conflict of Id's, removed the parent wrapper Id's and applied the component ID for the validations

src/interfaces/stepper-interfaces/YourCompany/address-interface.ts
cameron-freshworks: Agreed!
Future: create shared AddressIF's?

severinbeauvais: There is one already!

https://github.com/bcgov/bcrs-shared-components/blob/main/src/interfaces/addresses-interface.ts

Every project has their own 🙄

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -609,4 +622,10 @@ export default class YourCompany extends Mixins(
     min-width: 0.5rem;
   }
 }
+
+::v-deep {
+  .invalid-section .col-sm-3:first-child label {
+    color: $app-red !important;
+  }
```
severinbeauvais, line 629: I thought you were passing in `showErrors` prop so the component itself could style this? Or is that only for NOB and THIS is for the addresses?

Rhetorical comment. Just food for thought for later. Is it Friday yet?

cameron-freshworks: hahah very soon sir!

The show errors prop is used on the input field rules/validations specifically. 
This is just highlight the main label as red when our local UI here determines when it's invalid. 
I didn't think this red text styling could/would apply everywhere under all circumstances, so applied it conditionally from the parent here.

severinbeauvais: > I didn't think this red text styling could/would apply everywhere under all circumstances, so applied it conditionally from the parent here.

I think you're right. Resolving this conversation.

severinbeauvais: Actually, it makes sense for the parent to style its components how it wants.

cameron-freshworks:
> LGTM.

Thank you!

All this aside, the api response for Firms addresses is changing soonish, so will be back in here shortly 😁

PR #251, opened 2022-03-23
Change of Name Refactors: Corrections/Alterations/Change Filings

cameron-freshworks:
*Issue #:* /bcgov/entity#11501

*Description of changes:*
* Updated resource files to include name change options
* UI refactors to support change of name for Firms
* Confirm Dialog version update and refactors for reusability
* Updated Validations
* Updated flow for invalid name request types
* Updated unit testing


package.json
```diff
@@ -17,14 +17,14 @@
     "@bcrs-shared-components/action-chip": "1.0.6",
     "@bcrs-shared-components/bread-crumb": "1.0.1",
     "@bcrs-shared-components/certify": "1.0.18",
-    "@bcrs-shared-components/confirm-dialog": "1.0.1",
-    "@bcrs-shared-components/contact-info": "1.0.29",
+    "@bcrs-shared-components/confirm-dialog": "1.0.3",
```
severinbeauvais, line 20: Should this be v1.1.0 ?

package.json
```diff
@@ -17,14 +17,14 @@
...
+    "@bcrs-shared-components/contact-info": "1.0.31",
```
severinbeauvais, line 21: Should this be v1.1.2 ?

src/mixins/name-request-mixin.ts
```diff
@@ -131,4 +122,14 @@ export default class NameRequestMixin extends Mixins(DateMixin, LegalApiMixin) {
...
+
+  /** Map Nr specific types to Lear recognized entity types. */
+  mapNameRequestType (nrType: CorpTypeCd): CorpTypeCd {
+    switch (nrType) {
+      case CorpTypeCd.BC_COMPANY: return CorpTypeCd.BENEFIT_COMPANY
+      case CorpTypeCd.BC_CORPORATION: return CorpTypeCd.BC_COMPANY
+      case CorpTypeCd.NR_SOLE_PROP: return CorpTypeCd.SOLE_PROP
+      default: return nrType
+    }
+  }
```
vysakh-menon-aot, line 134: I am not sure about this. @severinbeauvais did we not changed the approach to use legalType instead of entity_type_cd.

src/App.vue
```diff
@@ -765,6 +765,25 @@ export default class App extends Mixins(AuthApiMixin, CommonMixin, DateMixin, Fi
 
     this.setIsSaving(false)
   }
+
+  /**
+   * Helper to show the confirm dialogs.
+   * @param title The title content in dialog header
+   * @param message The content body
+   * @param action The action label
+   * @param cancel The cancel label
```
severinbeauvais, line 774: Clever boy!

I'm going to copy some of what you're done here. Just a couple of changes:
1. rename `action` to `yes`
2. rename `cancel` to `no`
3. use `no` option instead of cancel (which resolves the promise as False instead of rejecting it); subtle difference

severinbeauvais: (*) resolving to False means the result of open() below is false, whereas rejecting means to throw an error, which the catch() handles

severinbeauvais: (**) No is a valid negative action, where Cancel is "not an action".

If the dialog supported pressing Esc to close it, or a top-right "X" button, then that would be cancel.

src/App.vue
```diff
@@ -428,6 +428,11 @@ export default class App extends Mixins(AuthApiMixin, CommonMixin, DateMixin, Fi
       this.goToDashboard()
     })
 
+    // listen to confirm events
+    this.$root.$on('confirm-dialog', (title, message, action) => {
+      this.showConfirmDialog(title, message, action)
+    })
```
severinbeauvais, line 434: I like the idea here but the calling code can't wait for the result of the confirm dialog. Is this more of an "acknowledgement dialog" then?

I suggest renaming this or moving the dialog to the component that uses it.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -152,9 +152,7 @@
       <!-- Name Request Applicant -->
       <v-row no-gutters v-if="(isAlterationFiling || isChangeFiling) && hasNewNr" class="sub-section">
         <v-col cols="3">
-          <v-layout column>
-            <label><strong>Name Request Applicant</strong></label>
-          </v-layout>
+          <label class="pr-4"><strong>Name Request Applicant</strong></label>
```
severinbeauvais, line 155: Check whether we have base CSS to make all labels bold. If not, I know a lot of components have local CSS to do so.

I think we should move towards global CSS for this if we don't it already. (I can't think of when labels wouldn't be bold.)

src/mixins/common-mixin.ts
```diff
@@ -128,15 +128,6 @@ export default class CommonMixin extends Vue {
     }
   }
 
-  /**
-   * Checks if the specified entity type matches the current entity's type.
-   * @param entityType the entity type to check
-   * @return True if the check passes, else False
-   */
-  entityFilter (entityType: CorpTypeCd): boolean {
```
severinbeauvais, line 136: Good riddance :)

src/App.vue
severinbeauvais: ...Meanwhile, in a few cases in Create UI, I'm naming the No button "Cancel" 🙄 But semantically this is correct: the user is selecting a valid negative action instead of dismissing (canceling) the dialog.

Functionally this makes no difference, but it's more correct FWIW.

package.json
cameron-freshworks: Doh, yes to this and the Q below 👍 
Updated!

package.json
cameron-freshworks: Updated!

src/mixins/name-request-mixin.ts
cameron-freshworks: Oh nice! 
Will look into this, thank you :)

src/App.vue
cameron-freshworks: Thanks! 
Sounds good, will update

src/App.vue
cameron-freshworks: You are right, in this scenario it is more of an acknowledgement. 
However, i did want to get away from having multiple instances of this confirm dialog and i think it still fits this purpose. 

Long term, i want to remove all instances and refs of the confirm dialog in this app, and just utilize the singular instance on the root level.

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: Will look! 👁️

src/mixins/common-mixin.ts
cameron-freshworks: hahaha
👋

src/App.vue
severinbeauvais: Yes, I like the idea, but I don't think calling code can ever wait for this dialog and get a result. Events are async.

cameron-freshworks: You are probably right. 
Will look into this, if there isn't a relatively elegant solution, will remove and relocate locally into the component :)

src/mixins/name-request-mixin.ts
cameron-freshworks: After looking into this, the `legalType` in the nr response doesn't exactly match up with what lear would expect. 

For example, a change of name Sole Prop NR comes back as `CSP` and for General Partners `CGP`. Our system recognizes these types as `SP` and `GP`.

TLDR: some mapping is still going to be required, which property is the best to map i'm not sure so unless the old property is going to become deprecated, i think i will leave this mapping?

severinbeauvais: That's weird. I'm using a NR to create a sole prop and it's coming across as SP:

image

CSP = CHANGE of SP ?

vysakh-menon-aot: Yeah CSP and CGP

cameron-freshworks: After a discussion offline, we found a solution. Will update when api changes have been implemented! 🔥

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: Labels were already `bolded`. 
This was redundant! Removed

src/App.vue
cameron-freshworks: I tried a state approach but didn't like it. Reverted changes. 
Looking at the Mixin approach now 👀

vysakh-menon-aot (approved):

src/interfaces/correction-interfaces/correct-name-interface.ts
```diff
@@ -19,6 +19,7 @@ export interface NrCorrectionIF {
 }
 
 export interface NrResponseIF {
```
severinbeauvais, line 21: At some point we should commonize/share the NR IFs.

Create UI uses type "any" for the NR API response and then maps it to this:
https://github.com/bcgov/business-create-ui/blob/main/src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts

... here:
https://github.com/bcgov/business-create-ui/blob/main/src/mixins/name-request-mixin.ts#L17

🤷‍♂️

src/components/common/YourCompany/CompanyName/CorrectNameRequest.vue
```diff
@@ -138,36 +145,42 @@ export default class CorrectNameRequest extends Mixins(CommonMixin, NameRequestM
...
+        if (this.getEntityType !== nr.legalType) {
+          // Invalid NR type, inform parent the process is done and prompt confirm dialog
+          this.emitDone()
+
+          const dialogContent = `<p class="info-text">This ${this.getCorpTypeDescription(nr.entity_type_cd)} ` +
+            `Name Request does not match the current business type ` +
+            `<b>${this.getCorpTypeDescription(this.getEntityType)}</b>.\n\n` +
+            `The Name Request type must match the business type before you can continue.</p>`
+
+          this.$root.$emit(
```
severinbeauvais, line 174: Since this call is not async (we don't wait for it), what is the next thing the user sees, and what are their options?

src/interfaces/correction-interfaces/correct-name-interface.ts
severinbeauvais: ^^ not an action or request to change

Just something to keep in mind for sooner or later.

src/App.vue
```diff
@@ -765,6 +765,23 @@ export default class App extends Mixins(AuthApiMixin, CommonMixin, DateMixin, Fi
...
+   * @param yes The YES action label
+   * @param no The NO cancel label
+   * */
+  async showConfirmDialog (title: string, message: string, yes: string, no: string = null): Promise<any> {
+    return this.$refs.confirm.open(title, message, {
+      width: '45rem',
+      persistent: true,
+      yes: yes,
+      no: null,
+      cancel: no
```
severinbeauvais, line 782: Do you think it makes more sense to make no=no and leave cancel for the case where the user dismisses the dialog? (I think Esc does work to dismiss the dialog, and will reject the promise, ie, throw an exception here.)

src/App.vue
```diff
@@ -765,6 +765,23 @@ export default class App extends Mixins(AuthApiMixin, CommonMixin, DateMixin, Fi
...
+
+  /**
+   * Helper to show the confirm dialogs.
+   * @param title The title content in dialog header
+   * @param message The content body
+   * @param yes The YES action label
+   * @param no The NO cancel label
+   * */
+  async showConfirmDialog (title: string, message: string, yes: string, no: string = null): Promise<any> {
+    return this.$refs.confirm.open(title, message, {
```
severinbeauvais, line 777: Check your interface for ConfirmDialogType -- it should return boolean.

src/App.vue
cameron-freshworks: Sorry, absolutely. this is was a mistake when i was reverting some testing i was doing with regards to globalizing the Confirm Dialog.

src/interfaces/correction-interfaces/correct-name-interface.ts
cameron-freshworks: Absolutely, i imagine this is quite a common request response!

severinbeauvais: In Create UI this is NOT the request response. It's the local object.

Actually, we do have a shared IF for this:
https://github.com/bcgov/bcrs-shared-components/blob/main/src/interfaces/nameRequest-interface.ts

Ultimately we should have an IF (schema) for the API response -- the raw NR JSON.

cameron-freshworks:
> I think it's OK.
> 
> But I don't think you'll be able to use your `confirm-dialog` event since that displays a dialog that we don't wait for (which is usually the whole point of a dialog).

Current implementation works only because it's a confirmation. 
It's more of a notification to the user the type mismatched, and prompts them to close the dialog. 
Locally where the dialog is called, it's stops the process and returns the user to the NR form. 

You are right though, im still working at trying to globalize the confirmation dialog, if not, will relocate locally to the Component.

src/App.vue
cameron-freshworks: Yep sorry, i was reverting changes here in order to push up the other commits for PR comments. 
Still wip

severinbeauvais:
> ... im still working at trying to globalize the confirmation dialog, if not, will relocate locally to the Component.

I totally agree with this. I hope the first way works; if not then the second way is fine.

src/App.vue
cameron-freshworks: Fixed!

src/App.vue
cameron-freshworks: Fixed!

src/components/common/YourCompany/CompanyName/CorrectNameRequest.vue
cameron-freshworks: There only option here is 'OK' 
So no options really 😆

PR #253, opened 2022-04-05
SP People and Roles Integration (+Init GP's)

cameron-freshworks:
*Issue #:* /bcgov/entity#10689

*Description of changes:*
* Refactoring PeopleAndRoles/OrgPersons for Corrections/Alterations for SP/GP (ChangeOfReg) functionality
* Layout and design updates for previous iterations of components to bring up to date
* Implements HelpText component
* Updating unit test for previous Corrections and Change of Registrations
* Includes resourcing in components to accommodate new variations
* Update field validations on People and Roles and when to apply them, also includes 

**NOTE:**
* Partners Orgs is still work to come. This just sets up the foundations for it, the nuances are still to come.
* API is still missing `officer.email` values but the filing template mixin is built to handle it when it is returned from the `/parties`


src/components/Edit/ChangeSummary.vue
```diff
@@ -43,25 +43,44 @@
...
+      <v-divider class="mx-8" />
       <div id="address-summary-section" class="section-container">
         <OfficeAddresses :isSummaryView="true" />
       </div>
     </template>
+
+    <!-- Org Persons -->
+    <template v-if="hasOrgPersonChanged">
+      <v-divider class="mx-8" />
+      <div id="org-person-summary-section" class="section-container">
```
severinbeauvais, line 57: Have you considered making this an actual `<section>` ? (Or article.) (And some of the other divs on this page.)

src/components/Edit/ChangeSummary.vue
```diff
@@ -43,25 +43,44 @@
...
+        <v-row no-gutters>
+          <v-col cols="12" sm="3">
+            <label><strong>Proprietor Information</strong></label>
```
severinbeauvais, line 60: Labels may already have bold styles, and if not then they should :)

src/assets/styles/base.scss
```diff
@@ -82,6 +82,7 @@ label {
 }
 
 p {
+  font-size: 1rem;
```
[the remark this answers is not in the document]
cameron-freshworks: Indeed! Will fix

src/components/Edit/ChangeSummary.vue
cameron-freshworks: I hadn't.
What is the difference/relevance, is it mostly semantics?
Real question :) I don't actually know the purpose!

src/components/Edit/ChangeSummary.vue
cameron-freshworks: They do indeed, thank you, will fix!

src/components/common/HelpSection.vue
```diff
@@ -0,0 +1,78 @@
...
+        </header>
+
+        <p v-for="(item, index) in items" :key="index" class="mt-4 py-1" v-html="item" />
+
+        <!-- BC Registry Contacts -->
+        <BcRegContacts class="mb-6" :direction="'col'"/>
+
+        <label>Hours of Operation:</label>
+        <p>Monday to Friday, 8:30am - 4:30pm Pacific Time</p>
+      </section>
```
severinbeauvais, line 22: No "hide" button?

src/components/common/PeopleAndRoles/CurrentDirectors.vue
```diff
@@ -3,22 +3,22 @@
 
     <v-card flat>
       <!-- Header -->
-      <div class="header-container">
+      <div class="header-container section-container">
```
severinbeauvais, line 6: `<section>` or `<article>`? Just a thought.

src/components/common/PeopleAndRoles/CurrentDirectors.vue
```diff
@@ -31,7 +31,7 @@
             :key="`director:${index}`"
           >
             <!-- Name + Badge -->
-            <td class="text-truncate">
+            <td class="text-truncate px-0">
```
severinbeauvais, line 34: By the way, UX/UI decided to remove this truncation/tooltip mechanism in Create UI. Did they say anything about this, here?

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-  <div flat id="list-people-roles" class="mt-0">
+  <div id="list-people-roles" class="mt-0">
```
severinbeauvais, line 2: I'd remove the mt-0 as it's probably needed and may conflict with classes for this component set in the parent view.

src/components/common/HelpSection.vue
```diff
@@ -0,0 +1,78 @@
...
+    <v-expand-transition v-if="helpToggle">
+      <section class="help-section info-text">
+        <header class="help-header mt-4">
+          <h2>Contact BC Registries</h2>
+        </header>
+
+        <p v-for="(item, index) in items" :key="index" class="mt-4 py-1" v-html="item" />
+
+        <!-- BC Registry Contacts -->
+        <BcRegContacts class="mb-6" :direction="'col'"/>
```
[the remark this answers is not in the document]
cameron-freshworks: Yea :( 

Definitely think we should. Due to the scope of this ticket and it's following commit, i would like to defer adding scope. 
Obviously there has been a few design/ux updates during development so my concern is there is already a lot of hidden tasks.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -362,8 +468,10 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin) {
 }
 
 // italicize the delivery instructions in the base address component
-::v-deep .address-block__info-row:last-of-type {
-  padding-top: 0.75rem;
-  font-style: italic;
+::v-deep {
+  .address-block {
```
severinbeauvais, line 472: If you're targeting only the delivery instructions then you'll need to use `::v-deep .base-address .address-block .delivery-instructions`... otherwise, the comment on line 470 is wrong.

severinbeauvais: For comparison: https://github.com/bcgov/business-create-ui/blob/299527f5e6b0f353f0b16f0891eb2ce52d547f52/src/components/common/ListPeopleAndRoles.vue#L316

src/components/common/HelpSection.vue
cameron-freshworks: Well, initially there wasn't, but i see the designs have since been updated to include it. 
I promise i triple checked this because i noticed this specifically. 

Looks like i will be putting it in now 😩

src/components/common/PeopleAndRoles/CurrentDirectors.vue
cameron-freshworks: Definitely can do! I'm just curious on the meaning behind it

src/components/common/PeopleAndRoles/CurrentDirectors.vue
cameron-freshworks: Oh, they did not. 
How do they handle the really long person/org names?

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Removed 👍

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Comment is obsolete, carry over. The component itself now provides the styling that was being implemented here(previously)
Removed comment

src/assets/styles/base.scss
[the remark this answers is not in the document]
cameron-freshworks: Fixed

src/components/Edit/ChangeSummary.vue
cameron-freshworks: Fixed

src/components/Edit/ChangeSummary.vue
cameron-freshworks: After a little research, will update. 
Looks like article is the more appropriate tag. Section does sound a little more root level, ie Header / Body / Footer
Obviously this brings into question some of our generic class names, but i won't get too hung up on it at the moment.

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -425,16 +468,16 @@ export default class OrgPerson extends Mixins(CommonMixin) {
       person.officer.id = uuidv4()
     }
     person.mailingAddress = { ...this.inProgressMailingAddress }
-    if (this.isDirector) {
+    if (this.isDirector || this.isProprietor || this.isPartner) {
```
severinbeauvais, line 471: What about the completing party's delivery address (firms only)?

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -480,6 +523,55 @@ export default class OrgPerson extends Mixins(CommonMixin) {
     }
   }
 
+  /** Apply input field validations. */
+  applyValidation (): void {
```
severinbeauvais, line 527: I kinda like having all validations in the same place here.

I also like using `:rules="isValidation ? myRule : []"`.

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -664,4 +756,20 @@ li {
 ::v-deep .v-input--checkbox .v-messages:not(.error--text) {
   display: none;
 }
+
+::v-deep {
```
severinbeauvais, line 760: Please add a comment explaining what you're overriding here.

cameron-freshworks: For sure

src/components/common/PeopleAndRoles/OrgPerson.vue
cameron-freshworks: Indeed, i do as well. 
In this scenario, it seemed easy to apply in one place but both are "valid" :)

src/components/common/PeopleAndRoles/OrgPerson.vue
cameron-freshworks: Completing Party isn't a factor in this component (for firms) afaik
Completing party is manually added or pulled from the account for this filing in another component on the review page (not completed).

src/enums/roleTypes.ts
```diff
@@ -1,6 +1,8 @@
 export enum RoleTypes {
   COMPLETING_PARTY = 'Completing Party',
-  INCORPORATOR = 'Incorporator',
   DIRECTOR = 'Director',
-  SUBSCRIBER = 'Subscriber',
+  INCORPORATOR = 'Incorporator',
+  PROPRIETOR = 'proprietor',
+  PARTNER = 'partner',
```
vysakh-menon-aot, line 6: Why these 2 are not Pascal Case?

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -211,6 +273,18 @@ export default class PeopleAndRoles extends Mixins(CommonMixin) {
...
 
+  private get orgPersonLabel (): string {
+    return this.getResource.changeData?.orgPersonInfo.orgPersonLabel
+  }
+
+  private get orgPersonSubtitle (): string {
+    return this.getResource.changeData?.orgPersonInfo.subtitle
+  }
+
+  private get orgPersonHelp (): HelpSectionIF {
```
severinbeauvais, line 284: I've been changing my gets to protected since the template uses them. For these, it wouldn't matter if they were public (which is the default if not specified) since code external to this class could safely call them. Just food for thought.

severinbeauvais: (In VS Code, private variable/getters/methods show as "unused" when they are only used in the template.)

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -580,5 +580,8 @@ export default class YourCompany extends Mixins(
   .invalid-section .col-sm-3:first-child label {
     color: $app-red !important;
   }
+  #contact-info-edit-btn {
+    padding-right: 4px;
+  }
```
severinbeauvais, line 585: For a tiny tweak like this, I'd use `pr-1` in the template. Might make it easier to do future tweaking.

src/enums/roleTypes.ts
cameron-freshworks: Good question, currently the API is returning (`/parties`) these roles all lowercase, so this was in concert with that.

src/enums/orgPersonTypes.ts
```diff
@@ -0,0 +1,4 @@
+export enum OrgPersonTypes {
+  PARTIES = 'parties',
+  DIRECTORS = 'directors'
+}
```
severinbeauvais, line 4: What are these used for? Can PartyType or RoleType be used instead?

src/enums/roleTypes.ts
cameron-freshworks: Screen Shot 2022-04-05 at 12 10 13 PM

src/components/Edit/ChangeSummary.vue
cameron-freshworks: Updated here

src/interfaces/stepper-interfaces/PeopleAndRoles/org-person-interface.ts
```diff
@@ -1,31 +1,22 @@
...
     id: string
-    partyType: IncorporatorTypes
+    partyType: PartyTypes
     firstName: string
     middleName?: string
     lastName: string
-    organizationName: string
+    organizationName?: string
     email?: string
   }
```
severinbeauvais, line 13: Do you need `businessNumber` or `incorpNumber` here as well for prorietor-orgs ?

src/store/getters/state-getters.ts
```diff
@@ -401,15 +401,16 @@ export const hasFirmChanged = (state: StateIF): boolean => {
   return (
     hasBusinessNameChanged(state) ||
     hasNatureOfBusinessChanged(state) ||
-    officeAddressesChanged(state)
+    officeAddressesChanged(state) ||
+    hasOrgPersonChanged(state)
```
severinbeauvais, line 405: I suggest:
- rename officeAddressesChanged -> hasOfficeAddressesChanged
- rename hasOrgPersonChanged -> hasPeopleAndRolesChanged

Up to you though.

src/store/state/state-model.ts
```diff
@@ -97,7 +98,7 @@ export const stateModel: StateModelIF = {
     documentOptionalEmail: ''
   },
   officeAddresses: null,
-  peopleAndRolesStep: {
+  peopleAndRoles: {
```
severinbeauvais, line 101: 👍

src/components/common/PeopleAndRoles/OrgPerson.vue
cameron-freshworks: Updated

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: I haven't noticed this in Webstorm but it is probably due to some configuration i have?

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: For sure, not sure why i did this here!

tests/unit/ListPeopleAndRoles.spec.ts
```diff
@@ -136,6 +140,116 @@ const peopleAndRoles = [
...
+const gpPeopleAndRoles = [
+  {
+    officer: {
+      id: '0',
+      firstName: 'Romeo',
+      lastName: 'Whitehead',
+      middleName: 'D',
+      organizationName: '',
+      partyType: 'person',
+      email: 'completing-party@example.com'
```
severinbeauvais, line 152: I see differences between this and Create UI.

I created a separate `describe` for FM due to slightly different data. See https://github.com/bcgov/business-create-ui/blob/299527f5e6b0f353f0b16f0891eb2ce52d547f52/tests/unit/ListPeopleAndRoles.spec.ts#L249.

src/components/common/YourCompany/YourCompany.vue
cameron-freshworks: O right, because this is a Shared Component. There is no local template to apply the helper class too.

tests/unit/OrgPerson.spec.ts
```diff
@@ -327,6 +329,7 @@ describe('Org/Person component', () => {
 
   it('Emits "reset" event when clicking Cancel button', async () => {
     const wrapper = createComponent(validOrgData, 0, null)
+    wrapper.vm.applyValidation()
```
severinbeauvais, line 332: You don't need to wait a tick after this (as you do in the main code)?

I'd be tempted to move the "nextTick" in the main code to the bottom of applyValidation() and then await it (just so it's self-contained).

src/enums/orgPersonTypes.ts
cameron-freshworks: This is used for requests to the API, where we have different end points to retrieve directors vs parties etc.
PartyType is person or org, RoleType is Incorporator, completingParty proprietor etc..
I can update the name of the enum to `apiOrgPersonType` ?

src/interfaces/stepper-interfaces/PeopleAndRoles/org-person-interface.ts
cameron-freshworks: Probably yes, will be done the following commit focused on the Corp Partners

src/main.ts
```diff
@@ -81,7 +81,8 @@ async function start () {
             appDkBlue: '#38598a', // same as $app-dk-blue
             error: '#d3272c', // same as $app-red
             success: '#1a9031', // same as $app-green
-            gray7: '#495057' // same as $gray7
+            gray7: '#495057', // same as $gray7
+            gray9: '#212529' // Same as $gray9
```
severinbeauvais, line 85: You _may_ want to add these to the Vuetify theme as well, so you can use these named colours in Vuetify components (eg, `color="gray9"`).

src/enums/roleTypes.ts
vysakh-menon-aot: This could impact other roles as well. eg: `incorporator`, `completing_party`. So either we need to fix the parties api or ui to consider other roles as well.

src/mixins/filing-template-mixin.ts
```diff
@@ -315,6 +316,15 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
    */
   buildChangeFiling (isDraft: boolean): ChangeFirmIF {
+    let parties = this.getPeopleAndRoles
+
+    // if filing and paying, filter out removed entities and omit the 'action' properties
+    if (!isDraft) {
+      // Filter out parties actions
+      parties = parties.filter(x => x.action !== ActionTypes.REMOVED)
+        .map((x) => { const { action, ...rest } = x; return rest })
+    }
```
severinbeauvais, line 326: You probably don't need to do this. The API should ignore fields it doesn't care about.

Nice map though :)

src/mixins/filing-template-mixin.ts
```diff
@@ -359,6 +369,10 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
       filing.changeOfRegistration.business.naics = this.getCurrentNaics
     }
 
+    if (this.hasOrgPersonChanged) {
```
severinbeauvais, line 372: I'd move the `parties` code above inside this statement, so it only gets executed if needed instead of always.

src/mixins/legal-api-mixin.ts
```diff
@@ -23,6 +24,8 @@ export default class LegalApiMixin extends Vue {
    * @returns a promise to return the filing of the specified type
    */
   async fetchFilingById (id: number): Promise<any> {
+    // Override api version for /filings requests as V2 is currently broken for Correction Filings
+    axios.defaults.baseURL = 'https://legal-api-dev.apps.silver.devops.gov.bc.ca/api/v1'
```
severinbeauvais, line 28: Is this only temporary?

If short-term, is there a ticket to clean this up?

If long-term then I'd like to see a different approach.

src/resources/Correction/BenefitCompanyStatementResource.ts
```diff
@@ -13,7 +13,13 @@ export const BenefitCompanyStatementResource = {
       CorrectionTypes.CORRECT_NEW_NR,
       CorrectionTypes.CORRECT_NAME_TO_NUMBER,
       CorrectionTypes.CORRECT_NAME
-    ]
+    ],
+    orgPersonInfo: {
+      orgPersonLabel: 'People and Roles',
+      subtitle: 'You must have a minimum of two partners. You can add or remove partners (individual person or ' +
+        'business) as well as change the mailing and delivery addresses and email address of individual people and ' +
+        'business partners that were added manually during registration.'
```
severinbeauvais, line 21: Where did this come from?

severinbeauvais (approved):
Nice looking code! I realize some of it is in flux atm.

Do you need a FF, or no because there are no firms in the system atm?

I assume you've fully tested these changes w.r.t. other entity types. Please also call out these changes so Riyaz learns about them.

src/components/Edit/ChangeSummary.vue
severinbeauvais: A section can contain articles, and an article can contain sections. Theoretically, they are both at the same level. In practise, I try for consistency within the current app or page. In other words, if the main blocks are sections then I use articles for sub-blocks. Or vice-versa.

src/enums/roleTypes.ts
cameron-freshworks: It could be that it's different end points. The `/parties` returns lowercase, the `/directors` roles are structured a little different and they are pascal case.

src/components/common/HelpSection.vue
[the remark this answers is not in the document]
severinbeauvais: I know how you feel. And I agree with you: defer this.

src/enums/roleTypes.ts
cameron-freshworks: If the api is updated, we can update here as well. Currently the UI just handles both variations.

src/components/common/HelpSection.vue
severinbeauvais: I laughed a little but I also feel for you. Here:

```
<div class="help-btn bottom" @click="helpToggle = !helpToggle">Hide Help</div>
```

cameron-freshworks: hahah its all we can do..

thank you, will insert 😁

src/components/common/PeopleAndRoles/CurrentDirectors.vue
severinbeauvais: It's giving a different name to a block of code instead of using "div" all the time. Some (maybe all) accessibility programs look at such things to present a hierarchical description of the page.

It's like in Word: you use styles for headings and stuff, right? (And on that note, it can make it easier to style articles or sections instead of misc divs.)

severinbeauvais: If you don't have time, defer this. It's that extra 10% to get to 110% :)

src/enums/roleTypes.ts
cameron-freshworks: I'm sorry, i was incorrect, the first variation is being returned via CORRECTION filing, thats where the structure differs
NOT the `/directors` endpoint

vysakh-menon-aot: oh! ok

cameron-freshworks: That being said, i'm not sure if the upper/lowercase will even matter. Will update the UI and see if it has an impact

src/components/common/PeopleAndRoles/OrgPerson.vue
severinbeauvais: Ask UI/UX for confirmation. In Create UI, CP gets a delivery address for firms.

src/enums/roleTypes.ts
cameron-freshworks: Yea, it actually doesn't matter. The UI is handling it either way. 
Yes the api is a tiny bit inconsistent between Corrections and our Change/Alteration filings, but it could also be how the value is saved from an Incorporation to begin with 🤷 
Either way, super low impact. Updated locally here for consistency, good catch!

severinbeauvais: And for what it's worth, Create UI save the reg filing with roleType in sentence case:

image

@vysakh-menon-aot Let me know if I need to change this.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Works either way atm (looks like TS doesn't enforce visibility rules with the template). Resolving.

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: You're right. I didn't notice this is under a v-deep. Resolving.

src/components/common/HelpSection.vue
severinbeauvais: Oh, and this, too:

https://github.com/bcgov/business-create-ui/blob/299527f5e6b0f353f0b16f0891eb2ce52d547f52/src/components/common/HelpSection.vue#L69

src/enums/roleTypes.ts
vysakh-menon-aot: @severinbeauvais This impacts api validation if you do so. So we need to take it as separate issue and analyze the impacts.

severinbeauvais: The roleType above was not changed for registrations -- only extended.

The current API code seems to work: I filed the above registration and the filing output looks OK.

vysakh-menon-aot: Right. It won't break registration validation but report breaks. Also I was looking into IA validation and report templates. which will break

severinbeauvais: Thanks for investigating. Is it a quick fix or do we need to coordinate multiple, dependent changes?

What of this? https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/party.json#L74

severinbeauvais: My goal is to identify (and keep track of) anything that needs changing/fixing.

cameron-freshworks: For sure, also following here as i don't know what the solution is. 
Should we discuss on a call tomorrow?

vysakh-menon-aot: Yes @severinbeauvais if we are changing it we need to start from schema to fixing existing filing_json data in Prod. Since the schema is defined to use Pascal Case we can update `/parties and `/directors` endpoint to provide Pascal Case role.

severinbeauvais: How about we chat at 10am tomorrow (after splanning)?

vysakh-menon-aot: sure

src/components/Edit/ChangeSummary.vue
cameron-freshworks: I like it! Will try and follow this more closely in the future. 
Updated this file

src/components/common/HelpSection.vue
cameron-freshworks: Added!

src/components/common/PeopleAndRoles/CurrentDirectors.vue
cameron-freshworks: Updated!

src/store/getters/state-getters.ts
cameron-freshworks: Updated!

tests/unit/ListPeopleAndRoles.spec.ts
cameron-freshworks: There is for sure. I created a peopleAndRoles for the separate GP test suite below this.
Will expand in following commits as well.

tests/unit/OrgPerson.spec.ts
cameron-freshworks: Ok, good idea. Will try that

cameron-freshworks: It wasn't required in the test, probably because it awaits a process after updating the input fields.

src/main.ts
cameron-freshworks: By doing this, you can already use them within the Vuetify component templates

src/mixins/filing-template-mixin.ts
cameron-freshworks: Just following what we've done in the past with the actions. Will do a little research and remove if possible

src/mixins/filing-template-mixin.ts
cameron-freshworks: Updated

src/mixins/legal-api-mixin.ts
cameron-freshworks: The api is(was) broken on V2 for fetching these filings, will create a ticket.

src/main.ts
severinbeauvais: Whoops, I thought this was the styles.scss file. You're all good (and ahead of me).

src/resources/Correction/BenefitCompanyStatementResource.ts
cameron-freshworks: Probably a copy past mistake. Didn't notice it because it doesn't render for non firms. Will remove.

lekshmimallika-aot (approved):

tests/unit/OrgPerson.spec.ts
cameron-freshworks: Added them for consistency though, as not all tests had similar processes.

src/resources/Correction/BenefitCompanyStatementResource.ts
cameron-freshworks: > Nice looking code! I realize some of it is in flux atm.
> 
> Do you need a FF, or no because there are no firms in the system atm?
> 
> I assume you've fully tested these changes w.r.t. other entity types. Please also call out these changes so Riyaz learns about them.

Yea, the dev work has sort of a required a Correction/Alteration/ChangeOfReg filing to be open at all times, at least thats how i do it. Especially since we were pulling some bits out and into resource files. 

We currently have the flag to protect the filing as a whole, which should be enough for now! Only changes visible in our Correction filing is aesthetic stuff, layout improvements :)

cameron-freshworks:
> Nice looking code! I realize some of it is in flux atm.
> 
> Do you need a FF, or no because there are no firms in the system atm?
> 
> I assume you've fully tested these changes w.r.t. other entity types. Please also call out these changes so Riyaz learns about them.

Thank you!
Yea, the dev work has sort of a required a Correction/Alteration/ChangeOfReg filing to be open at all times, at least thats how i do it. Especially since we were pulling some bits out and into resource files. 

We currently have the flag to protect the filing as a whole, which should be enough for now! Only changes visible in our Correction filing is aesthetic stuff, layout improvements :)

src/components/Edit/ChangeSummary.vue
severinbeauvais: Looks good!

src/components/common/PeopleAndRoles/CurrentDirectors.vue
severinbeauvais: By wrapping.

Maybe ask them about it. It's just taking out a bit of code so pretty easy if needed.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Please add comment describing what this override is doing.

src/mixins/filing-template-mixin.ts
```diff
@@ -587,28 +601,20 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     }
     this.setOfficeAddresses(addresses || entitySnapshot.addresses)
 
+    let parties = filing.changeOfRegistration.parties || entitySnapshot.orgPersons
```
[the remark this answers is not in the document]
severinbeauvais: Can `orgPersons` be null or undefined? If so the line 608 would break.

src/enums/orgPersonTypes.ts
severinbeauvais: You could rename it, or you could just add a JS Doc comment explaining what this is used for.

cameron-freshworks: Updated

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Updated

src/components/common/PeopleAndRoles/CurrentDirectors.vue
cameron-freshworks: Will leave it in for now, as it was a feature as part of Corrections. 
Will bring up during UX review 👍

src/mixins/filing-template-mixin.ts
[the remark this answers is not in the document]
cameron-freshworks: I don't believe `parties` could be defined as `{}`, no. 
It's either defined in the Filing because it was modified and added, or the snapshot will provide the data. If the snapshot failed to get the information it is handled in `fetchFirmSnapshot()`

src/enums/roleTypes.ts
severinbeauvais: What we decided:
- API should return role types according to schema (title case); Vysakh will create a ticket for this and will ask Linda whether it's OK to do it this sprint
- Filings UI will need its enum updated; Cam can do this as part of THIS ticket (10689) (if Vysakh's ticket is approved)
- the code above will need to be updated; Cam can do this in this PR (if Vysakh's ticket is approved)
- if Vysakh's ticket is not approved for this sprint then we carry on with the current enum differences and fix it later

cameron-freshworks: Thanks you!

tests/unit/OrgPerson.spec.ts
severinbeauvais: I was thinking you'd move the "nextTick" inside "applyValidation".

The principle behind this is that the caller shouldn't be concerned about calling nextTick after calling applyValidation: it's one call (though it will have to awaited). If applyValidation requires a short delay to properly complete then it should be done inside applyValidation.

This is [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns), where each chunk of code should need to know as little as possible about other code.

cameron-freshworks: I did move it inside. 

The test suites however still require this (some of them), i just added to all for consistency.

src/mixins/legal-api-mixin.ts
severinbeauvais: To be fixed later. Resolving.

src/mixins/filing-template-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: If it's not in the filing then it's undefined or null? That works.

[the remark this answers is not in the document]
cameron-freshworks: You bet, most parts of the change filing can be null or undefined.

tests/unit/OrgPerson.spec.ts
severinbeauvais: Oh I'm sorry, I didn't notice you updated this already.

Probably the reason some tests require this is because they need to await applyValidation.

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -351,6 +377,8 @@ export default class OrgPerson extends Mixins(CommonMixin) {
    * Called when user clicks Done button.
    */
   private validateOrgPersonForm (): void {
+    this.applyValidation()
```
severinbeauvais, line 380: At first glance, this seems like it should be awaited.

It probably works anyway because the `validate()` calls below just handle the new rules. I'm not entirely sure.

I think it's good form to await an async function in all cases.

src/mixins/filing-template-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: Thanks for checking. Resolving.

src/components/common/PeopleAndRoles/OrgPerson.vue
cameron-freshworks: kk, will update

tests/unit/OrgPerson.spec.ts
severinbeauvais: Side note: I've found (to my surprise) that most things that don't look right can be explained. So, now, if something doesn't look right, I look for the explanation... Typically it's because I'm doing something incorrectly 😆 

Moral of the story: if it doesn't look right, try to figure out why.

cameron-freshworks: Found that i didn't need the NextTick inside, once we awaited the call

PR #256, opened 2022-04-13
Implements Partners, persons and orgs

cameron-freshworks:
*Issue #:* /bcgov/entity#1173

*Description of changes:*
* Implements Partners, Orgs Adding/Editing improved UX and functionality
* Updates Corrections stying
* Incorporates unique badges dependant on content changed, preserves badges for Corrections
* Updates unit test
* updated action to support multiple if required.
* removed V1 api override for /filings GET requests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "2.1.5",
+  "version": "2.1.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -8791,68 +8791,6 @@
           "dev": true,
           "requires": {
             "has-flag": "^3.0.0"
-          }
-        }
-      }
-    },
-    "fork-ts-checker-webpack-plugin-v5": {
-      "version": "npm:fork-ts-checker-webpack-plugin@5.2.1",
-      "resolved": "https://registry.npmjs.org/fork-ts-checker-webpack-plugin/-/fork-ts-checker-webpack-plugin-5.2.1.tgz",
-      "integrity": "sha512-SVi+ZAQOGbtAsUWrZvGzz38ga2YqjWvca1pXQFUArIVXqli0lLoDQ8uS0wg0kSpcwpZmaW5jVCZXQebkyUQSsw==",
-      "dev": true,
-      "optional": true,
-      "requires": {
-        "@babel/code-frame": "^7.8.3",
-        "@types/json-schema": "^7.0.5",
-        "chalk": "^4.1.0",
-        "cosmiconfig": "^6.0.0",
-        "deepmerge": "^4.2.2",
-        "fs-extra": "^9.0.0",
-        "memfs": "^3.1.2",
-        "minimatch": "^3.0.4",
-        "schema-utils": "2.7.0",
-        "semver": "^7.3.2",
-        "tapable": "^1.0.0"
-      },
-      "dependencies": {
-        "chalk": {
-          "version": "4.1.2",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-4.1.2.tgz",
-          "integrity": "sha512-oKnbhFyRIXpUuez8iBMmyEa4nbj4IOQyuhc/wy9kY7/WVPcwIO9VA668Pu8RkO7+0G76SLROeyw9CpQ061i4mA==",
-          "dev": true,
-          "optional": true,
-          "requires": {
-            "ansi-styles": "^4.1.0",
-            "supports-color": "^7.1.0"
-          }
-        },
-        "lru-cache": {
-          "version": "7.7.3",
-          "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-7.7.3.tgz",
-          "integrity": "sha512-WY9wjJNQt9+PZilnLbuFKM+SwDull9+6IAguOrarOMoOHTcJ9GnXSO11+Gw6c7xtDkBkthR57OZMtZKYr+1CEw==",
-          "dev": true,
-          "optional": true
-        },
-        "schema-utils": {
-          "version": "2.7.0",
-          "resolved": "https://registry.npmjs.org/schema-utils/-/schema-utils-2.7.0.tgz",
-          "integrity": "sha512-0ilKFI6QQF5nxDZLFn2dMjvc4hjg/Wkg7rHd3jK6/A4a1Hl9VFdQWvgB1UMGoU94pad1P/8N7fMcEnLnSiju8A==",
-          "dev": true,
-          "optional": true,
-          "requires": {
-            "@types/json-schema": "^7.0.4",
-            "ajv": "^6.12.2",
-            "ajv-keywords": "^3.4.1"
-          }
-        },
-        "semver": {
-          "version": "7.3.6",
-          "resolved": "https://registry.npmjs.org/semver/-/semver-7.3.6.tgz",
-          "integrity": "sha512-HZWqcgwLsjaX1HBD31msI/rXktuIhS+lWvdE4kN9z+8IVT4Itc7vqU2WvYsyD6/sjYCt4dEKH/m1M3dwI9CC5w==",
-          "dev": true,
-          "optional": true,
-          "requires": {
-            "lru-cache": "^7.4.0"
           }
         }
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "2.1.3",
+  "version": "2.1.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -3010,14 +3010,11 @@
           }
         },
         "lru-cache": {
-          "version": "6.0.0",
-          "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-6.0.0.tgz",
-          "integrity": "sha512-Jo6dJ04CmSjuznwJSS3pUeWmd/H0ffTlkXXgwZi+eq1UCmqQwCh+eLsYOYCwY991i2Fah4h1BEMCx4qThGbsiA==",
+          "version": "7.8.1",
+          "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-7.8.1.tgz",
+          "integrity": "sha512-E1v547OCgJvbvevfjgK9sNKIVXO96NnsTsFPBlg4ZxjhsJSODoH9lk8Bm0OxvHNm6Vm5Yqkl/1fErDxhYL8Skg==",
           "dev": true,
-          "optional": true,
-          "requires": {
-            "yallist": "^4.0.0"
-          }
+          "optional": true
         },
         "schema-utils": {
           "version": "2.7.0",
@@ -3032,21 +3029,14 @@
           }
         },
         "semver": {
-          "version": "7.3.5",
-          "resolved": "https://registry.npmjs.org/semver/-/semver-7.3.5.tgz",
-          "integrity": "sha512-PoeGJYh8HK4BTO/a9Tf6ZG3veo/A7ZVsYrSA6J8ny9nb3B1VrpkuN+z9OE5wfE5p6H4LchYZsegiQgbJD94ZFQ==",
+          "version": "7.3.6",
+          "resolved": "https://registry.npmjs.org/semver/-/semver-7.3.6.tgz",
+          "integrity": "sha512-HZWqcgwLsjaX1HBD31msI/rXktuIhS+lWvdE4kN9z+8IVT4Itc7vqU2WvYsyD6/sjYCt4dEKH/m1M3dwI9CC5w==",
           "dev": true,
           "optional": true,
           "requires": {
-            "lru-cache": "^6.0.0"
+            "lru-cache": "^7.4.0"
           }
-        },
-        "yallist": {
-          "version": "4.0.0",
-          "resolved": "https://registry.npmjs.org/yallist/-/yallist-4.0.0.tgz",
-          "integrity": "sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A==",
-          "dev": true,
-          "optional": true
         }
       }
     },
--- a/src/components/Edit/ChangeSummary.vue
+++ b/src/components/Edit/ChangeSummary.vue
@@ -57,7 +57,7 @@
       <article id="org-person-summary-section" class="section-container">
         <v-row no-gutters>
           <v-col cols="12" sm="3">
-            <label>Proprietor Information</label>
+            <label>{{ firmLabel }} Information</label>
           </v-col>
         </v-row>
         <v-row no-gutters class="mt-4">
@@ -74,7 +74,7 @@
 import { Component, Mixins, Prop } from 'vue-property-decorator'
 import { Action, Getter } from 'vuex-class'
 import { OfficeAddresses, ListPeopleAndRoles } from '@/components/common'
-import { ActionBindingIF, NaicsIF, OrgPersonIF } from '@/interfaces'
+import { ActionBindingIF, NaicsIF, OrgPersonIF, ResourceIF } from '@/interfaces'
 import { DateMixin, EnumMixin, FilingTemplateMixin, LegalApiMixin, PayApiMixin } from '@/mixins'
 
 @Component({
@@ -97,6 +97,7 @@ export default class ChangeSummary extends Mixins(
   @Getter getPeopleAndRoles!: OrgPersonIF[]
   @Getter hasOfficeAddressesChanged!: boolean
   @Getter hasPeopleAndRolesChanged!: boolean
+  @Getter getResource!: ResourceIF
 
   // Change flag getters
   @Getter hasBusinessNameChanged!: boolean
@@ -113,6 +114,11 @@ export default class ChangeSummary extends Mixins(
     if (this.getApprovedName) return this.getApprovedName
 
     return `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.`
+  }
+
+  /** The firm type label. */
+  get firmLabel (): string {
+    return this.getResource.changeData.orgPersonInfo.orgPersonLabel
   }
 }
 </script>
--- a/src/components/common/BcRegContacts.vue
+++ b/src/components/common/BcRegContacts.vue
@@ -24,14 +24,14 @@ export default class BcRegContacts extends Vue {
     {
       icon: 'mdi-phone',
       key: 'Canada &amp; U.S. Toll Free:',
-      val: '1-877-370-1033',
-      href: 'tel:+1-877-370-1033'
+      val: '1-877-526-1526',
+      href: 'tel:+1-877-526-1526'
     },
     {
       icon: 'mdi-phone',
       key: 'Victoria Office:',
-      val: '250-370-1033',
-      href: 'tel:+1-250-370-1033'
+      val: '250-387-7848',
+      href: 'tel:+1-250-387-7848'
     },
     {
       icon: 'mdi-email',
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -4,7 +4,6 @@
     <!-- conditionally render add component -->
     <v-expand-transition>
       <div id="people-roles-add" v-if="renderOrgPersonForm && isNaN(activeIndex)" class="section-container">
-        <v-divider class="px-4" />
         <org-person
           :currentOrgPerson="currentOrgPerson"
           :activeIndex="activeIndex"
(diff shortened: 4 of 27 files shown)


package-lock.json
```diff
@@ -2976,6 +2976,68 @@
         "tslint": "^5.20.1",
         "webpack": "^4.0.0",
         "yorkie": "^2.0.0"
+      },
```
severinbeauvais, line 2979: You should update the app version too.

src/components/common/BcRegContacts.vue
```diff
@@ -24,14 +24,14 @@ export default class BcRegContacts extends Vue {
...
+      val: '1-877-526-1526',
+      href: 'tel:+1-877-526-1526'
     },
     {
       icon: 'mdi-phone',
       key: 'Victoria Office:',
-      val: '250-370-1033',
-      href: 'tel:+1-250-370-1033'
+      val: '250-387-7848',
+      href: 'tel:+1-250-387-7848'
```
severinbeauvais, line 34: Does this change need to be done in Filings UI, Create UI, Auth UI, Namerequest UI, Registry UI?

severinbeauvais: Create UI already has this change.
PPR UI already has this change.

severinbeauvais: Well actually, Create has this Canada phone number but a different Victoria number 🤷‍♂️ 

```
    <li class="contact-container">
      <v-icon class="contact-icon" size="18px">mdi-phone</v-icon>
      <span class="px-2">Victoria Office:</span>
      <a href="tel:+1-250-952-0568" class="contact-value">250-952-0568</a>
    </li>
```

src/store/getters/state-getters.ts
```diff
@@ -563,7 +563,12 @@ export const hasPeopleAndRolesChanged = (state: StateIF): boolean => {
   let currentOrgPersons = getPeopleAndRoles(state)
   let originalOrgPersons = getEntitySnapshot(state)?.orgPersons
 
-  return (!isSame(currentOrgPersons, originalOrgPersons, ['action']))
+  return (!isSame(currentOrgPersons, originalOrgPersons, ['action', 'confirmNameChange']))
+}
+
+/** Is true when the minimum partners met. */
+export const hasMinimumPartners = (state: StateIF): boolean => {
+  return getPeopleAndRoles(state).filter(party => !party.action?.includes(ActionTypes.REMOVED)).length >= 2
```
severinbeauvais, line 571: Looks fine but could you add a comment explaining what is being computed here? (I understand this today because I'm reviewing this PR, but by next week I may forget why you're excluding parties with the REMOVED action,)

tests/unit/Alteration.spec.ts
```diff
@@ -311,7 +311,7 @@ describe('Alteration component', () => {
     // Validate People And Roles
     expect(store.state.stateModel.peopleAndRoles.orgPeople[0].officer.firstName).toBe('CAMERON')
     expect(store.state.stateModel.peopleAndRoles.orgPeople[0].officer.lastName).toBe('BOWLER')
-    expect(store.state.stateModel.peopleAndRoles.orgPeople[0].roles[0].roleType).toBe('Director')
+    expect(store.state.stateModel.peopleAndRoles.orgPeople[0].role).toBe('director')
```
severinbeauvais, line 314: I think a party can have more than 1 role (ie, array)?

src/components/Edit/ChangeSummary.vue
```diff
@@ -114,6 +115,11 @@ export default class ChangeSummary extends Mixins(
 
     return `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.`
   }
+
+  /** The firm type label. */
+  get firmLabel (): string {
+    return this.getResource.changeData.orgPersonInfo.orgPersonLabel
```
severinbeauvais, line 121: Is this chain of objects guaranteed to be declared (in store state model)?

package-lock.json
cameron-freshworks: Doh

src/components/common/BcRegContacts.vue
cameron-freshworks: Interesting. Designs call for these specifically 🤔

src/store/getters/state-getters.ts
cameron-freshworks: Great point, will add!

tests/unit/Alteration.spec.ts
cameron-freshworks: This is a `/director` response where the property `role` is a singular string value.
Different from our more recently updated `roles` property which is part of our `/parties` response.

cameron-freshworks: Alterations only calls `/directors` not the `/parties`
I think `/parties` is actually somewhat new, possibly introduced at the start of firms? I'm not sure 🤔

src/components/Edit/ChangeSummary.vue
cameron-freshworks: Yes, but in the store resource model.

src/components/common/BcRegContacts.vue
cameron-freshworks: https://projects.invisionapp.com/d/main/#/console/21805570/463592649/preview

cameron-freshworks: What is the number in Create, i could double check with design on this one for sure

cameron-freshworks: If Create UI is different for the Contact # in step 2, it might need to be updated, potentially. 
Yui shared designs for Step 2 contact: https://invis.io/3511UPXPMT2P#/463492960_Step_2G-help_DBA
Which matches the designs linked above for Edit UI

src/store/getters/state-getters.ts
cameron-freshworks: Added!

package-lock.json
cameron-freshworks: Updated!

vysakh-menon-aot (approved):

src/components/common/BcRegContacts.vue
severinbeauvais: I've put the onus on Yui to note any required changes somewhere.

cameron-freshworks: Copy that, sounds good 👍

tests/unit/Alteration.spec.ts
severinbeauvais: Actually, I think `/directors` is the new endpoint. I only see `/parties` used in Filings UI (not Create UI). So this is probably fine. This doesn't even violate the schema because the directors schema doesn't mention roles 🤦‍♂️

severinbeauvais: It does violate the [party](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/party.json) schema though.

We just have to remember that directors are not a subclass of party.

cameron-freshworks: Directors is what we use for Alterations, so its been around for a bit i think. 

But yea, this is role because thats the property returned in the /directors response. I'm not sure if it was updated recently.

cameron-freshworks: > It does violate the [party schema](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/party.json) though.
> 
> We just have to remember that directors are not a subclass of party.

For sure

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -647,6 +713,15 @@ li {
   }
 }
 
+.toggle-business-entry {
+  cursor: pointer;
```
severinbeauvais, line 717: You may have a base class for this already (for next time).

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -647,6 +713,15 @@ li {
...
+  color: $app-blue;
+  text-decoration: underline;
+  float: right;
```
severinbeauvais, line 720: FYI, there's a Vuetify class for this.

My approach: if it's a layout thing, try to do it in the template (as it's easier to manage layouts that way imo). Otherwise declare a class.

This class contains a bunch of styling stuff so I'm totally OK with it.

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -758,6 +833,10 @@ li {
 
 // Overrides for vuetify components (Checkbox alignment, inputField labels/text size/colour)
 ::v-deep {
+  #btn-remove.v-btn.v-btn--disabled {
+    color: $app-red !important;
+    opacity: .4;
+  }
```
severinbeauvais, line 839: In future tickets, we should work towards adding Vuetify overrides like this in the overrides.scss file.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -205,7 +209,7 @@ export default class PeopleAndRoles extends Mixins(CommonMixin) {
       addressCountry: '',
       deliveryInstructions: ''
     },
-    action: null
+    action: []
```
severinbeauvais, line 212: Unless you've changed this, it's not an array.

If it is an array then I'd like to see it pluralized (actions).

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -364,55 +372,50 @@ export default class PeopleAndRoles extends Mixins(CommonMixin) {
...
+      // get ID of person to undo
+      const id = person?.officer?.id
+
+      // get a copy of original person from original IA
+      const thisPerson = (id !== undefined) && cloneDeep(this.originalParties.find(x => x.officer.id === id))
+
+      // safety check
+      if (!thisPerson) {
+        // eslint-disable-next-line no-console
+        console.log('Failed to find original person with id =', id)
```
severinbeauvais, line 388: As this should never happen, it means we screwed up somewhere. Therefore, I recommend changing this to `console.error()`.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -488,7 +553,7 @@ export default class PeopleAndRoles extends Mixins(CommonMixin) {
 
     // just set the action (ie, soft-delete)
     // person will be filtered out on file and pay
-    person.action = ActionTypes.REMOVED
+    person.action = [ActionTypes.REMOVED]
```
severinbeauvais, line 556: Property "action" should be "actions" since it's an array.

severinbeauvais: If this messes up the schema (API side) then I recommend converting action <-> actions in the code that saves/restores the filing. In other words, put a work-around as close to the API as possible so that the rest of the UI code follows our naming scheme.

severinbeauvais (approved):
Really, really good looking HTML and code, best practises and all that 🚀 

Just a couple things to look into but gtg when you're happy with it.

src/enums/partyTypes 2.ts
```diff
@@ -0,0 +1,4 @@
+export enum PartyTypes {
+  PERSON = 'person',
+  ORGANIZATION = 'organization'
+}
```
severinbeauvais, line 4: Why does this filename (and orgPersonTypes) contain a "2" in it?

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Sorry, duplicate conversation. You can resolve one of them.

src/components/common/PeopleAndRoles/OrgPerson.vue
cameron-freshworks: Will 👀

src/components/common/PeopleAndRoles/OrgPerson.vue
cameron-freshworks: Thanks, for sure!
Will make sure to use it where it exclusive additional styling 🦖

src/components/common/PeopleAndRoles/OrgPerson.vue
cameron-freshworks: Definitely. 
The main issue i experience is the colour override. Can't seem to get it to keep it's inherited colour without being explicit.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Definitely an Array now
Will update!

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Sounds good, will do!

src/enums/partyTypes 2.ts
cameron-freshworks: Good question, my editor pulling some tricks i think. Will remove this file as it was most likely duplicated. This sometimes happens.

src/components/common/PeopleAndRoles/OrgPerson.vue
severinbeauvais: > Thanks, for sure! Will make sure to use it where it exclusive additional styling 🦖

Perfect. This is my new rule :)

cameron-freshworks: I like it!

src/components/common/PeopleAndRoles/OrgPerson.vue
severinbeauvais: There are multiple rules for which CSS selector "wins". In this case it's probably a matter of specificity: some other selector is more specific than yours.

This is fine for now. We can improve this kind of thing next time we're here.

src/components/common/BcRegContacts.vue
severinbeauvais: Yui responded:

"_The contact information is contextual. Sometimes we show Victoria number and sometimes not. Sometimes the BC Registries number is different depending on whether we want to direct the calls._

_If the designs coming in have different numbers, that’s a Linda/Sienna question for sure. Though, I doubt even they have a master list of which filing’s helps are supposed to go where._

_**With SPGP, there was no intention to introduce new phone numbers**. So this is worthwhile flagging if you see discrepancies between what code we’re reusing vs the design._"

^^ The emphasis is mine. Recommend you go ahead and double-check what numbers to use here.

src/enums/partyTypes 2.ts
cameron-freshworks: Removed these duplicate files 😵 
Still not sure how this sometimes happens. 100% an editor config

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Updated! Pluralizing a single word has never been so much work and regression testing 😆

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Updated

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -311,7 +311,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin) {
 
   /** Helper method to filter REMOVED orgPersons from PeopleAndRoles list. */
   private filterRemovedOrgPersons (peopleAndRoles: Array<OrgPersonIF>): Array<OrgPersonIF> {
-    return peopleAndRoles.filter(orgPerson => !orgPerson.action?.includes(ActionTypes.REMOVED))
+    return peopleAndRoles.filter(orgPerson => !orgPerson.actions?.includes(ActionTypes.REMOVED))
```
severinbeauvais, line 314: Does `includes` still work here? I see you changed it to `find` in Filings UI.

src/mixins/filing-template-mixin.ts
```diff
@@ -364,16 +359,24 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
-      // if filing and paying, filter out removed entities and omit the 'action' properties
+      // if filing and paying, filter out removed entities and omit the 'actions' properties
       if (!isDraft) {
         // Filter out parties actions
-        parties = parties.filter(x => !x.action.includes(ActionTypes.REMOVED))
+        parties = parties.filter(x => !x.actions?.includes(ActionTypes.REMOVED))
           .map((x) => {
-            const { action, ...rest } = x
+            // Remove null or undefined properties to adhere to schema.
+            // ie taxId is an optional property but schema doesn't accept defined empty properties
```
severinbeauvais, line 368: I shudder when I see `Object.keys` -- it should not be needed.
You might be able to just set known keys (that the API doesn't like) to undefined. Remember that trick about JS object <-> JSON where it doesn't include undefined properties?

src/mixins/filing-template-mixin.ts
```diff
@@ -364,16 +359,24 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
+                delete x.officer[key]
+              }
+            })
+
+            const { actions, ...rest } = x
             return rest
           })
       }
-
+      console.log(parties)
```
severinbeauvais, line 379: Remove when you're done.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Yeah, you have my sympathy -- and my gratitude. It was the right thing to do :)

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: I'm not sure about Filings UI. I think the recent change in Filings UI was related to roleTypes no? 
Was updated to use `some` i believe.

In this context, includes works as we are looking at an array of strings(ActionTypes) built for identifying the changes, or type of changes.

src/mixins/filing-template-mixin.ts
cameron-freshworks: Hmm i'm not sure of the trick. 
In this case it is the `keys` that we want to omit, so if there is another way i'm all for it. 
I do agree, this issue is a result of the API returned a defined key with an empty value, which in turn the Schema does NOT like when returned back to the api. 

Was a pain to find 😵

src/mixins/filing-template-mixin.ts
cameron-freshworks: Thank you! Will do

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: You're right: Filings UI was roleType and this is action.

I think in both cases they're an array of enums. I believe `find(some)` and `includes` works the same for those.

No change needed here.

cameron-freshworks: Totally, i think if were dealing with a plain array (ie strings) i would default to includes, but some/find where we have to check the nested properties of objects in an array etc (ie { roleType:  someRole })

src/mixins/filing-template-mixin.ts
cameron-freshworks: Removed in my current work 👍

src/mixins/filing-template-mixin.ts
severinbeauvais: The trick is, if you set a prop to undefined then it doesn't get included in the JSON.

Earlier today I did this:
```
taxId: this.getRegistration.businessNumber || undefined // can't be null
```

Your code searches for any empty key. But if you knew which keys need to be stripped you could just assign undefined to them.

cameron-freshworks: This UI doesn't assign these values, purely receive from API, displays them and returns them. 
This could change once we use business look however. 

Would the above statement resolve to `taxId:  ''` or remove the property entirely from the object?

cameron-freshworks: We also can't eliminate it entirely because the api sometimes provides it.

cameron-freshworks: Honestly, this could be a schema issue, if it's going to provide empty properties, it should probably pass validation to receive it as such, i'm not sure.

severinbeauvais: So you're saying, you can include values provided by the API but not the others?

Yes, since `''` is falsy, `|| undefined` will assign undefined to that variable. That will remove it entirely from the JSON that JS magically generates from this object.

severinbeauvais: Obviously your code works. I'm trying to understand why it's needed and whether it can be less Object-specific, but maybe it's not worth further discussion today. Just resolve this conversation if you agree :)

severinbeauvais: > We also can't eliminate it entirely because the api sometimes provides it.

That's where the `||` comes in (unfortunately that will also catch legitimate false values and numerical 0). You could check for "null or ``' and then assign undefined (instead of deleting).

At this point I'm not sure which is going to be cleaner, so again, I propose we just leave this alone.

For now. << kidding, haha

cameron-freshworks: haha totally, let's discuss on a call next week!
Thank you :)

PR #259, opened 2022-04-21
Implements Completing Party Shared Component

cameron-freshworks:
*Issue #:* /bcgov/entity#11742

*Description of changes:*
* Imports and implements shared component
* Basic props applied.
* Mocked data input + placeholder event handler *for future dev.

*Note: there is dev code in the PR. It can be removed, however i left it in incase future developers might find it helpful*


vysakh-menon-aot (approved):

severinbeauvais (commented):
OK, so the component is there but not functional. Does this affect any other filing types in this UI, or could this be deployed to Prod as-is with no impact?

cameron-freshworks:
> OK, so the component is there but not functional. Does this affect any other filing types in this UI, or could this be deployed to Prod as-is with no impact?

Yes you betcha, it's currently only visible in the Change of Registration filing, so protected by feature flags.

severinbeauvais (approved):

PR #267, opened 2022-05-03
11932 Fix Proprietor and Partner - UX Improvements 

lewischenstudio:
Issue #: [bcgov/entity/11932](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11932)

*Description of changes:*
UXA Observations
**SP; Individual**

- Future*: The optional BN for the proprietor is not displayed in the proprietor list.
- Future*: The optional BN for the proprietor is not displayed in the proprietor edit form; it should be displayed as read-only field and value.

**SP; DBA**

- [x] The BN entered for Business Unregistered in BC appears as unrendered code.
- [x] When editing a business proprietor, "I confirm the proprietor has legally changed their name and that they remain the same person." should read, "I confirm the proprietor has legally changed their name and that they remain the same business."
- [x] The business proprietor's name text is not vertically aligned well with the icon in the Proprietor list section. See how CreateUI does it as that one, during registration, properly aligns the icon and the text.

**GP**

- [x] When editing a business partner, "I confirm the proprietor has legally changed their name and that they remain the same person." should read, "I confirm the proprietor has legally changed their name and that they remain the same business."
- [x] In the error state when the user removes partners to the point where there are less than 2, the vertical error marker line that appears needs to be left aligned and also extend to the bottom of the white background.
- [x] On the Review and Confirm page, in the Partner Information section, the partner name and their icons are not properly vertically aligned. See how CreateUI for reference.
- [x] On the Review and Confirm page, in the Partner Information section, the information of the partner whose data was not edited is displayed. Only the changed partner should appear here.
- [x] When adding a business partner that is Unregistered in B.C., when using non-BC address, the form throws an error. The partner address can be anywhere and thus should not throw an error.

Future*:

- Business Number isn't hooked from states/BE (ready to display when hooked up)
- Incorporation Number isn't hooked from states/BE (ready to display when hooked up)


src/schemas/out-of-bc-address-schema.ts
```diff
@@ -0,0 +1,33 @@
...
+    required,
+    maxLength: maxLength(40)
+  },
+  addressCountry: {
+    required,
+    // FUTURE: create new validation function isCountry('CA')
+    isCanada: (val: string) => Boolean(val === 'CA')
+  },
+  addressRegion: {
+    maxLength: maxLength(2)
```
lekshmimallika-aot, line 24: Shouldnt this have another rule here that checks region not equals BC?

lewischenstudio: Yes Lekshmi. There is another rule, but that rule is for business inside BC. For this ticket, when adding a business partner that is Unregistered in B.C., we don't want the form to throw the error, so I removed the validation check for BC. Since this is a schema, the best way I could think of was to make another schema to avoid the BC validation. That's why I made this out-of-bc-address-schema ts file.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -488,6 +488,11 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin) {
   .dropdown-action {
     border-left: 1px solid $gray3;
   }
+
+  // Override section container to show error bar
+  .section-container {
+    padding: 1.25rem 1.875rem 2.5rem;
```
severinbeauvais, line 494: `section-container` is a global class. You should not override its padding.

Instead you should update the padding of whatever object is inside the section.

severinbeauvais (changes requested):
I would prefer if you did not change anything in Edit UI at the moment. I am working on a [large cleanup PR](https://github.com/bcgov/business-edit-ui/pull/266) that renames things, moves things, etc.

I think it would be far easier for you to apply your changes after my merge than vice versa.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
lewischenstudio: Correct me if I'm wrong about the scoped style. I thought if the style is scoped, then `.people-roles-content .section-container` only makes changes inside the current component (`ListPeopleAndRoles`), even if `.section-container` is global.

severinbeauvais: Yes, that is how scoped style works.

However, all "section containers" should have the same whitespace definition. What you are trying to do here is add whitespace, but it may belong on an inner element instead.

If you _really_ want to set a new CSS property here, use a different classname. Or use the [Vuetify spacing helpers](https://vuetifyjs.com/en/styles/spacing/).

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -535,5 +539,20 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin) {
...
   }
+
+  // align badge icon with text
+  .badges .v-icon{
+    margin-top: 4px
+  }
+}
+
+// adjust error container padding for last section
+.people-roles-content.last-section {
```
severinbeauvais, line 550: Could you use [last-of-type](https://www.w3schools.com/cssref/sel_last-of-type.asp) to set this padding without needing a conditional classname?

lewischenstudio: Thanks Severin. This last-of-type style is really good.

lewischenstudio:
> I would prefer if you did not change anything in Edit UI at the moment. I am working on a [large cleanup PR](https://github.com/bcgov/business-edit-ui/pull/266) that renames things, moves things, etc.
> 
> I think it would be far easier for you to apply your changes after my merge than vice versa.

For sure. My fixes are small, so it's better and easier to merge my changes after your large cleanup is merged to the main branch first.

src/schemas/out-of-bc-address-schema.ts
lekshmimallika-aot: ok

lekshmimallika-aot (approved):

severinbeauvais (approved):
You have a merge conflict to resolve, but after that I approve this PR to be merged.

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -861,12 +863,6 @@ li {
     color: $gray9;
   }
 
-  // align top of text with checkbox icons
-  .v-input--checkbox .v-label {
-    padding-top: 2px;
-  }
-}
```
severinbeauvais, line 868: This doesn't look right: you deleted 1 open-curly-brace but deleted 2 close-curly-braces.

severinbeauvais (approved):
You have a build error to resolve, but after that I approve this PR to be merged.

src/components/common/PeopleAndRoles/OrgPerson.vue
lewischenstudio: Yep. My bad.

lewischenstudio:
> You have a build error to resolve, but after that I approve this PR to be merged.

Thank you for the merge.

PR #275, opened 2022-05-18
11960 Edit UI - Create new Nature of Business component for conversion filing

lewischenstudio:
*Issue #:* [/bcgov/entity#11960](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11960)

*Description of changes:*
- [x] Nature of business will have a text box instead of NAICS code lookup, 
- [x] Max 300 characters and validation.
- [x] No validation for empty text field
- [x] Shows "(Not Entered)" for empty text
- [x] Review and Confirm shows the equivalent result.

Comment:
- New ticket will be created to handle unexpected NAICS format with missing code for Edit UI (change filing) in the `NatureOfBusiness` component in bcrs shared components repository: [/bcgov/entity#12310](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12310)


lambert-alex (approved):

vysakh-menon-aot (approved):

tests/unit/ConversionNatureOfBusiness.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+
+const vuetify = new Vuetify({})
+
+const businessInformation = {
+  legalType: 'SP',
+  identifier: '',
+  legalName: '',
+  foundingDate: '',
+  hasRestrictions: false,
+  naicsCode: '1234',
```
argush3, line 17: the NAiCS code will always be six digits in the db so we should update this to use a 6 digit code

severinbeauvais: The [schema](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/naics.json) states 5 to 9 digits. Does this change anything? (Other than this test which should be fixed.)

src/components/Conversion/ConversionNOB.vue
```diff
@@ -0,0 +1,145 @@
...
+    return {
+      naicsCode: this.getCurrentNaics.naicsCode,
+      naicsDescription: this.getCurrentNaics.naicsDescription
+    }
+  }
+
+  /** Monitor naics text change */
+  private onChangeNaics (naicsText:string): void {
+    this.naicsRules = naicsText.length > 300 ? ['Maximum 300 characters reached'] : []
+  }
```
severinbeauvais, line 106: First, I don't think you need to do this. I believe the v-text-field can handle this "max length" internally.

Second, the rules can be static. They don't need to be set here.

src/components/Conversion/ConversionNOB.vue
```diff
@@ -0,0 +1,145 @@
...
+        </v-chip>
+      </v-col>
+
+      <v-col cols="12" sm="9">
+        <div v-if="onEditMode">
+          <p class="ma-0">
+            Provide a brief description of the nature of business (e.g., corner grocery store,
+            automotive repair service, landscaping, etc.).
+          </p>
+          <v-text-field
```
severinbeauvais, line 23: Is this supposed to be multi-line? (I haven't seen the designs, sorry.)

If so, consider using v-textarea instead.

src/components/Conversion/ConversionNOB.vue
```diff
@@ -0,0 +1,145 @@
...
+
+  /** Called when user has clicked the Change button. */
+  onChangeClicked (): void {
+    this.onEditMode = true
+  }
+
+  /** Submited when user has clicked the Done button. */
+  onSubmitClicked (): void {
+    if (this.naicsText.length > 300) {
+      this.naicsRules = ['Maximum 300 characters reached']
```
severinbeauvais, line 116: In some cases, the input component(s) would be wrapped in a v-form, and you could simply validate the form here and verify that it's valid. See https://vuetifyjs.com/en/api/v-form/#functions.

In this case, you can actually do the same thing with the text field component. You'll need to declare a ref to it, and then you can call `myComponentRef.validate()` and check that the result is True.

Basically, use the functionality already built in the component to declare validation rules and check if the component is valid.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -305,6 +316,7 @@ import { CommonMixin, SharedMixin, DateMixin, LegalApiMixin, NameRequestMixin }
 import { CorrectionTypes } from '@/enums/'
 import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module/'
 import { cloneDeep } from 'lodash'
+import { ConversionNOB } from '../../Conversion'
```
severinbeauvais, line 319: I think using "local path" is OK (`./`) but anything more should use the normal path specification. In this case, I think `from '@/components/Conversion/'` will work.

tests/unit/ConversionNatureOfBusiness.spec.ts
severinbeauvais: Please use valid-length codes throughout this test suite.

tests/unit/ConversionNatureOfBusiness.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+    wrapper.find('#nob-change-btn').trigger('click')
+    await Vue.nextTick()
+
+    expect(wrapper.vm.$data.onEditMode).toBeTruthy()
+
+    const input = wrapper.find('input[type="text"]')
+    await input.setValue('5678 - cake')
+
+    wrapper.find('#nob-cancel-btn').trigger('click')
+    await Vue.nextTick()
```
severinbeauvais, line 92: I think you can get rid of this nextTick if you await the trigger call.

src/components/common/YourCompany/YourCompany.vue
lewischenstudio: This will raise an error: `"Cannot find module '@components/Conversion/' or its corresponding type declarations."`. I already exported `ConversionNOB` in the file `@/components/Conversion/index.ts`. The same goes to line 313 and 314, it's using `./` not `@`. It happens in many places, and I had to use local path for `.vue` files. If that's a `.ts` file, then `@` won't raise any error.

tests/unit/ConversionNatureOfBusiness.spec.ts
lewischenstudio: Thanks to both of you. I'll use 6 digits for the NAICS code for this test suite.

src/components/Conversion/ConversionNOB.vue
lewischenstudio: I modified ConversionNOB.vue file in my latest commit.

argush3 (approved):

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: What if you try from `'@/components/Conversion/'` (slash between @ and components)?

The difference is:
- `@blah` refers to an imported package
- `@/blah` refers to a folder in /src as per https://github.com/bcgov/business-edit-ui/blob/0dd1590dc82f7f29c83dbcf41c80b807d7144e6b/tsconfig.json#L21

severinbeauvais: The whole point of using an index.ts file to export components is so we can import them from "the folder" instead of from the source file itself.

src/components/Conversion/ConversionNOB.vue
```diff
@@ -0,0 +1,143 @@
...
+@Component({})
+export default class NatureOfBusiness extends Vue {
+  @Getter getCurrentNaics!: NaicsIF
+  @Getter hasNatureOfBusinessChanged!: boolean
+
+  @Action setNaics!: ActionBindingIF
+
+  // local variables
+  private onEditMode = false
+  private maxLength = 300
```
severinbeauvais, line 80: Technically this is constant, so you can use readonly instead of private.

(Also, technically the template can't see private properties, but at the moment Vue allows it.)

src/components/Conversion/ConversionNOB.vue
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+
+  /** Called when user has clicked the Cancel button. */
+  onCancelClicked (): void {
+    this.setNaics(this.originalNaics)
+    this.onEditMode = false
+  }
+}
+</script>
```
severinbeauvais, line 134: Looks really good now.

I think you need to emit an event to inform the parent whether this form is valid.

src/components/Conversion/ConversionNOB.vue
severinbeauvais: Looks good!

src/components/Conversion/ConversionNOB.vue
severinbeauvais: Looks good!

src/components/Conversion/ConversionNOB.vue
severinbeauvais: ^^ This component is invalid if the required data has not been entered, of if the form is still open (edit mode). You can look at other components for examples.

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: Looks good.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -269,7 +269,7 @@
 <script lang="ts">
 import { Component, Emit, Mixins, Prop, Watch } from 'vue-property-decorator'
 import BaseAddress from 'sbc-common-components/src/components/BaseAddress.vue'
-import { OrgPerson } from './'
+import OrgPerson from './OrgPerson.vue'
```
severinbeauvais, line 272: The old code imports from the index file in the current folder.
The new code is more explicit.

I'm OK either way.

lewischenstudio: The old code will case the unit test to fail with `TypeError: Cannot read properties of undefined (reading 'default')`. That's why I changed it, otherwise I wouldn't.

src/components/Conversion/ConversionNOB.vue
lewischenstudio: This form is not required, but I will do an emit to check if the form is still open.

src/components/Conversion/ConversionNOB.vue
lewischenstudio: I'll remove `maxLength` as it's only used in two places. `onEditMode` is a variable, so setting it to readonly will probably not work.

severinbeauvais: I like the idea of a single definition.

Did you try it as readonly?

lewischenstudio: Yes, I did. It will give me an error because it's a constant: `Cannot assign to 'onEditMode' because it is a read-only property.`

severinbeauvais: Ah, oops, the syntax is different:
`readonly maxLength: 300`

src/components/Conversion/ConversionNOB.vue
lewischenstudio: I'm not sure what to display since there is no UX/UI for it. If we click on "Review and Confirm" when the form is still open, it will be closed automatically.

lewischenstudio: However, I missed the "undo" button, and I will add that now.

src/components/Conversion/ConversionNOB.vue
severinbeauvais: Wait a minute.

Max Length is a constant (and can be readonly).

On Edit Mode is a variable that you set True and False. Obviously this cannot be readonly.

Sorry, long day for me.

src/components/Conversion/ConversionNOB.vue
severinbeauvais: It's a general rule that a component is not valid if it is still "open".

Is it possible that the user (staff) enters an empty NAICS description? Would that be invalid?

src/components/Conversion/ConversionNOB.vue
lewischenstudio: I'll remove maxLength as it's only used in two places.

src/components/Conversion/ConversionNOB.vue
lewischenstudio: I added validation for edit mode and more in my latest commit. The staff can leave the text area empty, and it will not raise an error. If the text area is left empty, the summary will display "(Not Entered)".

severinbeauvais (approved):
I didn't get into this super deep but it looks good to me.

Did you verify that the change can be filed with and without a NAICS code/description?

Please add some Test Notes to your ticket to inform Riyaz what scenarios need testing (and that you have tested yourself, obviously).

lewischenstudio:
> e filed with and without a NAICS code/description

I was testing the API for conversion. Apparently, the backend doesn't accept data from conversion filing. There is 400 bad request to the `/fee/SP/CONVERSION` endpoint, and the post call to `/businesses/FM1000018/filings` endpoint is unauthorized with 401. 

image


image

severinbeauvais:
> I was testing the API for conversion. Apparently, the backend doesn't accept data from conversion filing. There is 400 bad request to the `/fee/SP/CONVERSION` endpoint, and the post call to `/businesses/FM1000018/filings` endpoint is unauthorized with 401.

Yes, this is incomplete code that will be implemented in [11954](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11954) and [11889](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11889).

lewischenstudio:
> > I was testing the API for conversion. Apparently, the backend doesn't accept data from conversion filing. There is 400 bad request to the `/fee/SP/CONVERSION` endpoint, and the post call to `/businesses/FM1000018/filings` endpoint is unauthorized with 401.
> 
> Yes, this is incomplete code that will be implemented in [11954](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11954) and [11889](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11889).

In that case, I think the frontend is good to go. The new naics data goes to `naicsDescription`. 

image

image

lewischenstudio:
> I didn't get into this super deep but it looks good to me.
> 
> Did you verify that the change can be filed with and without a NAICS code/description?
> 
> Please add some Test Notes to your ticket to inform Riyaz what scenarios need testing (and that you have tested yourself, obviously).

Thanks. I left a comment in the ticket and mentioned Yui and Riyaz.

PR #276, opened 2022-05-18
11702 edit ui create staff components 

lambert-alex:
*Issue #:* /bcgov/entity#11702

*Description of changes:*

- Added a staff payment component
- Inserted court order component

The certify section was already written for a third party submission. 


src/App.vue
```diff
@@ -557,7 +560,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
           await this.validateReviewCertifyPage()
         } else {
           // Check validity, and if OK then go to summary page.
-          await this.validateCompanyInfoPage()
+          await this.onClickSave()
```
lambert-alex, line 563: onClickSave() is what directs to the pay URL but I'm not sure if I am still supposed to validate the company info page

lambert-alex: This actually breaks the review and confirm button

severinbeauvais: As this change is in App.vue, it will affect all filings (views) in this UI.

I think you should revert this change. What are you trying to do here? (We can have a quick chat if you want.)

src/views/Change.vue
```diff
@@ -55,6 +55,34 @@
...
+              id="court-order"
+              :autoValidation="getAppValidate"
+              :draftCourtOrderNumber="getFileNumber"
+              :hasDraftPlanOfArrangement="getHasPlanOfArrangement"
+              :invalidSection="invalidCourtOrder"
+              @emitCourtNumber="setFileNumber($event)"
+              @emitPoa="setHasPlanOfArrangement($event)"
+              @emitValid="setValidCourtOrder($event)"
+            />
+          </div>
```
severinbeauvais, line 77: It would be cleaner to move this header + blurb + sub-component into a new component (like the other ones).

src/App.vue
```diff
@@ -107,7 +107,7 @@
                 <!-- Corrections still use the basic Fee Summary component -->
                 <aside v-if="isCorrectionFiling">
                   <SbcFeeSummary
-                    :filingData="[...getFilingData]"
+                    :filingData="getFilingData"
```
severinbeauvais, line 110: So there is a significant difference between the old code and the new code.

src/App.vue
```diff
@@ -234,6 +234,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   @Action setSummaryMode!: ActionBindingIF
   @Action setFilingType!: ActionBindingIF
   @Action setFilingId!: ActionBindingIF
+  @Action setFeePrices!: ActionBindingIF
```
severinbeauvais, line 237: Is this used?

src/views/Change.vue
```diff
@@ -55,6 +55,23 @@
...
           :validate="getAppValidate"
         />
+        <template v-if="isRoleStaff">
+          <CourtOrderPoa
+            class="'mt-10"
+            sectionNumber="4."
+            :autoValidation="getAppValidate"
+            :draftCourtOrderNumber="getFileNumber"
+            :hasDraftPlanOfArrangement="getHasPlanOfArrangement"
+            :invalidSection="invalidCourtOrder"
```
severinbeauvais, line 65: You're passing in 4 props but the component only accepts 1 prop.

Actually, I would move these getters into the component (except getAppValidate).

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,49 @@
...
+    <div :class="{ 'invalid-section': invalidCourtOrder }">
+      <CourtOrderPoaShared
+        id="court-order"
+        :autoValidation="getAppValidate"
+        :draftCourtOrderNumber="getFileNumber"
+        :hasDraftPlanOfArrangement="getHasPlanOfArrangement"
+        :invalidSection="invalidCourtOrder"
+        @emitCourtNumber="setFileNumber($event)"
+        @emitPoa="setHasPlanOfArrangement($event)"
+        @emitValid="setValidCourtOrder($event)"
```
severinbeauvais, line 20: These 3 event handlers need to be implemented in this component.

severinbeauvais (commented):
Looks _really_ good so far. You got pushed into the deep end but you learn fast! 😀

src/views/Change.vue
lambert-alex: Okay will do I only need to keep getAppValidate as it relies on user input correct?

src/components/common/CourtOrderPoa.vue
lambert-alex: Will do thanks!

src/views/Change.vue
severinbeauvais: Yes, leave the `getAppValidate` getter since it's at the "same level" as this component. Besides, most of the other components already use it :)

By "same level" I'm thinking of a hierarchical architecture where components only need to know about things at their level of functionality. They don't need to know about what's happening above them or below them (except for formal parameters/properties passed in/out of them).

src/App.vue
severinbeauvais: If staff payment is needed then this possibly belongs in the filing views (change, correction, etc), not here in App.vue. Check what the other filing views are doing for setting fee prices or updating staff payment data.

src/App.vue
severinbeauvais: Obsolete code.

src/App.vue
severinbeauvais: I think you should revert this.

The old code takes all properties in getFilingData, spreads them, and puts them in a new array. The new code just passes in a reference (pointer) to the getFilingData object (not an array).

Create UI passes this prop into SbcFeeSummary indirectly using a local getter (which itself uses the Store getter), which [returns an array](https://github.com/bcgov/business-create-ui/blob/da55ed1951a58cb88b18da309cf44f8a3133b8c4/src/App.vue#L339).

So this prop needs to be changed back to an array.

src/App.vue
severinbeauvais: You may have copied this code from Create UI. It's used for a different reason there (not staff payment).

severinbeauvais: Well actually, I see `setFeePrices` is also used in Edit UI -- in the filing views themselves, not App.vue. If you look at that code, it will hopefully make sense what it's used for (not staff payment).

lambert-alex: No I will drop it.

src/App.vue
lambert-alex: Okay I will revert the changes I made to this

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "3.1.2",
+  "version": "3.1.4",
```
lambert-alex, line 3: Same reason as package-lock change

src/App.vue
```diff
@@ -314,7 +315,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   get feeSummaryConfirmLabel (): string {
     const isNoFee = this.isChangeFiling || this.isConversionFiling
     if (this.isSummaryMode) {
-      return isNoFee ? 'File Now (No Fee)' : 'File and Pay'
+      return isNoFee && !this.getFilingData.priority ? 'File Now (No Fee)' : 'File and Pay'
```
severinbeauvais, line 318: For additional clarity, please wrap the two conditions in parentheses. Thanks!

Explanation: I find ternary expressions like this easy to read/review when they have a simple condition. When there are multiple conditions (or the result is itself an condition) then it's usually better to add parenthese, or break it up into 3 lines, or use if-then-else.

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,66 @@
...
+  }
+})
+export default class CourtOrderPoa extends Vue {
+  /** Prop to provide section number. */
+  @Prop({ default: '' }) readonly sectionNumber: string
+
+  /** Store getters */
+  @Getter getFlagsReviewCertify!: FlagsReviewCertifyIF
+  @Getter getAppValidate!: boolean
+  @Getter hasDraftPlanOfArrangement!: boolean
```
severinbeauvais, line 43: I don't think this Vuex Store getter exists. Did you mean to implement it? Or did you mean to use `getHasPlanOfArrangement` here?

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,66 @@
...
+  private courtOrderNumber: string = null
+  private planOfArrangement: boolean = false
+
+  /** Flags to enable validation */
+  private setFileNumber (courtOrderNumber: string): void {
+    this.courtOrderNumber = courtOrderNumber
+  }
+
+  private setHasPlanOfArrangement (planOfArrangement: boolean): void {
+    this.planOfArrangement = planOfArrangement
```
severinbeauvais, line 58: This saves the event value in this component, but not in the store. Did you mean to call action `setHasPlanOfArrangement`?

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /** True if is a valid court order */
+  get invalidCourtOrder (): boolean {
+    return (this.getAppValidate && !this.getFlagsReviewCertify.isValidCourtOrder)
```
severinbeauvais, line 63: Same comment as above re: action.

src/App.vue
lambert-alex: 👍

src/App.vue
severinbeauvais: resolved

src/components/common/CourtOrderPoa.vue
lambert-alex: I had meant to us `getHasplanOfArrangement` here

src/components/common/CourtOrderPoa.vue
lambert-alex: So saving the event value in the component would be lost after you navigate away from the page and instead using `@Action setHasPlanOfArrangement` would instead change the object in store?

severinbeauvais: Yes.

To be precise, the value in the component will continue to exist until the component is destroyed.

We save data in the Vuex Store so it survives component creation/destruction, but also because it's global and convenient. (The store is lost if the user navigates away from this web app. But the app may support several routes (eg, steps in Create UI) and in that case the app is still loaded and running and everything is fine.)

Notably, the code that builds the filing for saving to the API (in filing-template-mixin.ts) gets the data from the store.

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,59 @@
...
+        If this filing is pursuant to a court order, enter the court order number. If this
+        filing is pursuant to a plan of arrangement, enter the court order number and select
+        Plan of Arrangement.
+      </div>
+    </header>
+    <div :class="{ 'invalid-section': invalidCourtOrder }">
+      <CourtOrderPoaShared
+        id="court-order"
+        :autoValidation="getAppValidate"
+        :hasDraftPlanOfArrangement="hasDraftPlanOfArrangement"
```
severinbeauvais, line 15: Prop is using the wrong getter/value.

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,59 @@
...
+        :invalidSection="invalidCourtOrder"
```
severinbeauvais, line 16: This prop value is not declared anywhere.

In [Alteration.vue](https://github.com/bcgov/business-edit-ui/blob/0dd1590dc82f7f29c83dbcf41c80b807d7144e6b/src/views/Alteration.vue#L205), this value is a local getter that considers a couple of flags in the store. I believe the same should exist here.

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,59 @@
...
+
+  /** Store getters */
+  @Getter getFlagsReviewCertify!: FlagsReviewCertifyIF
+  @Getter getAppValidate!: boolean
+  @Getter getHasPlanOfArrangement!: boolean
+
+  /** Global actions */
+  @Action setValidCourtOrder!: ActionBindingIF
+  @Action setHasPlanOfArrangement!: ActionBindingIF
+  @Action invalidCourtOrder!: ActionBindingIF
```
severinbeauvais, line 48: Should be `setValidCourtOrder`.

severinbeauvais: Oops. That's already declared on line 46. Just delete this invalid action then.

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,59 @@
...
+
+  /** LocalProperties */
+  private courtOrderNumber: string = null
+  private planOfArrangement: boolean = false
```
severinbeauvais, line 52: I think you don't need these any more.

Does VS Code show you grayed-out variables/properties when they're unused? Like this:

image

If not, you're probably missing a VS Code extension. RC me for details.

src/components/common/CourtOrderPoa.vue
```diff
@@ -0,0 +1,59 @@
...
+
+  /** Flags to enable validation */
+  private setFileNumber (courtOrderNumber: string): void {
```
severinbeauvais, line 55: This probably overrides the the `setFileNumber` action that was imported when this component extended file-template-mixin.ts.

If a mixin used by this component already includes a store getter or setter (or any variable or method, really) then you should get a warning when you try to re-declare it here.

In this case, the action from the mixin doesn't have to be redeclared here, and re-declaring it probably overrides the original method. In other words, this method is called (which only sets a local variable, not the value in the store) instead of the correct store action method.

src/components/common/CourtOrderPoa.vue
severinbeauvais: To be clear, I think this component needs both the Store getter and the local getter,

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "business-edit-ui",
-  "version": "3.1.2",
+  "version": "3.1.4",
```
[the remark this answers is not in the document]
severinbeauvais: Haha, good idea but you'd still need to rebase to pick up the other PR's version changes.

And who's to say this PR won't go in first? 😀

src/components/common/CourtOrderPoa.vue
lambert-alex: Is `setValidCourtOder` not being used on line 19 at `@emitValid="setValidCourtOrder($event)"`

severinbeauvais: This was an example from a different project. Sorry for the confusion.

lambert-alex: Right on you are right they are not needed.

lewischenstudio (approved):
This looks really good.

package-lock.json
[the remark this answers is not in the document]
lewischenstudio: You can merge this branch before my branch. You have 11 commits, and it's risky to rebase with too many commits.

src/components/common/CourtOrderPoa.vue
severinbeauvais: obsolete comment

src/components/common/CourtOrderPoa.vue
severinbeauvais: obsolete comment

tests/unit/CourtOrderPoa.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  
+  afterEach(() => {
+    wrapper.destroy()
+  })
+
+  it('renders the components', async () => {
+    expect(wrapper.findComponent(CourtOrderPoa).exists()).toBe(true)
+    expect(wrapper.findComponent(CourtOrderPoaShared).exists()).toBe(true)
+  })
+})
```
severinbeauvais, line 35: This test suite is good enough for now.

Please fix the lint errors identified by the PR CI check. (You may be able to run `npm run lint` to fix them automatically, or at least identify them locally.)

lambert-alex: Linting errors have been fixed. Thanks for all the help!

severinbeauvais:
Alex, please fix the conflicts. Also can you change this PR to Ready for Review?

severinbeauvais (approved):

package-lock.json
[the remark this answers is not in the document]
lambert-alex: Sweet thanks!

PR #277, opened 2022-05-24
12206 Edit UI - handle business addresses with HTTP 404 for SP/GP conversion filings

lewischenstudio:
*Issue #:* [/bcgov/entity#12206](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12206)

*Description of changes:*
- [x] Allow Edit UI display and modify address data for invalid business address with status 404

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "3.1.5",
+  "version": "3.1.6",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/src/components/Conversion/ConversionNOB.vue
+++ b/src/components/Conversion/ConversionNOB.vue
@@ -119,8 +119,8 @@ export default class NatureOfBusiness extends Vue {
   /** The naics data on record for the business. */
   get originalNaics (): NaicsIF {
     return {
-      naicsCode: this.getEntitySnapshot.businessInfo?.naicsCode || '',
-      naicsDescription: this.getEntitySnapshot.businessInfo?.naicsDescription || ''
+      naicsCode: this.getEntitySnapshot.businessInfo?.naicsCode,
+      naicsDescription: this.getEntitySnapshot.businessInfo?.naicsDescription
     }
   }
 
@@ -152,7 +152,7 @@ export default class NatureOfBusiness extends Vue {
   emitUndo (): void {
     let code = this.originalNaics.naicsCode
     let desc = this.originalNaics.naicsDescription
-    this.naicsText = '(Not Entered)'
+    this.naicsText = null
     if (code && desc) {
       this.naicsText = this.hasNatureOfBusinessChanged ? this.naicsText : `${code} - ${desc}`
     } else if (desc) {
--- a/src/views/Conversion.vue
+++ b/src/views/Conversion.vue
@@ -230,7 +230,9 @@ export default class Conversion extends Mixins(
 
     const addresses = await this.fetchAddresses()
       .catch(reason => {
-        if (reason.response?.status === NOT_FOUND) return { businessOffice: null }
+        // error message for business address has the pattern "FMXXXXXXX address not found"
+        if (reason.response?.status === NOT_FOUND &&
+          reason.response?.data.message.includes('address')) return { businessOffice: null }
         throw new Error('Failed to fetch entity addresses')
       })
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-edit-ui",
-  "version": "3.1.5",
+  "version": "3.1.6",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "business-edit-ui",
-      "version": "3.1.5",
+      "version": "3.1.6",
       "dependencies": {
         "@babel/compat-data": "^7.11.0",
         "@bcrs-shared-components/action-chip": "1.0.6",


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "3.1.4",
+  "version": "3.1.6",
```
severinbeauvais, line 3: There is no need to skip a patch version.

severinbeauvais: ... Even if you wait for Alex's PR to be merged first (he sets the version to 3.1.5), you will have to rebase before this PR is ready to merge.

src/services/auth-services.ts
```diff
@@ -63,9 +63,9 @@ export default class AuthServices {
       if (response?.data) {
         return {
           contact: {
-            email: response.data.contacts[0].email,
-            phone: response.data.contacts[0].phone,
-            extension: response.data.contacts[0].phoneExtension
+            email: response.data.contacts[0]?.email || '',
+            phone: response.data.contacts[0]?.phone || '',
+            extension: response.data.contacts[0]?.phoneExtension || ''
```
severinbeauvais, line 68: As I work through my ticket, I've done almost the same thing here!

image

I found that I didn't need to assign empty strings (undefined is OK). What do you think?

I'll let you commit this change but please copy over my comment.

src/views/Conversion.vue
```diff
@@ -171,10 +169,18 @@ export default class Conversion extends Mixins(
         }
 
         // parse firm conversion filing and original business snapshot into store
+        const firmSnapshot = await this.fetchFirmSnapshot()
         await this.parseConversionFiling(changeFiling, firmSnapshot)
       } else {
         // parse business data into store
-        await this.parseEntitySnapshot(firmSnapshot)
+        if (this.isConversionFiling) {
```
severinbeauvais, line 176: This will always be True because we are on the conversion filing page (view).

src/views/Conversion.vue
```diff
@@ -171,10 +169,18 @@ export default class Conversion extends Mixins(
...
+          const conversionSnapshot = await this.fetchConversionSnapshot()
+          const conversionAddress = await this.fetchStoreConversionAddress()
+          await this.parseEntitySnapshot(conversionSnapshot, conversionAddress)
+        } else {
+          const firmSnapshot = await this.fetchFirmSnapshot()
+          await this.parseEntitySnapshot(firmSnapshot)
+        }
```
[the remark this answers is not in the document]
severinbeauvais: And instead of using "404", we use this package:

image

src/mixins/legal-api-mixin.ts
```diff
@@ -166,6 +166,16 @@ export default class LegalApiMixin extends Vue {
...
 
+  /**
+   * Fetches conversion addresses.
+   * @returns a promise to return the data
+   */
+  async fetchConversionAddresses (): Promise<any> {
+    if (!this.getBusinessId) throw new Error('Invalid business id')
+    const url = `businesses/${this.getBusinessId}/addresses`
+    return axios.get(url)
+  }
```
severinbeauvais, line 177: I see nothing conversion-specific in here. It looks like `fetchAddresses()` but with fewer checks.

I think this method is not needed with different code to fetch the conversion snapshot (see my other comment).

src/views/Conversion.vue
```diff
@@ -247,6 +253,50 @@ export default class Conversion extends Mixins(
...
+    }
+  }
+
+  private async fetchStoreConversionAddress ():Promise<any> {
+    let hasBAError = false
+    const response = await Promise.resolve(
+      this.fetchConversionAddresses()
+    ).catch(error => {
+      if (error.response.status === 404 &&
+        error.response.config.url.includes('addresses') &&
```
severinbeauvais, line 287: This is probably redundant because we know (*) that we called the addresses endpoint.

(*) And we shouldn't need to know that anyway, since this layer of code just deals with conversion data. It's the lower layer that deals with whatever the URL is. The principle here is separation of concerns.

src/services/auth-services.ts
lewischenstudio: Sure.

src/mixins/legal-api-mixin.ts
lewischenstudio: `fetchAddresses()` will throw error if axios receives errors, so I had to use another function.

src/views/Conversion.vue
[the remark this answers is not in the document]
lewischenstudio: Thanks for the suggestion. `fetchAddress()` will throw an error if it receives errors from BE, so I had to use a different function and change things around. I used `fetchConversionSnapshot()` instead of `fetchFirmSnapshot()` because it is much more complicated than just making an `axios` get call inside `fetchFirmSnapshot()`. I will use `NOT_FOUND`.

src/views/Conversion.vue
lewischenstudio: Thanks. I'll remove the condition since it is the conversion filing.

src/mixins/legal-api-mixin.ts
severinbeauvais: You can catch the error in Conversion.vue and handle it appropriately there.

src/views/Conversion.vue
lewischenstudio: There are two types of 404. One with `FM... addresses not found` and one with `FM.... not found`. I had to make sure that this is for `FM... addresses not found` with `addresses` in the error message by including `addresses`.

lewischenstudio: image

I need to change `addresses` to `address`.

src/views/Conversion.vue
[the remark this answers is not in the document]
severinbeauvais: My code on line 232 catches the error from `fetchAddresses()`.

It doesn't need to be complicated. Please simplify it. (Feel free to use my code above. If needed, add the extra checks for the error response message.)

src/views/Conversion.vue
severinbeauvais: OK, thanks for the info. I agree that you should keep the error response message check in here.

package.json
lewischenstudio: I'll change it to 3.1.5. I was thinking Alex's PR will be merged before mine.

src/views/Conversion.vue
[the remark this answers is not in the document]
lewischenstudio: image

image


Sorry Severin. I tried your code but this doesn't work because of the Promise reject.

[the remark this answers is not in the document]
severinbeauvais: It's working for me locally.

Try without the space before `reason =>`.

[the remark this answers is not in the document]
lewischenstudio: Thanks Severin. It's the space that causes the error. With this code, things can be much simpler now.

src/views/Conversion.vue
severinbeauvais: Please add a code comment explaining why it's necessary to check the message string. (So a future developer -- or me, after I forget -- will understand.)

src/components/Conversion/ConversionNOB.vue
```diff
@@ -119,8 +119,8 @@ export default class NatureOfBusiness extends Vue {
   /** The naics data on record for the business. */
   get originalNaics (): NaicsIF {
     return {
-      naicsCode: this.getEntitySnapshot.businessInfo?.naicsCode || '',
-      naicsDescription: this.getEntitySnapshot.businessInfo?.naicsDescription || ''
+      naicsCode: this.getEntitySnapshot.businessInfo?.naicsCode,
+      naicsDescription: this.getEntitySnapshot.businessInfo?.naicsDescription
     }
   }
```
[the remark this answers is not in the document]
severinbeauvais: Sometimes it makes sense to have `|| null`. Not sure if that will help here.

(Null in a object that axios converts to JSON is still null. Undefined and the field won't existin in the JSON. FWIW)

src/views/Conversion.vue
lewischenstudio: Alright. I added the comment and included "address" differentiate invalid business address from other 404 status.

src/components/Conversion/ConversionNOB.vue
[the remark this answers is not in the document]
lewischenstudio: I agree. It's working well for this case, so it should be fine without `|| null` for now.

lewischenstudio:
> LGTM
> 
> But you lost your app version changes in the rebase.

Thanks. I've rebased the code and updated the app version.

PR #287, opened 2022-06-01
12169 Prefill Certify component for non-staff user. Updated certify component to latest version.

lambert-alex:
*Issue #:* /bcgov/entity#[12169](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12169)

*Description of changes:*
Commit 1:
- Updated bcrs-shared-components/certify version 1.0.18 => 2.0.4
- Removed CSS from certify common component as it is now covered by updated certify component
- Non-staff users now cannot edit the certify field, and it is prefilled based on the users login name
- App version = 3.1.14

Commit 2-3:
- Resolving merge conflicts

Commit 4:
- Removing code that I added that was unneeded

Commit 5-7:
- Applying requested fixes.

Test notes:
-Alteration, Change and Conversion filings all use the certify component

Tested locally that the changes were not applied to other views that use the CertifySection component, specifically alteration and conversion filings.

### DIFF
--- a/src/components/common/CertifySection.vue
+++ b/src/components/common/CertifySection.vue
@@ -56,10 +56,7 @@ export default class CertifySection extends Mixins(DateMixin, SharedMixin) {
   @Prop({ default: false }) readonly validate: boolean
 
   /** To determine whether user input is enabled. */
-  @Prop({ required: true }) readonly disableEdit!: boolean
-
-  /** To determine if user is staff. */
-  @Prop({ required: true }) readonly isStaff!: boolean
+  @Prop({ default: true }) readonly disableEdit: boolean
 
   /** Called when component is mounted. */
   mounted (): void {
--- a/src/components/common/CertifySection.vue
+++ b/src/components/common/CertifySection.vue
@@ -56,7 +56,7 @@ export default class CertifySection extends Mixins(DateMixin, SharedMixin) {
   @Prop({ default: false }) readonly validate: boolean
 
   /** To determine whether user input is enabled. */
-  @Prop({ default: true }) readonly disableEdit: boolean
+  @Prop({ default: false }) readonly disableEdit: boolean
 
   /** Called when component is mounted. */
   mounted (): void {
--- a/src/components/common/CertifySection.vue
+++ b/src/components/common/CertifySection.vue
@@ -56,7 +56,7 @@ export default class CertifySection extends Mixins(DateMixin, SharedMixin) {
   @Prop({ default: false }) readonly validate: boolean
 
   /** To determine whether user input is enabled. */
-  @Prop({ required: false }) readonly disableEdit: boolean
+  @Prop({ required: true }) readonly disableEdit!: boolean
 
   /** To determine if user is staff. */
   @Prop({ required: true }) readonly isStaff!: boolean
--- a/src/components/common/CertifySection.vue
+++ b/src/components/common/CertifySection.vue
@@ -44,8 +44,6 @@ export default class CertifySection extends Mixins(DateMixin, SharedMixin) {
   @Getter getCertifyState!: CertifyIF
   @Getter getCurrentDate!: string
   @Getter getResource!: ResourceIF
-  @Getter getUserFirstName!: string
-  @Getter getUserLastName!: string
   @Getter isRoleStaff!: boolean
 
   @Action setCertifyState!: ActionBindingIF
@@ -81,11 +79,6 @@ export default class CertifySection extends Mixins(DateMixin, SharedMixin) {
   /** Get the certify resource message */
   get certifyMessage (): string {
     return this.getResource?.certifyClause
-  }
-
-  /** Get if the user is not staff and is doing a change filing */
-  get notStaffChange (): boolean {
-    return (!this.isRoleStaff && this.isChangeFiling)
   }
 
   /** Handler for Valid change event. */
--- a/src/components/common/CertifySection.vue
+++ b/src/components/common/CertifySection.vue
@@ -7,20 +7,23 @@
     </div>
 
     <div :class="{ 'invalid-section': invalidSection }">
-      <CertifyShared
-        :currentDate="getCurrentDate"
-        :certifiedBy="getCertifyState.certifiedBy"
-        :isCertified="getCertifyState.valid"
-        :entityDisplay="readableEntityType"
-        :message="certifyMessage"
-        :isStaff="isRoleStaff"
-        :firstColumn="3"
-        :secondColumn="9"
-        :validate="validate"
-        :invalidSection="invalidSection"
-        @update:certifiedBy="onCertifiedBy($event)"
-        @update:isCertified="onIsCertified($event)"
-      />
+      <v-card flat class="section-container py-6">
+        <CertifyShared
+          :currentDate="getCurrentDate"
+          :certifiedBy="getCertifyState.certifiedBy"
+          :isCertified="getCertifyState.valid"
+          :entityDisplay="readableEntityType"
+          :message="certifyMessage"
+          :isStaff="isRoleStaff"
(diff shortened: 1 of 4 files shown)


src/components/common/CertifySection.vue
```diff
@@ -7,20 +7,23 @@
...
-        :message="certifyMessage"
-        :isStaff="isRoleStaff"
-        :firstColumn="3"
-        :secondColumn="9"
-        :validate="validate"
-        :invalidSection="invalidSection"
-        @update:certifiedBy="onCertifiedBy($event)"
-        @update:isCertified="onIsCertified($event)"
-      />
+      <v-card flat class="section-container py-6">
```
lambert-alex, line 10: Replaces the CSS code removed at the bottom.

src/views/Change.vue
```diff
@@ -54,6 +54,8 @@
           class="mt-10"
           sectionNumber="3."
           :validate="getAppValidate"
+          :invalidSection="isCertifyInvalid"
+          :disableEdit="!isRoleStaff"
```
severinbeauvais, line 58: These 2 props don't match the prop declarations in the CertifySection component.

And while you're here, I had the idea that this page (Change.vue) doesn't need to know if certify is invalid, so you could move that logic into the CertifySection component.

src/views/Change.vue
```diff
@@ -127,6 +132,7 @@ export default class Change extends Mixins(
   @Action setCurrentFees!: ActionBindingIF
   @Action setFeePrices!: ActionBindingIF
   @Action setResource!: ActionBindingIF
+  @Action setCertifyStateValidity!: ActionBindingIF
```
severinbeauvais, line 135: It looks like this isn't used in this component.

src/views/Change.vue
lambert-alex: Does that mean that disableEdit should not be set with isRoleStaff?

lambert-alex: I changed the prop declaration in CertifySection component.  @Prop({ required: false }) readonly disableEdit: boolean =>  @Prop({ required: true }) readonly disableEdit!: boolean. I think this fixes disableEdit. For isCertifyIsInvalid I removed it and let the get invalidSection in the CertifySection component deal with the logic.

severinbeauvais: My comment was that you were passing in 2 props that didn't have `@Prop` declarations in the sub-component. I see you've changed the code since that comment.

No, I think it's correct that edit should be disabled except for staff.

I'm resolving this conversation so I can comment cleanly on your new code.

src/components/common/CertifySection.vue
```diff
@@ -52,6 +55,12 @@ export default class CertifySection extends Mixins(DateMixin, SharedMixin) {
   /** Whether to perform validation. */
   @Prop({ default: false }) readonly validate: boolean
 
+  /** To determine whether user input is enabled. */
+  @Prop({ required: true }) readonly disableEdit!: boolean
+
+  /** To determine if user is staff. */
+  @Prop({ required: true }) readonly isStaff!: boolean
```
severinbeauvais, line 62: You're "accepting" this prop but you're not passing it in from any of the views that use this component.

Option 1: You could pass it in in all 3 views. Pro: this component doesn't need to know about the store. Con: a bit more code.

Option 2: You could just declare the store getter for it here. Pro/con: the opposite of option 1.

On the one hand, I think it's OK if this component imports the store. On the other hand, the 3 views already import isRoleStaff so it would be easy to just pass the prop in.

Another consideration is: does this component need to know that "is staff" means it's editable? That's pretty specific. Maybe this component should just be told whether it's editable, and the caller should determine the logic.

src/views/Change.vue
```diff
@@ -149,6 +153,11 @@ export default class Change extends Mixins(
     return null
   }
 
+  /** Is true when the certify conditions are not met. */
+  get isCertifyInvalid (): boolean {
+    return this.getValidateSteps && !(this.getCertifyState.certifiedBy && this.getCertifyState.valid)
+  }
```
severinbeauvais, line 159: I think you don't need this, since it's already computed in the sub-component, and also THIS component doesn't need to know this state.

src/components/common/CertifySection.vue
lambert-alex: So for option 1 I should pass in a value in each of the three views ie) `:isStaff="isRoleStaff"` and have each view insert `@Getter isRoleStaff boolean` in each of the views and this would sync the isStaff prop with the store?

For option 2 I would just declare the store getter ie) `@Getter isRoleStaff boolean`

severinbeauvais: I think you understand it :)

src/components/common/CertifySection.vue
```diff
@@ -52,6 +55,12 @@ export default class CertifySection extends Mixins(DateMixin, SharedMixin) {
   /** Whether to perform validation. */
   @Prop({ default: false }) readonly validate: boolean
 
+  /** To determine whether user input is enabled. */
+  @Prop({ required: true }) readonly disableEdit!: boolean
```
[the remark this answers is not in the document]
severinbeauvais: If you make this a required prop then you'll need to pass the prop in all 3 views where this component it used.

An alternative is to default this prop to True, in which case edit will be disabled unless the view passes False.

(You could use both required and default but that's probably overkill.)

severinbeauvais:
Please add a short description to the PR title.

src/views/Change.vue
```diff
@@ -219,6 +230,18 @@ export default class Change extends Mixins(
...
+      // do this only if we are not staff
+      if (!this.isRoleStaff) {
+        // pre-populate Certified By name
+        this.setCertifyState(
+          {
+            valid: this.getCertifyState.valid,
+            certifiedBy: `${this.getUserFirstName} ${this.getUserLastName}`
+          }
+        )
+      }
```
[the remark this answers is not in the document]
severinbeauvais: OK, let's not let this hold up this PR.

Could you ask her again at the next stand-up meeting? If this ticket is still open then use it, otherwise we can create a new ticket.

src/views/Change.vue
```diff
@@ -113,6 +114,9 @@ export default class Change extends Mixins(
   @Getter isSummaryMode!: boolean
   @Getter getFilingData!: FilingDataIF
   @Getter getAppValidate!: boolean
+  @Getter getUserFirstName!: string
+  @Getter getUserLastName!: string
+  @Getter getValidateSteps!: boolean
```
severinbeauvais, line 119: Unneeded getter,

severinbeauvais (approved):
Looks good.

One tiny comment.

Did you find any unit tests you could add to? Did you test locally using all types of filings that use the CertifySection component?

src/views/Change.vue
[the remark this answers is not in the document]
lambert-alex: Yes I will she never got back to me, she is probably to busy.

src/components/common/CertifySection.vue
```diff
@@ -52,6 +55,9 @@ export default class CertifySection extends Mixins(DateMixin, SharedMixin) {
   /** Whether to perform validation. */
   @Prop({ default: false }) readonly validate: boolean
 
+  /** To determine whether user input is enabled. */
+  @Prop({ default: false }) readonly disableEdit: boolean
```
severinbeauvais, line 59: This default will be used when this component is used without specifying this prop.

What is the requirement for the Alteration page and the Correction page? To enable edits or disable them?

severinbeauvais: I think you're right here: it needs to be enabled (not disabled) for those other pages because we don't pre-populate the user's name.

Resolving.

PR #306, opened 2022-06-22
Add transactional folio to SP/GP change and correction filings 12522

lambert-alex:
*Issue #:* /bcgov/entity#12522

*Description of changes:*

- Made the Transactional Folio Number component a common component, instead of an alteration component
- Updated all imports for Transactional Folio Number to reflect above change
- Added the Transactional Folio Number component to change filings
- Confirmed Transactional Folio Number worked in staff payment for correction filings
- Added tests confirming components are correctly displayed for the change and correction views
- Verified that the Transactional Folio Number is correctly saved for change and correction filings
- App version  = 3.2.9


tests/unit/Change.spec.ts
```diff
@@ -230,7 +234,24 @@ describe('Change component', () => {
     localVue.use(VueRouter)
     const router = mockRouter.mock()
     await router.push({ name: 'change' })
-    wrapper = shallowMount(Change, { localVue, store, router, vuetify })
+    wrapper = shallowMount(Change, { localVue,
+      store,
+      router,
+      vuetify,
+      data () {
```
lambert-alex, line 241: I can mock behavior for computed value changes by using the following code, this allowed me to use all of the pretesting/post testing methods instead of making new ones for a new computed value. I thought this was neat!

tests/unit/Change.spec.ts
```diff
@@ -242,8 +263,66 @@ describe('Change component', () => {
...
+    expect(wrapper.findComponent(ChangeSummary).exists()).toBe(false)
+    expect(wrapper.findComponent(DocumentsDelivery).exists()).toBe(false)
+    expect(wrapper.findComponent(CompletingParty).exists()).toBe(false)
+    expect(wrapper.findComponent(CertifySection).exists()).toBe(false)
+  })
+
+  it('renders Change view review and confirm page', async () => {
     expect(wrapper.findComponent(Change).exists()).toBe(true)
+    store.state.stateModel.summaryMode = true
+    wrapper.vm.showFee = true
```
lambert-alex, line 283: Changing the computed value triggers a re-rendering so a wait was necessary,

src/views/Change.vue
```diff
@@ -50,23 +50,30 @@
...
+          v-if="showTransactionalFolioNumber"
+          class="mt-10"
+          sectionNumber="2."
+          :validate="getAppValidate"
+        />
+
         <CertifySection
           class="mt-10"
-          sectionNumber="3."
+          :sectionNumber="showTransactionalFolioNumber ? '3.' : '2.'"
```
severinbeauvais, line 62: Nice use of a getter to simplify the template 👍

tests/unit/Change.spec.ts
```diff
@@ -158,7 +162,7 @@ describe('Change component', () => {
       }))
 
     // GET business share classes
-    get.withArgs('businesses/BC1234567/share-classes')
+    get.withArgs('businesses/FM1234567/share-classes')
```
severinbeauvais, line 165: This isn't used by Firm Registration Corrections, so you can remove it. Same with name translations, resolutions, ... (check what the change filing snapshot fetches).

tests/unit/Change.spec.ts
```diff
@@ -48,7 +52,7 @@ describe('Change component', () => {
     'JiIUlDmKZ2ow7GmmDabic8igHnEDYD6sI7OFYnCJhRdgVEHN-_4KUk2YsAVl5XUr6blJKMuYDPeMyNreGTXU7foE4AT-93FwlyTyFzQGddrDv' +
     'c6kkQr7mgJNTtgg87DdYbVGbEtIetyVfvwEF0rU8JH2N-j36XIebo33FU3-gJ5Y5S69EHPqQ37R9H4d8WUrHO-4QzJQih3Yaea820XBplJeo0' +
     'DO3hQoVtPD42j0p3aIy10cnW2g')
```
severinbeauvais, line 54: Check is KEYCLOAK_TOKEN needs to be declared here. I'm thinking, since you're mounting the Change component directly, and it doesn't do anything with the KC token, auth, etc, that this test suite can be simplified.

See App.vue::fetchData() -- this is the code that is NOT run when you mount your page instead of the entire app.

tests/unit/Change.spec.ts
severinbeauvais: Please add a code comment explaining why the wait.

severinbeauvais:
Re: build-check errors

You may need to declare (in `data` block) variable showFee in your unit tests when you mount the component.

severinbeauvais (approved):
OK to merge when:
- build-check errors are resolved
- you have verified that TFN is correctly saved and restored in a change filing

tests/unit/Change.spec.ts
lambert-alex: Okay should I remove the test checks from the xit tests that are marked to be completed in the future?

tests/unit/Change.spec.ts
```diff
@@ -230,7 +234,24 @@ describe('Change component', () => {
...
+        return { showFee: false }
+      },
+      computed: {
+        showFeeSummary: {
+          get (): boolean {
+            return this.showFee
+          },
+          set (val: boolean) {
+            this.showFee = val
+          }
```
lewischenstudio, line 251: I think you can set state with `store.state`... instead of setting the getter explicitly. The same goes to Alteration.spec.ts.

severinbeauvais: True for state properties. In this case, `showFeeSummary` is a computation, so he's trying to work around that by overriding/mocking it.

Ref: https://github.com/bcgov/business-edit-ui/blob/59ab050c2e617017807bc8efa055ca8406e9f2f0/src/store/getters/state-getters.ts#L831

tests/unit/Change.spec.ts
lambert-alex: I think it can be for the tests that I currently have running here. However I think these tests marked for future need them. https://github.com/bcgov/business-edit-ui/blob/59ab050c2e617017807bc8efa055ca8406e9f2f0/tests/unit/Change.spec.ts#L250

severinbeauvais: I think the KC token is only used for the user's roles.

The other tests in this file do not mount App, so fetchData() will not be executed.

PR #331, opened 2022-07-22
12577 update correction title to Register Correction

lambert-alex:
*Issue #:* /bcgov/entity#12577

*Description of changes:*

- Correction header is now marked as Register correction
- App version = 3.4.7


lambert-alex (commented):
When I merge PR 330 then I can fix the merge conflicts.

src/mixins/common-mixin.ts
```diff
@@ -119,9 +119,9 @@ export default class CommonMixin extends Vue {
         return 'Record Conversion'
       case RouteNames.CORRECTION:
         if (this.isCorrectedIncorporationApplication) {
-          return 'Correction - Incorporation Application'
+          return 'Register Correction - Incorporation Application'
```
severinbeauvais, line 122: I'm not certain what the requirement is, but I think it doesn't include the IA correction title. Please confirm with Linda/Dave and Yui. Thanks.

lambert-alex: Okay I'll ask during stand up to confirm

lewischenstudio:
I have a question about this ticket: [link](https://github.com/bcgov/entity/issues/12577). The following checkboxes are not displayed in the description:
- [ ] display only if client error (staff are able to enter the name that they have on the form)
- [ ] for staff error, verify that Filed By uses default user

Correct me if I'm wrong. In our current dev state, we are forcing the Edit UI to use `CorrectionErrorType == "STAFF"`. If that's the case, whether it's an client or staff error, the completing party will display the default user. Either way, the Edit UI doesn't display "Completing Party". Even if it does, it should display "Completing Party" with the input field filled by default user and the input field is disabled. 

If this is correct, then it will be nice to include those in this ticket.

severinbeauvais:
> In our current dev state, we are forcing the Edit UI to use CorrectionErrorType == "STAFF". 

Subtle amplification: we're checking if type is CLIENT. If not (including if it's blank) then we assume STAFF.

src/mixins/common-mixin.ts
severinbeauvais: I heard in grooming today that it's "Register Correction" for all correction types, so this code can be simplified.

severinbeauvais (approved):

lewischenstudio (approved):

PR #333, opened 2022-07-22
12572 SP/GP Correction Edit UI - Update Business Name

lewischenstudio:
*Issue #:* [/bcgov/entity#12572](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12572)

*Description of changes:*
-  reuse the existing component
- only support "new NR" version for name change
- added `isEntityTypeFirm` in name-request-mixin.ts

### DIFF
--- a/src/components/common/YourCompany/YourCompany.vue
+++ b/src/components/common/YourCompany/YourCompany.vue
@@ -344,7 +344,6 @@ export default class YourCompany extends Mixins(
   @Getter getBusinessContact!: ContactPointIF
   @Getter getResource!: ResourceIF
   @Getter isEntityTypeBEN!: boolean
-  @Getter isEntityTypeFirm!: boolean
   @Getter isBenIaCorrectionFiling!: boolean
   @Getter isFirmCorrectionFiling!: boolean
 
@@ -469,7 +468,7 @@ export default class YourCompany extends Mixins(
       return this.getResource.changeData.nameChangeOptions
         .filter(option => option !== CorrectionTypes.CORRECT_NAME_TO_NUMBER)
     }
-
+    if (this.isEntityTypeFirm) return Array(CorrectionTypes.CORRECT_NEW_NR)
     return this.getResource.changeData.nameChangeOptions
   }
   /** show change business for Business Type
--- a/src/components/common/YourCompany/YourCompany.vue
+++ b/src/components/common/YourCompany/YourCompany.vue
@@ -342,8 +342,7 @@ export default class YourCompany extends Mixins(
   @Getter isPremiumAccount!: boolean
   @Getter getEntitySnapshot!: EntitySnapshotIF
   @Getter getBusinessContact!: ContactPointIF
-  @Getter getResource!: ResourceIF
-  @Getter isEntityTypeBEN!: boolean
+  @Getter isEntityTypeFirm!: boolean
   @Getter isBenIaCorrectionFiling!: boolean
   @Getter isFirmCorrectionFiling!: boolean
 
@@ -468,7 +467,6 @@ export default class YourCompany extends Mixins(
       return this.getResource.changeData.nameChangeOptions
         .filter(option => option !== CorrectionTypes.CORRECT_NAME_TO_NUMBER)
     }
-    if (this.isEntityTypeFirm) return Array(CorrectionTypes.CORRECT_NEW_NR)
     return this.getResource.changeData.nameChangeOptions
   }
   /** show change business for Business Type
--- a/src/mixins/name-request-mixin.ts
+++ b/src/mixins/name-request-mixin.ts
@@ -1,15 +1,15 @@
 import { Component, Mixins } from 'vue-property-decorator'
+import { Getter } from 'vuex-class'
 import { NameRequestStates, NameRequestTypes } from '@/enums/'
 import { DateMixin } from '@/mixins/'
 import { LegalServices } from '@/services/'
-import { NrResponseIF } from '@/interfaces/'
-import { Getter } from 'vuex-class'
+import { NrResponseIF, ResourceIF } from '@/interfaces/'
 /**
  * Mixin for processing Name Request objects.
  */
 @Component({})
 export default class NameRequestMixin extends Mixins(DateMixin) {
-  @Getter isEntityTypeFirm!: boolean
+  @Getter getResource!: ResourceIF
 
   /**
    * Fetches an NR and validates it against the applicant's information.
@@ -59,15 +59,15 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
    * @param nr the name request response payload
    * */
   isNrValid (nr: any): boolean {
-    let checkRequestActionCDList = [NameRequestTypes.CHANGE_OF_NAME, NameRequestTypes.CONVERSION]
-    if (this.isEntityTypeFirm) checkRequestActionCDList.push(NameRequestTypes.NEW)
+    let requestActionCDList = [NameRequestTypes.CHANGE_OF_NAME, NameRequestTypes.CONVERSION]
+    if (this.getResource.changeData) requestActionCDList = this.getResource.changeData.nameRequestTypes
     return Boolean(nr &&
       nr.state &&
       nr.expirationDate &&
       !!this.getNrApprovedName(nr) &&
       nr.nrNum &&
       nr.requestTypeCd &&
-      checkRequestActionCDList.includes(nr.request_action_cd)
+      requestActionCDList.includes(nr.request_action_cd)
     )
   }
 
--- a/src/mixins/name-request-mixin.ts
+++ b/src/mixins/name-request-mixin.ts
@@ -1,14 +1,16 @@
 import { Component, Mixins } from 'vue-property-decorator'
-import { CorrectionTypes, NameRequestStates, NameRequestTypes } from '@/enums/'
+import { NameRequestStates, NameRequestTypes } from '@/enums/'
 import { DateMixin } from '@/mixins/'
 import { LegalServices } from '@/services/'
 import { NrResponseIF } from '@/interfaces/'
-
+import { Getter } from 'vuex-class'
 /**
  * Mixin for processing Name Request objects.
  */
 @Component({})
 export default class NameRequestMixin extends Mixins(DateMixin) {
+  @Getter isEntityTypeFirm!: boolean
+
   /**
    * Fetches an NR and validates it against the applicant's information.
    * Throws an error if there is a problem.
@@ -17,12 +19,7 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
    * @param email the applicant's email address
    * @returns the name request response payload
    */
-  async validateNameRequest (
-    nrNumber: string,
-    phone?: string,
-    email?: string,
-    nrType?: CorrectionTypes
-  ): Promise<NrResponseIF> {
+  async validateNameRequest (nrNumber: string, phone?: string, email?: string): Promise<NrResponseIF> {
     const nrResponse: NrResponseIF = await LegalServices.fetchNameRequest(nrNumber).catch(error => {
       this.$root.$emit('invalid-name-request', NameRequestStates.NOT_FOUND)
       throw new Error(`Fetch Name Request error: ${error}`)
@@ -42,14 +39,7 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
 
     // ensure NR is valid
     const isNrValid = this.isNrValid(nrResponse)
-    const isNewNr = nrType === CorrectionTypes.CORRECT_NEW_NR || false
-    if (!isNewNr && (!nrResponse || !isNrValid)) {
-      this.$root.$emit('invalid-name-request', NameRequestStates.INVALID)
-      throw new Error('Invalid Name Request')
-    }
-
-    const isNewNrValid = this.isNewNrValid(nrResponse)
-    if (isNewNr && (!nrResponse || !isNewNrValid)) {
+    if (!nrResponse || !isNrValid) {
       this.$root.$emit('invalid-name-request', NameRequestStates.INVALID)
       throw new Error('Invalid Name Request')
     }
@@ -69,28 +59,15 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
    * @param nr the name request response payload
    * */
   isNrValid (nr: any): boolean {
+    let checkRequestActionCDList = [NameRequestTypes.CHANGE_OF_NAME, NameRequestTypes.CONVERSION]
+    if (this.isEntityTypeFirm) checkRequestActionCDList.push(NameRequestTypes.NEW)
     return Boolean(nr &&
       nr.state &&
       nr.expirationDate &&
       !!this.getNrApprovedName(nr) &&
       nr.nrNum &&
       nr.requestTypeCd &&
-      [NameRequestTypes.CHANGE_OF_NAME, NameRequestTypes.CONVERSION].includes(nr.request_action_cd)
-    )
-  }
-
-  /**
-  * Returns True if the Name Request data is valid for New NR.
-  * @param nr the name request response payload
-  * */
-  isNewNrValid (nr: any): boolean {
-    return Boolean(nr &&
-      nr.state &&
-      nr.expirationDate &&
-      !!this.getNrApprovedName(nr) &&
-      nr.nrNum &&
-      nr.requestTypeCd &&
-      [NameRequestTypes.NEW].includes(nr.request_action_cd)
+      checkRequestActionCDList.includes(nr.request_action_cd)
     )
   }
 
--- a/src/mixins/name-request-mixin.ts
+++ b/src/mixins/name-request-mixin.ts
@@ -1,5 +1,5 @@
 import { Component, Mixins } from 'vue-property-decorator'
-import { NameRequestStates, NameRequestTypes } from '@/enums/'
+import { CorrectionTypes, NameRequestStates, NameRequestTypes } from '@/enums/'
 import { DateMixin } from '@/mixins/'
 import { LegalServices } from '@/services/'
 import { NrResponseIF } from '@/interfaces/'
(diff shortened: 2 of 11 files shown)


severinbeauvais:
> only support "new NR" version for name change

Did you confirm that this is the actual requirement? I think staff should be able to correct company names (same as with BEN IA corrections). In any case, you didn't change this in this PR (it's configurable in the resource files), so please update either way.

severinbeauvais:
You should update the app version.

src/mixins/name-request-mixin.ts
```diff
@@ -17,7 +17,12 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
    * @param email the applicant's email address
    * @returns the name request response payload
    */
-  async validateNameRequest (nrNumber: string, phone?: string, email?: string): Promise<NrResponseIF> {
+  async validateNameRequest (
+    nrNumber: string,
+    phone?: string,
+    email?: string,
+    nrType?: CorrectionTypes
```
severinbeauvais, line 24: This isn't the NR type. It's the type of company name correction being done.

Ref: https://github.com/bcgov/business-edit-ui/blob/main/src/enums/correctionTypes.ts

I think this method/mixin shouldn't care about the type of correction being done. It should only focus on NR. Therefore, I think it would be better if you passed in the information you really care about: is this is correction with a new NR?

severinbeauvais: Actually, is this method called in any scenario other than for a name change NR? It's not clear to me.

src/mixins/name-request-mixin.ts
```diff
@@ -37,7 +42,14 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
 
     // ensure NR is valid
     const isNrValid = this.isNrValid(nrResponse)
-    if (!nrResponse || !isNrValid) {
+    const isNewNr = nrType === CorrectionTypes.CORRECT_NEW_NR
```
severinbeauvais, line 45: OK, I'm pretty confused here. What is this variable exactly? Is it ever false? How does it affect the logic below -- isn't it always a new NR that we're validating?

I'm also confused about the scenario when either of the 2 following blocks is called.

Please add comments or review the variable names or review the scenarios.

src/mixins/name-request-mixin.ts
```diff
@@ -67,6 +79,21 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
...
+  * @param nr the name request response payload
+  * */
+  isNewNrValid (nr: any): boolean {
+    return Boolean(nr &&
+      nr.state &&
+      nr.expirationDate &&
+      !!this.getNrApprovedName(nr) &&
+      nr.nrNum &&
+      nr.requestTypeCd &&
+      [NameRequestTypes.NEW].includes(nr.request_action_cd)
```
severinbeauvais, line 93: I don't think we use a "new" NR for changing a business name. I believe we use a "change" NR, like this:

image

Please confirm with PO.

tests/unit/CorrectNameRequest.spec.ts
```diff
@@ -309,12 +309,13 @@ describe('CorrectNameRequest', () => {
...
             requestTypeCd: 'BC',
-            request_action_cd: 'CNV',
+            request_action_cd: 'NEW',
             entity_type_cd: 'BC',
             applicants: {
               phoneNumber: '250 516 8257',
               emailAddress: 'mock@example.com'
-            }
+            },
+            legalType: 'GP'
```
severinbeauvais, line 318: These unit test changes now only validate a GP/NEW NR and no longer validate a BEN CNV NR. Please add a unit test (or parameterize it) instead.

lewischenstudio:
> > only support "new NR" version for name change
> 
> Did you confirm that this is the actual requirement? I think staff should be able to correct company names (same as with BEN IA corrections). In any case, you didn't change this in this PR (it's configurable in the resource files), so please update either way.

1. This is the [Design UI](https://preview.uxpin.com/363a49b33bdb2a314ccac0f18daad8969e93b337#/pages/150032525/simulate/sitemap). Dev is now unable to use the NR number, so the entire ticket is solve this problem. 
2. The method [`validateNameRequest`](https://github.com/bcgov/business-edit-ui/blob/a0d42e7fe7f782e2c5a3d6932a563df2117ff4ed/src/mixins/name-request-mixin.ts#L20-L65) validates the NR with the backend. Before this PR, it doesn't check for `CorrectionTypes.CORRECT_NEW_NR` while [`onSubmit`](https://github.com/bcgov/business-edit-ui/blob/a0d42e7fe7f782e2c5a3d6932a563df2117ff4ed/src/components/common/YourCompany/CompanyName/CorrectNameRequest.vue#L147-L158) uses `validateNameRequest` only when `this.formType === CorrectionTypes.CORRECT_NEW_NR`.
3. The way I go with this is to preserve the method `validateNameRequest` in its original way by adding an additional, conditional property `nrType`, which takes the value from `CorrectionTypes`. If this method is used other place than `onSubmit`, then it is still good.

severinbeauvais:
> > Did you confirm that this is the actual requirement? I think staff should be able to correct company names (same as with BEN IA corrections). In any case, you didn't change this in this PR (it's configurable in the resource files), so please update either way.
> 
> 1. This is the [Design UI](https://preview.uxpin.com/363a49b33bdb2a314ccac0f18daad8969e93b337#/pages/150032525/simulate/sitemap). Dev is now unable to use the NR number, so the entire ticket is solve this problem.

Yes, and I expressed concern that the design was incorrect, and I asked you to confirm the requirement with the BAs. I still think staff should be able to correct the business name as with BEN IA corrections.

This PR is blocked until this requirement is confirmed.

severinbeauvais:
> 2\. The method [`validateNameRequest`](https://github.com/bcgov/business-edit-ui/blob/a0d42e7fe7f782e2c5a3d6932a563df2117ff4ed/src/mixins/name-request-mixin.ts#L20-L65) validates the NR with the backend. Before this PR, it doesn't check for `CorrectionTypes.CORRECT_NEW_NR` while [`onSubmit`](https://github.com/bcgov/business-edit-ui/blob/a0d42e7fe7f782e2c5a3d6932a563df2117ff4ed/src/components/common/YourCompany/CompanyName/CorrectNameRequest.vue#L147-L158) uses `validateNameRequest` only when `this.formType === CorrectionTypes.CORRECT_NEW_NR`.

I don't understand this response.

severinbeauvais:
> 4\. This ticket is for SP/GP corrections. I think BEN IA corrections have their tickets.

You have changed a method that is also used by BEN IA. You have also removed the BEN IA unit test. You do need to consider the other things you might be impacting in this UI.

severinbeauvais:
> 3\. The way I go with this is to preserve the method `validateNameRequest` in its original way by adding an additional, conditional property `nrType`, which takes the value from `CorrectionTypes`. If this method is used other place than `onSubmit`, then it is still good.

Please see if you can refactor a little bit to clarify this code. It's very confusin and unclear what scenario it handles. (Comments would help.) And again, the mixin should not know what happens outside its scope, such as what kind of form it's called from. Instead, if you can't refactor this out, rename the parameter to tell the method that it should do X instead of Y. I'm hoping you can refactor and greatly simplify the logic here and make it easily comprehensible. Thanks.

severinbeauvais (changes requested):
See comments above, especially:
- requirement for different name change types
- what kind of NR is needed (NEWvs CNV)
- confusing logic
- previous unit test essentially removed

src/mixins/name-request-mixin.ts
```diff
@@ -63,7 +63,8 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
       !!this.getNrApprovedName(nr) &&
       nr.nrNum &&
       nr.requestTypeCd &&
-      [NameRequestTypes.CHANGE_OF_NAME, NameRequestTypes.CONVERSION].includes(nr.request_action_cd)
+      [NameRequestTypes.CHANGE_OF_NAME, NameRequestTypes.CONVERSION, NameRequestTypes.NEW]
```
severinbeauvais, line 66: See other comment re: NEW not a valid NR for this; should be CHANGE OF NAME.

src/mixins/name-request-mixin.ts
severinbeauvais: This PR is blocked until this requirement is confirmed.

src/mixins/name-request-mixin.ts
```diff
@@ -63,7 +63,8 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
...
+        .includes(nr.request_action_cd)
```
severinbeauvais, line 67: Could this code change accidentally allow a "new" NR for a BEN IA correction, which previously didn't allow this type?

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -467,7 +466,7 @@ export default class YourCompany extends Mixins(
       return this.getResource.changeData.nameChangeOptions
         .filter(option => option !== CorrectionTypes.CORRECT_NAME_TO_NUMBER)
     }
-
+    if (this.isEntityTypeFirm) return Array(CorrectionTypes.CORRECT_NEW_NR)
```
severinbeauvais, line 469: You should use the resource files to do this, not special-case code.

We want the components to be as generic as possible. We have resource files to specify a lot of things.

src/mixins/name-request-mixin.ts
severinbeauvais: Latest code has special case for firms so it no longer impacts others.

severinbeauvais:
> > > Did you confirm that this is the actual requirement? I think staff should be able to correct company names (same as with BEN IA corrections). In any case, you didn't change this in this PR (it's configurable in the resource files), so please update either way.
> > 
> > 
> > 
> > 1. This is the [Design UI](https://preview.uxpin.com/363a49b33bdb2a314ccac0f18daad8969e93b337#/pages/150032525/simulate/sitemap). Dev is now unable to use the NR number, so the entire ticket is solve this problem.
> 
> Yes, and I expressed concern that the design was incorrect, and I asked you to confirm the requirement with the PO. I still think staff should be able to correct the business name as with BEN IA corrections.
> 
> This PR is blocked until this requirement is confirmed.

Confirmed by Linda that we are presenting just 1 option for firm corrections: use a new name request.

src/mixins/name-request-mixin.ts
severinbeauvais: Confirmed by Linda that a "new" name request cannot be used in a firm correction (and not a "conversion" NR either). Only a "change of name" NR can be used.

src/mixins/name-request-mixin.ts
severinbeauvais: Confirmed by Linda that a "new" name request cannot be used in a firm correction (and not a "conversion" NR either). Only a "change of name" NR can be used.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -466,7 +467,10 @@ export default class YourCompany extends Mixins(
       return this.getResource.changeData.nameChangeOptions
         .filter(option => option !== CorrectionTypes.CORRECT_NAME_TO_NUMBER)
     }
-    if (this.isEntityTypeFirm) return Array(CorrectionTypes.CORRECT_NEW_NR)
+    if (this.isEntityTypeFirm) {
```
severinbeauvais, line 469: No. We're trying to keep components generic (ie, to know as little as possible about the different variations depending on entity type).

As I have mentioned before, there are resource files that specify the name change options for each entity type.

src/components/common/YourCompany/YourCompany.vue
lewischenstudio: This is now changed.

src/mixins/name-request-mixin.ts
lewischenstudio: It is now using "CHG" for firm filings.

src/mixins/name-request-mixin.ts
```diff
@@ -57,13 +59,15 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
    * @param nr the name request response payload
    * */
   isNrValid (nr: any): boolean {
+    let requestActionCDList = [NameRequestTypes.CHANGE_OF_NAME, NameRequestTypes.CONVERSION]
+    if (this.isEntityTypeFirm) requestActionCDList.pop()
```
severinbeauvais, line 63: So now this mixin needs to know about entity type and has a special case for firms.

It would be better if the acceptable NR types were stored in the resource files (and if they were passed in to this method).

tests/unit/name-request-mixin.spec.ts
```diff
@@ -143,6 +146,43 @@ describe('Name Request Mixin', () => {
...
+    nr.expirationDate = '2021-11-05T07:01:00+00:00'
+    expect(vm.isNrValid(nr)).toBe(false)
+
+    nr.names = [{ state: 'APPROVED', name: 'name' }]
+    expect(vm.isNrValid(nr)).toBe(false)
+
+    nr.nrNum = 'NR 1234567'
+    expect(vm.isNrValid(nr)).toBe(false)
+
+    nr.requestTypeCd = 'CR'
```
severinbeauvais, line 175: Wrong request type code for a firm NR.

severinbeauvais: Sorry, I see you're testing the "false" case here.

However, I don't see where you set a firm request type code in here, so you're not actually testing firm NR validation.

tests/unit/CorrectNameRequest.spec.ts
```diff
@@ -396,7 +396,7 @@ describe('CorrectNameRequest', () => {
             }],
             nrNum: 'NR 1234567',
             requestTypeCd: 'GP',
-            request_action_cd: 'CNV',
+            request_action_cd: 'CHG',
             entity_type_cd: 'GP',
```
severinbeauvais, line 400: I'm thinking lines 398 and 400 should be SP.

I'm also wondering whether the mixin should verify that the type code matches the current filing type...? Or can a GP NR be used for a SP correction? Please check with PO/BA and report back. Thanks.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -466,10 +468,8 @@ export default class YourCompany extends Mixins(
       return this.getResource.changeData.nameChangeOptions
         .filter(option => option !== CorrectionTypes.CORRECT_NAME_TO_NUMBER)
     }
-    if (this.isEntityTypeFirm) {
-      return this.getResource.changeData.nameChangeOptions
-        .filter(option => option === CorrectionTypes.CORRECT_NEW_NR)
-    }
+    if (this.isEntityTypeSP) return SoleProprietorshipResource.changeData.nameChangeOptions
+    if (this.isEntityTypeGP) return GeneralPartnershipResource.changeData.nameChangeOptions
```
severinbeauvais, line 472: Are the resource objects already available in the store? Search for setResource and getResource.

severinbeauvais: Ditto in mixin below.

src/mixins/name-request-mixin.ts
```diff
@@ -60,7 +62,8 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
    * */
   isNrValid (nr: any): boolean {
     let requestActionCDList = [NameRequestTypes.CHANGE_OF_NAME, NameRequestTypes.CONVERSION]
```
severinbeauvais, line 64: It would be good to use the resource files for this case also (which I think is only Benefit Companies but please verify).

src/mixins/name-request-mixin.ts
severinbeauvais: This change now looks good. 👍

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: Looks good.

tests/unit/CorrectNameRequest.spec.ts
severinbeauvais: ^^ outstanding question to answer

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -467,7 +465,7 @@ export default class YourCompany extends Mixins(
       return this.getResource.changeData.nameChangeOptions
         .filter(option => option !== CorrectionTypes.CORRECT_NAME_TO_NUMBER)
     }
-
+    if (this.isEntityTypeFirm) return this.getResource.changeData.nameChangeOptions
     return this.getResource.changeData.nameChangeOptions
```
severinbeauvais, line 469: Lines 468 and 469 return the same value, so you don't need the first one.

lewischenstudio: ok

severinbeauvais: Did you have something else in mind? If so, please add a code comment explaining it.

lewischenstudio: No, I didn't. This makes sense now.

tests/unit/CorrectNameRequest.spec.ts
lewischenstudio: GP NR cannot be used for a SP correction. They must be the same type: GP NR for GP correction and SP NR for SP correction.

severinbeauvais: OK. Can you implement a type check in the NR mixin, or create a new ticket?

src/mixins/name-request-mixin.ts
severinbeauvais: Faiz's PR was ready and merged first, but his work is not complete and he's aware of your changes here, so he'll update this for coop special resolutions in his next PR.

tests/unit/CorrectNameRequest.spec.ts
lewischenstudio: I'll implement the type check in this PR.

lewischenstudio: After I implemented the type check, then I found out that there is already an entity type checker implemented: [here](https://github.com/bcgov/business-edit-ui/blob/5457c482ab6d6fbee20060f14525f29824f7b224/src/components/common/YourCompany/CompanyName/CorrectNameRequest.vue#L159-L176). If you think this is fine, I will revert my code and then push it.

severinbeauvais: Oh, oops.

Yes, I think the original type check is fine.

severinbeauvais (approved):
GTG after you fix the merge conflicts / update the app version.

@lambert-alex Can you review this too, please?

lewischenstudio:
> GTG after you fix the merge conflicts / update the app version.
> 
> @lambert-alex Can you review this too, please?

Thanks for the reviews 🙏

severinbeauvais:
Ref: /bcgov/entity#12572

PR #334, opened 2022-07-26
13031 editable start date for correction filings

lambert-alex:
*Issue #:* /bcgov/entity#13031

*Description of changes:*

 - StartDate implementation is moved out of YourCompany component and is now its own component
 - StartDate component implements DatePicker for correction filings.
 - StartDate now only displays the Month Day, Year for the start date instead of Month Day, Year Time in Pacific Time as per [UXPin](https://preview.uxpin.com/363a49b33bdb2a314ccac0f18daad8969e93b337#/pages/149995942/simulate/sitemap)
 - StartDate component allows for the setting of corrected start dates up to two years in the past and 90 days in the future 
 - Add Unit test suit for StartDate component
 - App version = 3.5.2
 - Build and parse correction filings already build and parse the corrected filing date that is changed/set by StartDate
 - Manually verified that the ComponentsCompanyInfo object still works with correction, alteration, change, and conversion filings


src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+
+  protected onUndoClicked (): void {
+    this.setCorrectionFilingDate(undefined)
+    this.newCorrectedStartDate = undefined
+    this.isCorrected = false
+  }
+
+  @Watch('onEditMode', { immediate: true })
+  protected syncValidity (): void {
+    const isValid = !this.onEditMode
```
lambert-alex, line 205: The only time this section is invalid is when you are picking a date, otherwise you cannot select invalid dates.

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+
+  /** The minimum start date string. */
+  protected get startDateMinStr (): string {
+    return this.dateToYyyyMmDd(this.startDateMin)
+  }
+
+  /** The maximum start date that can be entered (Up to 90 days from today). */
+  protected get startDateMax (): Date {
+    const startDateMax = new Date(this.getCurrentJsDate)
+    startDateMax.setDate(startDateMax.getUTCDate() + 90)
```
lambert-alex, line 138: This method is really similar to the function in [business-create-ui StartDate](https://github.com/bcgov/business-create-ui/blob/5bb6bd107580b0c0ef10e57c3bf2f2bf0160f347/src/components/Registration/StartDate.vue#L71) however I use getUTCDate instead of getDate because startDateMax is in UTC time and using getDate + 90 sets the max start date the current date plus 2 months in the future so I think  it works on the first of the month but that is it, I verified it in dev on a registration filing.

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+  readonly invalidSection: boolean
+
+  protected onEditMode = false
+  protected isCorrected = !!this.getCorrectedFilingDate
+  protected newCorrectedStartDate = undefined // date is "Month Day, Year"
+
+  /** The minimum start date that can be entered (Up to 2 years ago today). */
+  protected get startDateMin (): Date {
+    const startDateMin = new Date(this.getCurrentJsDate)
+    startDateMin.setFullYear(startDateMin.getUTCFullYear() - 2)
```
lambert-alex, line 124: See below comment first
StartDateMin is also in utc time however unlike below it didn't seem to effect the minimum start date however I thought it would be best practice to use geUTCFullYear() anyways

src/store/mutations/mutations-model.ts
```diff
@@ -143,6 +143,10 @@ export const mutateCorrectionInformation = (state: StateIF, correctionInformatio
   state.stateModel.correctionInformation = correctionInformation
 }
 
+export const mutateCorrectionFilingDate = (state: StateIF, correctedFilingDate: string) => {
+  state.stateModel.correctionInformation.correctedFilingDate = correctedFilingDate
+}
+
```
lambert-alex, line 149: If someone could verify that this works locally for them before we merge, it was working for me this morning but now I changed something with my local machine and my all of my unit tests are broken and my setters are broken as well something with wrong with node_modules/http-signature/lib/signer.js:253

src/components/common/YourCompany/StartDate.vue
severinbeauvais: What if you open the date picker and then click close/cancel?

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+        <v-tooltip v-if="!isCorrectionFiling"
+                    top
+                    content-class="top-tooltip"
+                    transition="fade-transition"
+                    nudge-right="3"
+        >
+          <template v-slot:activator="{ on }">
+            <v-icon v-on="on" class="info-icon">mdi-information-outline</v-icon>
+          </template>
+          <span>If the business start date is incorrect, it must be corrected through a correction filing.</span>
```
severinbeauvais, line 30: Isn't THIS a correction filing? Maybe we don't need the tooltip here. Check with Yui.

lambert-alex: Yes but this component is still used in change filings as well

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+          <DatePickerShared
+            title="Start Date"
+            nudge-right="80"
+            nudge-top="15"
+            :minDate="startDateMinStr"
+            :maxDate="startDateMaxStr"
+            @emitDate="onOkClicked($event)"
+          />
+        </div>
+        <div v-if="onEditMode" class="float-right mb-2">
```
severinbeauvais, line 45: Since you have a couple of sequential divs with the same v-if, and to make it easier to fold code, I recomment a template component around these divs and move the v-if to it.

src/components/common/YourCompany/StartDate.vue
severinbeauvais: This looks fine to me. It's the same code as in Create UI so if it's wrong we will now have twice as many opportunities to find the bug and it will simply need to be fixed in both UIs.

src/components/common/YourCompany/StartDate.vue
severinbeauvais: Do we need a new BUG ticket to fix this in Create UI?

src/components/common/YourCompany/StartDate.vue
lambert-alex: Then there won't be a corrected filing date and the correction will keep the business founding date.

src/components/common/YourCompany/StartDate.vue
lambert-alex: Okay will do!

src/components/common/YourCompany/StartDate.vue
lambert-alex: Yes I think we will, I can make one.

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+    return startDateMax
+  }
+
+  /** The maximum start date string. */
+  protected get startDateMaxStr (): string {
+    return this.dateToYyyyMmDd(this.startDateMax)
+  }
+
+  /** The recognition date or business start date-time string. */
+  protected get recognitionDateTime (): string {
```
severinbeauvais, line 148: Some filings display "recognition date and time" and other display "business start date". Does this new component work with the 2 types?

As far as I know:
- BC alteration: recognition date and time
- BC correction: recognition date and time
- firm correction: business start date
- firm change: business start date
- firm conversion: business start date

src/components/common/YourCompany/StartDate.vue
lambert-alex: You are right we do, I will verify with Yui

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+    if (
+      this.isFirmCorrectionFiling ||
+      this.isAlterationFiling ||
+      this.isChangeRegFiling ||
+      this.isFirmConversionFiling
+    ) {
+      if (this.getCorrectedFilingDate) {
+        return this.yyyyMmDdToPacificDate(this.getCorrectedFilingDate, true)
+      } else if (this.getBusinessFoundingDate) {
+        return this.apiToPacificDateLong(this.getBusinessFoundingDate)
```
severinbeauvais, line 163: Why the different display format? Don't we want the long date format always?

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+    if (this.newCorrectedStartDate) {
+      if (this.yyyyMmDdToPacificDate(this.newCorrectedStartDate, true) !==
+        this.apiToPacificDateLong(this.getBusinessFoundingDate)) {
+        this.setCorrectionFilingDate(this.newCorrectedStartDate)
+        this.isCorrected = true
+      } else {
+        this.newCorrectedStartDate = undefined
+      }
+    }
+    this.onEditMode = false
```
severinbeauvais, line 190: So it the user clicks done but they haven't select a date, then the previous date is retained?

src/components/common/YourCompany/StartDate.vue
lambert-alex: This is the same format, the yyMmDdToPacificDate() second argument determines if the date is in long format or not.

tests/unit/StartDate.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    await router.push({ name: 'correction' })
+    await Vue.nextTick()
+    wrapper = mount(StartDate, { store, vuetify, localVue, router })
+    await flushPromises()
+  })
+
+  it('renders the componenet and displays the data correctly', () => {
+    expect(wrapper.find('.pr-2').text()).toBe('Business Start Date')
+    expect(wrapper.find('.info-text').text()).toBe('Unknown')
+  })
```
severinbeauvais, line 34: You need some more tests in here to check the various titles, date formats, and component modes. (Some of these depend on the current entity type.)

src/enums/componentsCompanyInfo.ts
```diff
@@ -8,6 +8,7 @@ export enum ComponentsCompanyInfo {
   'company-name-section',
   'change-business-type',
   'name-translation',
+  'business-start-date',
```
severinbeauvais, line 11: As this object is used for various entity types, please test those other entity types (manually is OK) with respect to this component's validity state.

src/components/common/YourCompany/StartDate.vue
lambert-alex: Yes onOkClicked() sets the newCorrectedStartDate() which onDoneClicked() uses set the corrected filing date as long as it is not the same as the business founding date

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,214 @@
...
+    const isValid = !this.onEditMode
+    this.setValidComponent({ key: 'isValidStartDate', value: isValid })
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+</style>
```
severinbeauvais, line 214: You may need some styling in here to make the title the correct colour (gray9).

src/components/common/YourCompany/StartDate.vue
lambert-alex: I am not actually sure I will have to verify this

src/components/common/YourCompany/StartDate.vue
severinbeauvais: Oh right -- the tooltip is only displayed for "non correction" filings.

Actually it's bad practice to use "not something" here. It would be easier to understand (and extend) if this used "is something" (eg, is alteration) instead.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -207,30 +207,11 @@
 
     <v-divider class="mx-4 my-1" />
 
-    <!-- Business Start Date (changes, conversions and firm corrections only) -->
-    <template v-if="isChangeRegFiling || isFirmConversionFiling || isFirmCorrectionFiling">
```
severinbeauvais, line 211: I don't see this v-if in the new component you created. Isn't it still needed for this section?

src/enums/componentsCompanyInfo.ts
lambert-alex: Okay I will test this with the other entity types that use this locally and then add that they have been confirmed to the PR

tests/unit/StartDate.spec.ts
lambert-alex: Okay I will add more unit tests

src/components/common/YourCompany/StartDate.vue
severinbeauvais: I think I erred above. THIS component is for firm correction/change/conversion filings (business start date). There's a separate section for BC alteration/correction.

severinbeauvais: This does mean some of your code in here doesn't need to handle those other filings.

Or I'm really confused and you're trying to handle both cases in this component. (But if so, it's incomplete).

src/components/common/YourCompany/StartDate.vue
lambert-alex: From the UXpin in shows the title as being black colour #212925 the same as the label but I will verify that with Yui

severinbeauvais (commented):
Don't save these filings, but you can use them to verify things. (I am currently developing using these.)

BC Numbered Company alteration: http://localhost:8080/basePath/BC0871114/alteration/
BC Named Company alteration: http://localhost:8080/basePath/BC0871060/alteration/
BC Numbered Company correction: http://localhost:8080/basePath/BC0871114/correction/?correction-id=130783
BC Named Company correction: http://localhost:8080/basePath/BC0871060/correction/?correction-id=128220
GP correction: http://localhost:8080/basePath/FM1000407/correction/?correction-id=130630
SP correction: http://localhost:8080/basePath/FM1001317/correction/?correction-id=130599
GP change: http://localhost:8080/basePath/FM1000407/change/
GP conversion: http://localhost:8080/basePath/FM1000407/conversion/

src/components/common/YourCompany/StartDate.vue
lambert-alex: Okay I will change that to a verify that it is for something here instead of checking a negation

src/components/common/YourCompany/StartDate.vue
lambert-alex: Okay I will clean up this section after I verify what filing types are using this component

src/store/mutations/mutations-model.ts
lambert-alex: I reinstalled my vscode and that seems to have fixed the issue, not sure what was wrong there.

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Yes I will add those v-ifs in

src/components/common/YourCompany/StartDate.vue
severinbeauvais: So it's never blank?

src/components/common/YourCompany/StartDate.vue
severinbeauvais: Oh, OK. I see the value is different as well. This is fine.

src/enums/componentsCompanyInfo.ts
severinbeauvais: Awesome, perfect!

src/components/common/YourCompany/StartDate.vue
severinbeauvais: Title and label may mean the same thing.

I only noticed that Create UI is explicit about setting the label (left side), and I think the data (right side) is defaulted. They are typically a different gray value. If it works correctly then you're set (and probably don't need this style section).

src/store/mutations/mutations-model.ts
severinbeauvais: 😅

src/components/common/YourCompany/StartDate.vue
lambert-alex: The date displayed should never be blank, however if there is no business founding date or corrected filing date then it will default to Unkown.

severinbeauvais: Should the component should be in error if there is no date?

src/components/common/YourCompany/StartDate.vue
lambert-alex: Ya the component in create UI is displayed as its own component where this is part of the larger YourCompany component I think that is where the difference comes from.

severinbeauvais: OK. You can delete empty style blocks if you don't need them.

src/store/mutations/mutations-model.ts
```diff
@@ -143,6 +143,10 @@ export const mutateCorrectionInformation = (state: StateIF, correctionInformatio
   state.stateModel.correctionInformation = correctionInformation
 }
 
+export const mutateCorrectionFilingDate = (state: StateIF, correctedFilingDate: string) => {
+  state.stateModel.correctionInformation.correctedFilingDate = correctedFilingDate
```
severinbeauvais, line 147: The corrected filing's date should not be updated.

You should be updating the correction start date.

https://github.com/bcgov/business-edit-ui/blob/da53affd2acdf9d0123bbf6fde664ee9b12bb6c8/src/interfaces/store-interfaces/state-interfaces/correction-information-interface.ts#L41

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: Hey, please let the reviewer resolve the conversations they started.

I use open conversations as a bookmark of code to re-review.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -207,31 +207,15 @@
...
-    <v-divider class="mx-4 my-1" />
-
     <!-- Business Start Date (changes, conversions and firm corrections only) -->
     <template v-if="isChangeRegFiling || isFirmConversionFiling || isFirmCorrectionFiling">
-      <section id="business-start-date" class="section-container">
-        <v-row no-gutters>
-          <v-col cols="3" class="pr-2">
-            <label><strong>Business Start Date</strong></label>
-          </v-col>
+      <v-divider class="mx-4 my-1" />
```
severinbeauvais, line 212: Previously this divider was always rendered. Now it will be missing for some filing types. Intentional?

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -433,21 +423,16 @@ export default class YourCompany extends Mixins(
...
   }
 
-  /** The recognition date or business start date-time string. */
-  get recognitionDateTime (): string {
+  /** The recognition date or business start date string. */
+  protected get recognitionDate (): string {
     if (this.isBenIaCorrectionFiling) {
       if (this.getCorrectedFilingDate) {
-        return this.apiToPacificDateTime(this.getCorrectedFilingDate)
+        return this.yyyyMmDdToPacificDate(this.getCorrectedFilingDate) as string
```
severinbeauvais, line 430: No. The method returns a JS Date object. If you just cast it as a string then it uses the browser locale's toString() function to format it.

Instead, look in the date mixin for an appropriate method(s).

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -258,13 +242,14 @@
...
     <div class="section-container" v-if="isAlterationFiling || isBenIaCorrectionFiling">
+      <v-divider class="mx-4 my-1" />
       <v-row no-gutters>
         <v-col cols="3" class="pr-2">
           <label><strong>Recognition Date and Time</strong></label>
         </v-col>
 
         <v-col cols="9">
-          <div class="info-text">{{ recognitionDateTime }}</div>
+          <div class="info-text">{{ recognitionDate }}</div>
```
severinbeauvais, line 252: Well, now the label on line 248 is incorrect. Is it intentional to only show the date, and the not the date-time, for alteration filings?

src/components/common/YourCompany/StartDate.vue
severinbeauvais: What's the bug number?

src/components/common/YourCompany/StartDate.vue
severinbeauvais: Don't resolve the conversations that I started (unless I ask you to :) ).

src/components/common/YourCompany/StartDate.vue
lambert-alex: #13127

severinbeauvais:
The business' start date will need to be saved to the correction filing (and restored from a draft).

I am working through schema issues with Lekshmi and I think I can do this as part of that work.

src/components/common/YourCompany/StartDate.vue
lambert-alex: I am not sure I will ask would that be Mihai or Linda

src/store/mutations/mutations-model.ts
lambert-alex: This actually fixed some errors that I was having with the correction date.

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Will do

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Yes kind of... the shared date picker only returns a date no time so if you are showing a corrected start date there won't be a time, and the UXPin I looked at only show a date no time in the designs

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Yes. The divider I added on line 245 renders for all of the other filing types as well now. This seemed more explicit but is not necessary.

src/components/common/YourCompany/StartDate.vue
```diff
@@ -144,21 +158,15 @@ export default class StartDate extends Mixins(CommonMixin, DateMixin) {
...
-    if (this.isBenIaCorrectionFiling) {
-      if (this.getCorrectedFilingDate) {
-        return this.yyyyMmDdToPacificDate(this.getCorrectedFilingDate, true)
-      }
-    }
+  /** The business date or business start date string. */
+  protected get businessStartDate (): string {
     if (
       this.isFirmCorrectionFiling ||
-      this.isAlterationFiling ||
```
severinbeauvais: Does this component (or YourCompany component) still work for alteration filings?

src/components/common/YourCompany/StartDate.vue
severinbeauvais: Mihai is our BA and can answer business questions.

Dave, as our current PO, can answer business questions.

Yui might be a good choice to ask, as this is a UI interaction, and he can ask BA/PO if it depends on a business requirement.

src/components/common/YourCompany/StartDate.vue
severinbeauvais: This looks OK.

src/components/common/YourCompany/StartDate.vue
severinbeauvais: Did you test all 4 scenarios above?

src/store/mutations/mutations-model.ts
severinbeauvais: I have also done some work on this in my PR as I am updating/fixing the filing template mixin, etc.

Ref: https://github.com/bcgov/business-edit-ui/pull/335/

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: But previously there was a divider before the NOB and now there isn't...?

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: This line of code appears to display the date (no time) for some filing types. Previously this was the date-time. I'm not aware that this should be changed.

Also, the label on line 248 says "date and time" but only date is being displayed.

Do you have a requirement to change the alteration filing and BEN IA correction filing?

I'm thinking you need to revert the changes to the recognitionDate getter below.

src/components/common/YourCompany/StartDate.vue
severinbeauvais: I am still running the old code so just ask if you want a screenshot of any Edit UI filing type.

src/store/mutations/mutations-model.ts
severinbeauvais: I think this needs to be "mutateCorrectionStartDate". It should be a string (yyyy-mm-dd).

You can get the initial value from a correction's Corrected Filing Date but you'll have to convert it to yyyy-mm-dd. There's a future ticket to provided business start date in the business response and then it will be OK.

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: Again, if you want to compare with the "before" filings, ask me.

You may have a specific UI design that changes the "before" look, in which case do your own testing and make sure it's OK.

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Okay I will revert the changes to the recognitionDate getter

src/store/mutations/mutations-model.ts
lambert-alex: Okay I will/have changed it to mutateCorrectionStartDate, do I want to get an initial value from correctedFilingDate? Isn't that the date that the filing was made?

src/components/common/YourCompany/StartDate.vue
lambert-alex: Yes YourCompany component still works for alteration filings, they take the recognitionDateTime instead of businessStartDate

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: This component is used by alteration filings as well as firm correction filings, right? Then it needs to handle both.

src/store/mutations/mutations-model.ts
severinbeauvais: Good question.

For firm corrections -- where you display Business Start/founding Date -- you can get the initial value from the entity snapshot business object.

For other filing types, it would be best to ask at Standup or with a BA/PO. Moving forward, corrections will correct the current business state (snapshot), not the filing being corrected. (I'm thinking we can obsolete correctedFilingDate at some point, but I think the API looks for it atm.) It boils down to: what should the UI display? The date the corrected filing was filed? The business start/founding date? The recognition date and time (and I'm not sure what this is exactly -- ask).

src/components/common/YourCompany/StartDate.vue
severinbeauvais: Perfect.
Do they display date+time or just date atm?

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: Weird. OK, I'll leave this with you.

src/store/mutations/mutations-model.ts
severinbeauvais: The business response for BENs also include a founding date that looks usable:

image

severinbeauvais: But I wouldn't display founding date as Recognition Date and Time without checking with PO/BA/UI and perhaps changing the label.

src/components/common/YourCompany/StartDate.vue
lambert-alex: I changed it back to date time on my local machine, but on GH its just the date. Right now I am having issues with saving and getting the dates it sounds similar to the save issues that you are having with the NOB. I am wondering if the issue you are getting is a cloneDeep issue? That is what I am getting right now

src/components/common/YourCompany/StartDate.vue
lambert-alex: Yes I am now testing with all 5 of the scenarios above.

src/components/common/YourCompany/StartDate.vue
severinbeauvais: I am considering merging my PR ASAP and then you can rebase and see if things are fixed. Thoughts?

tests/unit/StartDate.spec.ts
```diff
@@ -5,32 +5,78 @@ import VueRouter from 'vue-router'
...
-const store = getVuexStore()
+
+let router: any = mockRouter.mock()
+let store: any = getVuexStore()
+
 
 describe('Start Date', () => {
   let wrapper: any
 
+  let correctionInformation = {
```
severinbeauvais, line 24: Use const if you're not reassigning this variable. Ditto on lines 17-18.

(This won't make any difference to how the code runs but it might prevent a developer from reassigning the variables without thinking it through.)

tests/unit/StartDate.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+localVue.use(VueRouter)
+
+let router: any = mockRouter.mock()
+let store: any = getVuexStore()
+
+
+describe('Start Date', () => {
+  let wrapper: any
+
+  let correctionInformation = {
```
severinbeauvais, line 24: Use const if you're not reassigning this variable. Ditto on lines 17-18.

(This won't make any difference to how the code runs but it might prevent a developer from reassigning the variables without thinking it through.)

src/store/mutations/mutations-model.ts
lambert-alex: When I mutate the correction start date in the same form as this the correction start date does not change. Do I have to do a clone deep to it or something?

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: Code looks good now 👍

severinbeauvais:
Please fix your lint errors (you can run `npm run lint` locally) before we merge this PR.

src/store/mutations/mutations-model.ts
severinbeauvais: No need to clone it because it's a primitive type (which is assigned by value), not an object (which is assigned by reference).

I don't think you're saving (storing) the right variable. `correctionInformation.correctedFilingDate` is the filing date of the filing being corrected (an IA or Registration or ChgReg) (although we are getting away from correcting a filing to correcting the current business state).

I think the variable you want to update is `correctionInformation.startDate`. See https://github.com/bcgov/business-edit-ui/blob/de4c94dc420be0d8e387c5384ade030c14056891/src/mixins/filing-template-mixin.ts#L183

severinbeauvais: Resolving this conversation and I will create a more concise (and shorter) one...

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,226 @@
...
+  protected get businessStartDate (): string {
+    if (
+      this.isFirmCorrectionFiling ||
+      this.isFirmChangeFiling ||
+      this.isFirmConversionFiling
+    ) {
+      if (this.getCorrectionStartDate && this.isCorrected) {
+        return this.yyyyMmDdToPacificDate(this.getCorrectionStartDate)
+      } else if (this.getBusinessFoundingDate) {
+        return this.apiToPacificDateLong(this.getBusinessFoundingDate)
```
severinbeauvais, line 171: The first thing I noticed is that the displayed date is wrong, eg:

image

This is because the Founding Date coming from the API is just a date in this case -- you can ignore the time portion and assume the date is already Pacific time zone. But confirm with a BE developer that this is only true for corrections or firms. (I think that, for alterations and/or BENs, this really is a date-time in UTC.)

It would be nice for Founding Date to have a single definition. Maybe the BE guys have some ideas about this and you guys can solve this. Failing that, use code comments to indicate what getBusinessFoundingDate is in this scenario (and also update what it means in recognitionDateTime() in YourCompany.vue).

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,226 @@
...
+  @Action setCorrectionStartDate!: ActionBindingIF
+  @Action setValidComponent!: ActionBindingIF
+
+  /** Whether to show invalid section styling. */
+  @Prop({ default: false })
+  readonly invalidSection: boolean
+
+  protected onEditMode = false
+  protected isCorrected = !!this.getCorrectionStartDate
+  protected newCorrectedStartDate = undefined // date is "Month Day, Year"
```
severinbeauvais, line 133: It's risky to initialize properties as undefined because Vue won't make them reactive (in the template or getters). A better choice here would be
```
protected newCorrectedStartDate = null as string
```

Or you could initialize it to `''` because an empty string is falsy.

severinbeauvais:
I see this template error in the dev console window. After fixing it, maybe things will work better.

image

src/components/common/YourCompany/StartDate.vue
lambert-alex: Okay I will change that to null as string

severinbeauvais: I think this is indeed part of your problem.
1. initialize startDate to null in state-models.ts
2. do not initialize/reset variables to undefined in StartDate.vue

I would add console.logs in various places to verify that variables are being set and read (get) correctly.

src/components/common/YourCompany/StartDate.vue
lambert-alex: I think that the founding date from the API is ahead by one day https://github.com/bcgov/business-edit-ui/blob/85c9b72fed839d86a58298fb072a916e01be449e/src/mixins/date-mixin.ts#L192 from the documentation for the apiToPacificDateLong description.

Okay I will reach out to someone from the BE to find out if we have a way to have a single definition. Failing that I will add some comments clarifying what getBusinessFoundingDate, and recogntionDateTime is.

severinbeauvais: > I think that the founding date from the API is ahead by one day

Yes, it is if you interpret it as a UTC date-time. But it's OK if you interpret it as "just a date". Confirm with BE dev that it is indeed just a date for firms and a UTC date-time for BENs (and coops too maybe?).

lambert-alex: Argus did confirm this. I think we will need a change to the date-mixin as there is no common method for handling this type of date. I think that is a shared component?

severinbeauvais: I believe all UIs have their own Date Mixin. There is also a shared component one for those components when they are used locally (ie, Storybook).

We have `yyyyMmDdToDate`, `yyyyMmDdToPacificDate`, etc. I would just slice the string to drop the time portion (firms only).

severinbeauvais: > Argus did confirm this. I think we will need a change to the date-mixin as there is no common method for handling this type of date. I think that is a shared component?

"2022-06-01T00:00:00+00:00" is actually the day before at 5pm Pacific -- not what is intended here. So I'm not concerned about converting this to a valid Date in a new mixin. Just hack off the time portion, then convert it as a date already in Pacific timezone. Also please leave code comments explaining what's going on here.

-> Straight dates (yyyy-mm-dd) should be interpreted as "a date in Pacific tz".

lambert-alex: Okay I will make these changes and add comments explaining the reasoning behind the changes.

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,232 @@
...
+      this.newCorrectedStartDate = date
+    }
+  }
+
+  protected onDoneClicked (): void {
+    if (this.newCorrectedStartDate) {
+      // For firms only the date is valid
+      if (
+        this.newCorrectedStartDate !==
+        this.getBusinessFoundingDate.slice(0, 10)
```
severinbeauvais, line 196: Does the entire if statement fit on one line now?

lambert-alex: Yes it does.

src/mixins/filing-template-mixin.ts
```diff
@@ -126,7 +126,8 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
         contactPoint: this.getContactPoint,
         nameRequest: this.getNameRequest,
         offices: this.getOfficeAddresses,
-        type: this.getCorrectionErrorType
+        type: this.getCorrectionErrorType,
+        startDate: this.getCorrectionStartDate
```
severinbeauvais, line 130: You should add this conditionally for firm corrections only. (This code already exists on lines 188-190 but please update it as needed.)

lambert-alex: Done!

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -458,19 +454,14 @@ export default class YourCompany extends Mixins(
     return this.toDisplayPhone(this.nrApplicant.phoneNumber)
   }
 
-  /** The recognition date or business start date-time string. */
-  get recognitionDateTime (): string {
+  /** The recognition date or business start date string. */
+  protected get recognitionDateTime (): string {
```
severinbeauvais, line 458: _Probably_ getters shouldn't be called externally, but there's no harm in them being public (default when no modifier is present), since they only expose data, not change it.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -271,17 +256,21 @@
     </template>
 
     <!-- Recognition Date and Time (alterations and BEN corrections only) -->
-    <div class="section-container" v-if="isAlterationFiling || isBenIaCorrectionFiling">
+    <template v-if="isAlterationFiling || isBenIaCorrectionFiling">
+
+    <v-divider class="mx-4 my-1" />
```
severinbeauvais, line 261: This code block should now be indented.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -458,19 +454,14 @@ export default class YourCompany extends Mixins(
     return this.toDisplayPhone(this.nrApplicant.phoneNumber)
   }
 
-  /** The recognition date or business start date-time string. */
-  get recognitionDateTime (): string {
+  /** The recognition date or business start date string. */
+  protected get recognitionDateTime (): string {
     if (this.isBenIaCorrectionFiling) {
       if (this.getCorrectedFilingDate) {
```
severinbeauvais, line 460: I'm thinking that BEN IA corrections should also display the business founding date (not the date of the original IA). Please confirm with BA/PO and change accordingly.

src/mixins/filing-template-mixin.ts
```diff
@@ -129,7 +130,8 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
         contactPoint: this.getContactPoint,
         nameRequest: this.getNameRequest,
         offices: this.getOfficeAddresses,
-        type: this.getCorrectionErrorType
+        type: this.getCorrectionErrorType,
+        ...(this.isEntityTypeFirm && { startDate: this.getCorrectionStartDate })
```
severinbeauvais, line 134: I think this isn't needed since you have the code on lines 191-193 already.

src/mixins/filing-template-mixin.ts
```diff
@@ -186,7 +188,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
       }
 
       // *** FUTURE: update logic as needed (eg, only save if changed)
-      if (filing.correction.startDate) {
+      if (filing.correction.startDate && this.isEntityTypeFirm) {
```
severinbeauvais, line 191: I think this doesn't apply to all firm filings. You might need to identify the filing types this applies to. See also line 184.

src/mixins/filing-template-mixin.ts
```diff
@@ -496,6 +498,14 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
     if (!filing.correction.type) filing.correction.type = CorrectionErrorTypes.STAFF
 
     // store Correction Information
+    if (
+      this.isCorrectedRegistration ||
+      this.isCorrectedChangeReg ||
+      this.isCorrectionFiling
```
severinbeauvais, line 504: These should match the conditions when you save (build) the filing above.

You don't need to check if this is a correction filing -- this method is specific to corrections already.

src/mixins/filing-template-mixin.ts
lambert-alex: As this is part of the buildCorrectionFiling method I thought I wouldn't have to filter out what type of filing is needed, and I would only need filter based on the type of entity?

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Okay I am waiting to hear back from Mihai.

src/mixins/filing-template-mixin.ts
lambert-alex: Is this a cleaner implementation that I could use instead of the code on lines 191-193?

severinbeauvais: It's OK, but ...

First, does this work? It looks like you are destructuring a `()` block...?

Second, you may need to add a condition for whether Start Date has changed (so you don't save it if not).

Third, I think it is actually cleaner to conditionally add something (as below) instead of conditionally adding a key or undefined.

src/mixins/filing-template-mixin.ts
severinbeauvais: Oh I see.

Yes, you're right.

And actually, you don't have to check if this is a firm because line 184 already does that (corrected Registrations or corrected Change of Registrations are firm-only). At some point (future) I may rename those getters to denote that, eg, isCorrectedFirmRegistration.

src/components/common/YourCompany/YourCompany.vue
severinbeauvais: FWIW, it will still be called Recognition Date and Time for BEN filings.

src/mixins/filing-template-mixin.ts
lambert-alex: Yes it uses the spread operator, so that the so the object `{ startDate: value}` is returned only if the condition is true and then it is spread into the object.
Okay I will change this to conditionally add it as it is done below.

src/mixins/filing-template-mixin.ts
lambert-alex: Right okay that makes sense

src/mixins/filing-template-mixin.ts
lambert-alex: Done

src/mixins/filing-template-mixin.ts
lambert-alex: Done

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Done

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Done

severinbeauvais: The v-divider should be indented too.

severinbeauvais: And you don't need the blank line before the divider (line 260) but it would be nice to have a blank line after it (for readability).

src/mixins/filing-template-mixin.ts
severinbeauvais: Please delete this line.

Also please delete the comment on line 190 (I'm assuming the logic is correct there).

src/components/common/YourCompany/YourCompany.vue
lambert-alex: From stand up on Friday I made these changes

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Done

lambert-alex: Done

src/mixins/filing-template-mixin.ts
lambert-alex: Done

lambert-alex: Done

src/mixins/filing-template-mixin.ts
```diff
@@ -497,9 +500,15 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
...
     // store Correction Information
-    if (this.isCorrectedRegistration || this.isCorrectedChangeReg) {
+    if (
+      this.isCorrectedRegistration ||
+      this.isCorrectedChangeReg ||
+      this.isCorrectionFiling
+    ) {
       // Ensures startDate isn't undefined, otherwise its getter is not reactive
-      filing.correction.startDate = null
+      if (!filing.correction.startDate) {
```
lambert-alex, line 509: When loading the correction from a save filing.correction.startDate would already be set and this line was setting it to null, erasing the saved change.

src/mixins/filing-template-mixin.ts
```diff
@@ -10,7 +10,6 @@ import { ActionTypes, AssociationTypes, CorrectionErrorTypes, EffectOfOrders, Fi
   RoleTypes } from '@/enums/'
 import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module/'
 import { StaffPaymentOptions } from '@bcrs-shared-components/enums/'
-import { StartDate } from '@/components/common'
```
lambert-alex: Unused import

src/mixins/filing-template-mixin.ts
```diff
@@ -10,6 +10,7 @@ import { ActionTypes, AssociationTypes, CorrectionErrorTypes, EffectOfOrders, Fi
   RoleTypes } from '@/enums/'
 import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module/'
 import { StaffPaymentOptions } from '@bcrs-shared-components/enums/'
+import { StartDate } from '@/components/common'
```
severinbeauvais: I believe this mixin doesn't need this visual component.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -209,43 +209,28 @@
...
-    <v-divider class="mx-4 my-1" />
-
     <!--- Association Type (coop only) -->
     <div v-if="isEntityTypeCP"
         id="association-type-section"
         class="section-container"
         :class="{'invalid-section': invalidAssociationTypeSection}"
     >
+      <v-divider class="mx-4 my-1" />
         <AssociationType
```
severinbeauvais, line 219: This code block should line up with the divider above.

src/mixins/filing-template-mixin.ts
```diff
@@ -179,15 +181,17 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
 
     // add in Registration / Change of Registration data
     // *** FUTURE: change this to "if firm correction"
-    if (this.isCorrectedRegistration || this.isCorrectedChangeReg) {
+    if (
+      this.isCorrectedRegistration ||
+      this.isCorrectedChangeReg ||
+      this.isCorrectionFiling
```
severinbeauvais, line 187: I think this is incorrect, since it will be True for BEN IA corrections (which don't have the data below).

If one of isCorrectedRegistration or isCorrectedChangeReg isn't True in your testing, it would be good to figure out why and fix that.

src/mixins/filing-template-mixin.ts
severinbeauvais: I think this still isn't right. See the comment I made on line 187.

src/store/getters/state-getters.ts
```diff
@@ -171,7 +171,7 @@ export const getCorrectionErrorType = (state: StateIF): CorrectionErrorTypes =>
 }
 
 /** The correction (business) start date (YYYY-MM-DD). */
-export const getCorrectionStartDate = (state: StateIF): string => {
+export const getCorrectedStartDate = (state: StateIF): string => {
```
severinbeauvais, line 174: I think this is incorrect.

The corrected start date is `correctionInformation.correctedFilingDate`. This is the date of the original filing (the filing being corrected).

This getter is supposed to return the start date in this correction filing (the filing doing the correcting), if the user corrected it.

Maybe it would make more sense to name this "business start date"? (It's still called "start date" the schema though.)

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  /** The business date or business start date string. */
+  protected get businessStartDate (): string {
+    if (
+      this.isFirmCorrectionFiling ||
+      this.isFirmChangeFiling ||
+      this.isFirmConversionFiling
+    ) {
+      if (this.getCorrectedStartDate) {
```
severinbeauvais, line 169: Would it make sense to use hasBusinessStartDateChanged here?

lambert-alex: Yes I changed it.

src/mixins/filing-template-mixin.ts
lambert-alex: Yes it has been removed already

src/store/getters/state-getters.ts
lambert-alex: The founding date is the original business start date I think we should have the getter indicate that this is the corrected start date so someone doesn't accidentally use this getter thinking it is "business start date" maybe "corrected business start date"?

severinbeauvais: Yes, the founding date is the original business start date. Agreed, this isn't that value. (And this is also not the corrected filing's date.) So we should not confuse things by using "corrected" in the name.

The naming convention for correction fields is "correctionXxx", so I rather like keeping this getter named getCorrectionStartDate, but I'm open to other ideas. What do you think?

lambert-alex: That makes sense to keep it consistent with the current naming conventions. I'll switch it back to getCorrectionStartDate.

src/components/common/YourCompany/YourCompany.vue
lambert-alex: Done.

src/mixins/filing-template-mixin.ts
lambert-alex: Fixed the issue was in checking for the filing type before I set the filing type.

src/mixins/filing-template-mixin.ts
lambert-alex: Okay I found the issue I was checking for the correction filing type before setting the correction filing type. I fixed it now.

src/store/getters/state-getters.ts
lambert-alex: Done.

src/components/common/YourCompany/StartDate.vue
```diff
@@ -0,0 +1,232 @@
...
+  @Action setCorrectionStartDate!: ActionBindingIF
+  @Action setValidComponent!: ActionBindingIF
+
+  /** Whether to show invalid section styling. */
+  @Prop({ default: false })
+  readonly invalidSection: boolean
+
+  protected dropdown = false
+  protected onEditMode = false
+  protected isCorrected = !!this.getCorrectionStartDate
```
severinbeauvais, line 134: This could use the getter, too.

I'm a bit concerned that the initial value will be incorrect (if this variable is initialized before the draft correction is loaded). Does it matter, since it's set in `businessStartDate` below, which is reactive? I dunno, probably needs some testing with it initially True and False.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -362,6 +352,7 @@ export default class YourCompany extends Mixins(
   @Getter isBenIaCorrectionFiling!: boolean
   @Getter isFirmCorrectionFiling!: boolean
   @Getter getEntityType!: CorpTypeCd
+  @Getter getCorrectionStartDate!: string
```
severinbeauvais, line 355: Looks like this isn't used in this component.

src/mixins/filing-template-mixin.ts
severinbeauvais: Good find! Yes, I see the change in parseCorrectionFiling(). 👍

severinbeauvais (approved):
LGTM. Please poke @lewischenstudio for a second review and approval.

src/components/common/YourCompany/YourCompany.vue
lambert-alex: I'll remove this

src/components/common/YourCompany/StartDate.vue
lambert-alex: I'll add this, I have just finished testing it, it does not matter since it is set in businessStartDate below.

severinbeauvais: If it is True initially, does it get set to False below?

lambert-alex: It does not. I can make it do that though by setting it the same way I set it to true below.

severinbeauvais: Sounds good.

Also, if you don't want to make any assumptions about the initial value, you can set it to null. It's kind of a 3rd value for booleans. You do have to set a value (and not undefined) or it won't be reactive and the v-ifs in the template won't see any changes to it.

It complicated and simple eh?

lambert-alex: Ya it really is, I like the setting it to null, the less assumptions the better probably.

lewischenstudio (approved):
LGTM

PR #336, opened 2022-07-27
Feature/13063

shabeeb-aot:
*Issue #:* /bcgov/entity#13063

*Description of changes:*
* tooltip for coop business type and mailing address

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "3.5.4",
+  "version": "3.5.5",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "3.5.5",
+  "version": "3.5.6",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -305,7 +305,7 @@ export default class ChangeBusinessType extends Mixins(CommonMixin, SharedMixin)
     return this.getResource.changeData?.typeChangeInfo
   }
 
-  /** v-if for tooltip (shoe only firm and coop)  */
+  /** only applies to firm and coop filings  */
   get showChangeInfoTooltip ():boolean {
     return this.isEntityTypeFirm || this.isEntityTypeCP
   }
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -22,7 +22,6 @@
                     content-class="top-tooltip"
                     transition="fade-transition"
                     nudge-right="3"
-                    class="tooltip-info"
         >
           <template v-slot:activator="{ on }">
             <v-icon v-on="on" class="info-icon">mdi-information-outline</v-icon>
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -305,9 +305,9 @@ export default class ChangeBusinessType extends Mixins(CommonMixin, SharedMixin)
     return this.getResource.changeData?.typeChangeInfo
   }
 
-  /** v-if for tooltip (change firm and coop)  */
+  /** v-if for tooltip (shoe only firm and coop)  */
   get showChangeInfoTooltip ():boolean {
-    return this.isChangeRegFiling || this.isEntityTypeFirm || this.isEntityTypeCP
+    return this.isEntityTypeFirm || this.isEntityTypeCP
   }
 
   /** Reset company type values to original. */
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -18,7 +18,7 @@
         </span>
 
         <!-- Firm info tooltip -->
-        <v-tooltip v-if="isChangeRegFiling || isEntityTypeFirm"
+        <v-tooltip v-if="isChangeRegFiling || isEntityTypeFirm || isEntityTypeCP"
                     top
                     content-class="top-tooltip"
                     transition="fade-transition"
@@ -242,6 +242,7 @@ export default class ChangeBusinessType extends Mixins(CommonMixin, SharedMixin)
   @Getter isConflictingLegalType!: boolean
   @Getter isEntityTypeBC!: boolean
   @Getter isEntityTypeFirm!: boolean
+  @Getter isEntityTypeCP!: boolean
 
   @Action setEntityType!: ActionBindingIF
 
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -18,7 +18,7 @@
         </span>
 
         <!-- Firm info tooltip -->
-        <v-tooltip v-if="isChangeRegFiling || isEntityTypeFirm || isEntityTypeCP"
+        <v-tooltip v-if="showChangeInfoTooltip"
                     top
                     content-class="top-tooltip"
                     transition="fade-transition"
@@ -302,6 +302,11 @@ export default class ChangeBusinessType extends Mixins(CommonMixin, SharedMixin)
   /** Type change helper information */
   get typeChangeInfo (): string {
     return this.getResource.changeData?.typeChangeInfo
+  }
+
+  /** v-if for tooltip (change firm and coop)  */
+  get showChangeInfoTooltip ():boolean {
+    return this.isChangeRegFiling || this.isEntityTypeFirm || this.isEntityTypeCP
   }
 
   /** Reset company type values to original. */
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -16,13 +16,13 @@
         <span class="info-text" :class="{ 'has-conflict': isConflictingLegalType && isNewName}">
           {{getCorpTypeDescription(getEntityType)}}
         </span>
-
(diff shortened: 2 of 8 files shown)


severinbeauvais:
I will review this after the other reviewers are done.

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -100,6 +100,21 @@
...
+                    top
+                    content-class="top-tooltip"
+                    transition="fade-transition"
+                    nudge-right="3"
+        >
+          <template v-slot:activator="{ on }">
+            <v-icon v-on="on" class="info-icon">mdi-information-outline</v-icon>
+          </template>
+          <span>{{ addressChangeInfo }}</span>
+        </v-tooltip>
```
lewischenstudio, line 115: I'm leaning towards having this `<v-tooltip>` inside `<div>` in line 92, so that you don't have to specify another class. But there might be some small margin change to the right.

src/interfaces/resource-interfaces/resource-interface.ts
```diff
@@ -9,6 +9,7 @@ export interface ResourceIF {
   nameRequestType: NameRequestEntityTypes
   addressLabel: string
   filingData: FilingDataIF
+  addressChangeInfo?: string
```
lewischenstudio, line 12: Can you move `addressChangeInfo` under `changeData`? I think it's more organized this way.

lewischenstudio:
I think we need unit test cases for ChangeBusinessType.spec and OfficeAddresses.spec since we are considering speicialResolution as one of the entity types.

src/components/common/YourCompany/OfficeAddresses.vue
lewischenstudio: I think we can stick with this way as it's more clear. No need to change anything.

src/interfaces/resource-interfaces/resource-interface.ts
shabeeb-aot: updated

shabeeb-aot:
added unit test case for both. There was no unit test case ChangeBusinessType.spec, created a new file and added some test

src/interfaces/resource-interfaces/resource-interface.ts
lewischenstudio: This is resolved 👍

lewischenstudio (commented):
LGTM. Waiting for more reviews.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "3.5.4",
+  "version": "3.5.5",
```
severinbeauvais, line 3: This needs to be updated (sorry).

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -303,6 +304,11 @@ export default class ChangeBusinessType extends Mixins(CommonMixin, SharedMixin)
     return this.getResource.changeData?.typeChangeInfo
   }
 
+  /** v-if for tooltip (change firm and coop)  */
+  get showChangeInfoTooltip ():boolean {
+    return this.isChangeRegFiling || this.isEntityTypeFirm || this.isEntityTypeCP
```
severinbeauvais, line 309: The first 2 conditions overlap: a ChangeReg filing can only be done by a firm.

I would keep the second condition since firms can also do Correction and Conversion filings.

src/components/common/YourCompany/OfficeAddresses.vue
```diff
@@ -100,6 +100,21 @@
               </v-btn>
             </div>
           </v-col>
+          <v-col cols="1" v-else-if="isSpecialResolutionFiling">
+            <div class="d-flex justify-end align-end align-sm-start tooltip-info">
```
severinbeauvais, line 104: How about moving this class to the tooltip itself as you did in ChangeBusinessType.vue?

src/interfaces/resource-interfaces/resource-interface.ts
```diff
@@ -18,7 +18,8 @@ export interface ResourceIF {
       subtitle: string
       helpSection?: HelpSectionIF
     }
-    nameRequestTypes?: Array<NameRequestTypes>
+    nameRequestTypes?: Array<NameRequestTypes>,
```
severinbeauvais, line 21: We don't need commas between properties in an interface.

I don't care which way we go here as long as it's consistent.

tests/unit/ChangeBusinessType.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+      propsData: { invalidSection: true }
+    })
+    const vm: any = wrapper.vm
+
+    expect(vm.invalidSection).toBe(true)
+
+    wrapper.destroy()
+  })
+
+  it('Should have tool for Cooperative', () => {
```
severinbeauvais, line 47: tooltip

tests/unit/ChangeBusinessType.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    store.state.stateModel.entitySnapshot = { authInfo: { folioNumber: 'A123' } }
+
+    const wrapper = mount(ChangeBusinessType, { vuetify, store })
+
+    expect(wrapper.find('.tooltip-info').exists()).toBe(true)
+
+    wrapper.destroy()
+  })
+
+  it('Should not have tool for Benifit company', () => {
```
severinbeauvais, line 61: benefit

tests/unit/OfficeAddresses.spec.ts
```diff
@@ -1085,3 +1085,36 @@ describe('actions and events', () => {
     expect(wrapper.vm.haveOfficeAddressesChanged).toEqual(true)
   })
 })
+
+describe('For Special rsolution', () => {
```
severinbeauvais, line 1089: resolution

severinbeauvais (approved):
Just a few small things.

I trust you to update these and merge when you're ready.

src/components/common/YourCompany/ChangeBusinessType.vue
shabeeb-aot: This was the existing condition in HTML moved added for cp and moved to getter. Sorry i didnt get this exactly something like this? return this.isChangeRegFiling || this.isEntityTypeFirm || this.isSpecialResolutionFiling   ?

severinbeauvais: Yes, you updated an existing condition -- thank you.

I'm saying you don't need the first condition here because only firms can do a ChangeReg filing, so if the second condition is true then the first condition is also/always true.

src/components/common/YourCompany/OfficeAddresses.vue
shabeeb-aot: class for the tooltip is not aligning properly, (seems like the tooltip class has flex-align-center)
Screen Shot 2022-07-28 at 11 36 14 AM

when we give class for div its aligning properly
Screen Shot 2022-07-28 at 11 39 44 AM

src/components/common/YourCompany/ChangeBusinessType.vue
shabeeb-aot: oh, I see, 
so we need only 
this.isEntityTypeFirm || this.isEntityTypeCP
or 
 this.isChangeRegFiling  || this.isSpecialResolutionFiling
?

shabeeb-aot: updated. Thank you :)

src/components/common/YourCompany/OfficeAddresses.vue
severinbeauvais: Why does the one in ChangeBusinessType.vue work? I'm hoping for consistency.

tests/unit/ChangeBusinessType.spec.ts
severinbeauvais: Also "tooltip" here.

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: Just this: `this.isEntityTypeFirm || this.isEntityTypeCP`

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -304,6 +305,11 @@ export default class ChangeBusinessType extends Mixins(CommonMixin, SharedMixin)
     return this.getResource.changeData?.typeChangeInfo
   }
 
+  /** v-if for tooltip (shoe only firm and coop)  */
```
severinbeauvais, line 308: Typo. But how about "only applies to firm and coop filings"?

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: Thanks for the cleanup!

src/components/common/YourCompany/OfficeAddresses.vue
shabeeb-aot: oh, There we don't need to align to the end, it is next to the text, so we don't need additional class there. Here we need to show at the end of the column
we can use like in the change button (actions class). but it has a minus margin to fix button, so we cant use it
Screen Shot 2022-07-28 at 11 50 22 AM

severinbeauvais: What bugs me here is that we have the same classname used for 2 different types of elements.

How about renaming this one "tooltip-wrapper" or "tooltip-container"?

Alternately, you could delete the classname from both (since you don't use the classname to set CSS properties) and then in your unit tests you can select the elements using other ways (for example, the Vuetify tootip component has its own class names already).

Up to you, but this bugs me a little :)

shabeeb-aot: Agreed, removed extra class and use vutify class for testing :)

shabeeb-aot: updated , shall i merge it ?

severinbeauvais: Looks great! Yes, please merge this.

PR #340, opened 2022-07-29
Fix folio number call to sbc-auth.

seeker25:
*Description of changes:*
Small bug fix, adding in new flag for special resolution filing as well. 

### DIFF
--- a/src/services/auth-services.ts
+++ b/src/services/auth-services.ts
@@ -114,7 +114,7 @@ export default class AuthServices {
 
     const authApiUrl = sessionStorage.getItem('AUTH_API_URL')
     const url = `${authApiUrl}entities/${businessId}`
-    const data = { businessId, folioNumber }
+    const data = { businessIdentifier: businessId, folioNumber: folioNumber ?? '' }
 
     return axios.patch(url, data)
   }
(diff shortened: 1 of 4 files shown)


seeker25:
@severinbeauvais This fixes this one:
image

I noticed there is this one as well, but it doesn't update right away:
image

Any idea which one we should be using for Special Resolution? 

Related ticket: https://github.com/bcgov/entity/issues/12897

src/services/auth-services.ts
```diff
@@ -114,7 +114,7 @@ export default class AuthServices {
 
     const authApiUrl = sessionStorage.getItem('AUTH_API_URL')
     const url = `${authApiUrl}entities/${businessId}`
-    const data = { businessId, folioNumber }
+    const data = { businessIdentifier: businessId, folioNumber: folioNumber ?? '' }
```
severinbeauvais, line 117: I think `??` isn't the right operator. Do you mean '||' (as in, if folioNumber is falsy then fall back to empty string)?

Also, as this is an object, sometimes it's nice to put the properties on separate lines. Not really needed here unless you want to add a comment beside the second property to say what you're doing?

Also, do you want to fall back to an empty string, null, or undefined (which takes the property out of the resulting JSON object)?

severinbeauvais (approved):
Looks good. Just a couple of small questions.

src/services/auth-services.ts
seeker25: `??` checks to see if undefined or null - They probably ported this over from C#, my intent was to avoid a null or undefined folioNumber, it causes the API call to fail

`||` = truthy  0 || "not found" -> "not found"

I think either operator could be used here, but i'll change it to ||

PR #353, opened 2022-08-04
12894 - UI for Change in Business Name

seeker25:
*Issue #:* /bcgov/entity#12894

*Description of changes:*
- New FeeMixin
- UI changes for Business Name for Special Resolution
- Divider fix


src/mixins/filing-template-mixin.ts
```diff
@@ -319,7 +319,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
 
     // Apply NR / business name / business type change to filing
     if (this.getNameRequestNumber || this.hasBusinessNameChanged || this.hasBusinessTypeChanged) {
-      filing.alteration.nameRequest = this.getNameRequest
+      filing.specialResolution.nameRequest = this.getNameRequest
```
seeker25, line 322: Might require some review under our finish off ticket:
#13231

seeker25:
Looks like there are some changes to other repos to make this work. I don't believe this is possible? 

image

seeker25:
I've tested this with Special Resolution, but haven't tested it with other business types. That should probably be looked at before this is reviewed.

src/interfaces/special-resolution-interfaces/special-resolution-interface.ts
```diff
@@ -9,4 +10,5 @@ export interface SpecialResolutionIF {
   nameRequest?: NameRequestIF,
   contactPoint?: ContactPointIF,
   courtOrder?: CourtOrderIF
+  cooperativeAssociationType: AssociationTypes
```
seeker25, line 13: I think I've found a way we can generate this interface from the JSON schema directly

removed the URLs.. removed the .json extension, used this: https://github.com/bcherny/json-schema-to-typescript

seeker25:
Alteration - Tested the future effective fee on review page, tested staff payment- /BC0871146/alteration  + Priority fee
	I couldn't seem to file (schema issues), but everything seemed to be working? I had to do a minor change here to make the summary update after a draft.
I tried a Change - seemed to work, no fee  - /FM1009715/change
I tried a conversion - seemed to work  /FM1009715/conversion
FM Correction / BEN Correction have minimal changes

src/interfaces/special-resolution-interfaces/special-resolution-interface.ts
severinbeauvais: Could you show this tomorrow at Developer Demo?

severinbeauvais:
I will review after others have reviewed.

seeker25:
Ok, still have to fix a unit test I just busted :(

src/interfaces/special-resolution-interfaces/special-resolution-interface.ts
seeker25: Yes, here are some examples
[special_resolution.d.txt](https://github.com/bcgov/business-edit-ui/files/9294947/special_resolution.d.txt)
[alteration.d.txt](https://github.com/bcgov/business-edit-ui/files/9294948/alteration.d.txt)
:

severinbeauvais: Nice! So when do the IFs get created -- are they available during implementation of a new feature?

seeker25: as long as someone updates the schema repo, and regenerate the types (preferably with a version number attached) I don't see why not

src/components/Alteration/AlterationSummary.vue
```diff
@@ -235,17 +234,22 @@ export default class AlterationSummary extends Mixins(
...
   // sum of alteration fees
   get alterationFees (): string {
-    if (this.getCurrentFees.filingFees !== null && this.getCurrentFees.futureEffectiveFees !== null) {
-      return `($${(this.getCurrentFees.filingFees + this.getCurrentFees.futureEffectiveFees).toFixed(2)} Fee)`
+    const validFees = this.getCurrentFees.filter(f => f.filingFees !== null && f.futureEffectiveFees !== null)
+    if (validFees.length === 0) {
+      return ''
     }
-    return ''
+    const filingFeesSum = validFees.map(f => f.filingFees).reduce((a, b) => a + b, 0)
```
severinbeauvais, line 241: This (and other statements in this method) needs comments to explain what you're doing, please.

lewischenstudio:
@seeker25 Do you have a NR number I can use to enter the NR field for testing?

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -604,4 +613,8 @@ export default class YourCompany extends Mixins(
 #contact-info-section {
   border-bottom-left-radius: 0 !important;
 }
+
+.capitalize {
```
severinbeauvais, line 617: There's already a Vuetify helper class for this.

https://vuetifyjs.com/en/styles/text-and-typography/#transform

src/views/Correction/FmCorrection.vue
```diff
@@ -135,15 +131,6 @@ export default class FmCorrection extends Mixins(CommonMixin, FilingTemplateMixi
...
   }
 
-  protected onStaffPaymentChanges (): void {
-    // update filing data with staff payment fields
-    this.setFilingData({
-      ...this.getFilingData,
-      priority: this.getStaffPayment.isPriority,
-      waiveFees: (this.getStaffPayment.option === StaffPaymentOptions.NO_FEE)
-    })
-  }
```
severinbeauvais, line 145: Thanks for the cleanup!

tests/unit/AlterationSummary.spec.ts
```diff
@@ -8,6 +8,7 @@ import AlterationSummary from '@/components/Alteration/AlterationSummary.vue'
 import { ConfirmDialog as ConfirmDialogShared } from '@bcrs-shared-components/confirm-dialog/'
 import EffectiveDateTime from '@/components/common/EffectiveDateTime.vue'
 import NameTranslation from '@/components/common/YourCompany/NameTranslations/NameTranslation.vue'
+import { FeesIF } from '@/interfaces'
```
severinbeauvais, line 11: No. I think the unit tests should not know about interfaces. They should be hard-coded. This will make them more fragile and pre-empt the possibility of 2 wrongs canceling each other out.

tests/unit/SpecialResolutionSummary.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      'priorityFees': 0,
+      'processingFees': 0,
+      'serviceFees': 0,
+      'tax': {
+        'gst': 0,
+        'pst': 0
+      },
+      'total': 70.0
+    }] as FeesIF[]
+    await Vue.nextTick()
```
severinbeauvais, line 54: You shouldn't need this after assigning state data, though you might need it after the mount itself.

seeker25:
NR 3930724 I was testing with @lewischenstudio

tests/unit/SpecialResolutionSummary.spec.ts
severinbeauvais: Or, if you need to wait for stuff to happen after you set this, you could try awaiting the commit (instead of direct-assigning)?

Or, just add a comment why you need to await the next tick.

tests/unit/AlterationSummary.spec.ts
seeker25: Ok, I added it in there, because the any object was being assigned an object (FeesIF) instead of an array FeesIF[]

I'll revert those changes

src/components/Alteration/AlterationSummary.vue
seeker25: `filingFeesSum` doesn't describe it enough? Ok I'll add a comment

lewischenstudio:
> NR 3930724 I was testing with @lewischenstudio

image

I tried it just now, but it doesn't work in my end.

src/components/Alteration/AlterationSummary.vue
severinbeauvais: The names are very good. Let's see what the other reviewers think. I was trying to understand the chained map and reduce.

src/views/SpecialResolution.vue
```diff
@@ -252,25 +239,30 @@ export default class SpecialResolution extends Mixins(
...
+            entityType: CorpTypeCd.COOP,
+            priority: false
+          })
+        }
+        filingData.forEach(fd => {
+          fd.futureEffective = this.getEffectiveDateTime.isFutureEffective
+        })
+        this.setFilingData(filingData)
+
+        this.setFilingData(filingData)
```
lewischenstudio, line 255: I find this part confusing, so other codes using this array style become difficult to understand too. Correct me if I'm wrong. I guess that you want to add the three key-values `filingTypeCode`, `entityType` and `priority` to the dictionary `filingData`, as well as `futureEffective` in the next line. What I don't understand is why you use an array for this, and as a result, all files using `filingData` need to be changed because of it? It's not very convincing to me that we should turn `filingData` into an array.

seeker25: You need an array because the base filing has an additional fee schedule for a NAME CHANGE.  (You'll see when you change the NR)

seeker25: Also use travis8814@gmail.com for the NR (you'll see what I mean)

seeker25: image

seeker25: Change of Name and Special resolution are two different fee schedules (which is why we needed the array, most of the other filings are just a single fee schedule (with priority / future included)

seeker25: Here's what it looks like in the pay admin: 

image

image

image

src/App.vue
```diff
@@ -354,6 +354,15 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
 
+  /** True if there is filing data for corrections - corrections has a single filing schedule. */
+  get correctionHasFilingData () : boolean {
+    return Boolean(
+      this.getFilingData?.length > 0 &&
+      this.getFilingData[0].filingTypeCode &&
+      this.getFilingData[0].entityType
+    )
+  }
+
```
lewischenstudio, line 365: Quoting from your last comment "... because the base filing has an additional fee schedule for a NAME CHANGE...". By "additional fee schedule", do you mean here?

seeker25: Corrections only has a single filing schedule 

image

Has two different fee schedules (depending on corp_type):

image

lewischenstudio: Okay. I understand now. Thanks.

src/views/SpecialResolution.vue
lewischenstudio: I find it difficult to process this array style. If the backend is not currently returning an array, or the frontend is not sending an array in the payload, I think it's easier to use another getter/setter for this special case. This is working properly, so it should be good. I'll resolve this conversation once this duplicate `this.setFilingData(filingData)` is removed.

seeker25: Arg, sorry about the duplicate I think that was from the merge from main I believe.

Would making another getter/setter for example `nameChangeFilingData` and append that to `filingData` would that make more sense?

Let me know what you have in mind and I'll gladly implement it.

src/App.vue
```diff
@@ -105,7 +105,7 @@
                 :offset="{ top: 86, bottom: 12 }"
               >
                 <!-- Corrections still use the basic Fee Summary component -->
-                <aside v-if="isCorrectionFiling && getFilingData.filingTypeCode && getFilingData.entityType">
+                <aside v-if="isCorrectionFiling && correctionHasFilingData ">
```
lewischenstudio, line 108: There is a space at the end and before the double-quote.

src/App.vue
```diff
@@ -314,7 +314,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   get feeSummaryConfirmLabel (): string {
     const isNoFee = this.isFirmChangeFiling || this.isFirmConversionFiling
     if (this.isSummaryMode) {
-      return (isNoFee && !this.getFilingData.priority) ? 'File Now (No Fee)' : 'File and Pay'
+      return (isNoFee && !this.getFilingData.some(fd => fd.priority)) ? 'File Now (No Fee)' : 'File and Pay'
```
lewischenstudio, line 317: [Lines 48-65](https://github.com/bcgov/business-edit-ui/blob/8bbfb5c46b40c4b4013b99b3ddf05430fce0a449/src/mixins/fee-mixin.ts#L48-L65) maps the priority and waveFees from getStaffPayment. If priority is false from getStaffPayment, then each priority in the array is false. While this way works too, we only need to verify one priority.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -502,8 +510,9 @@ export default class YourCompany extends Mixins(
 
   /** Reset company name values to original. */
   protected resetName () {
-    // reset business information
-    this.setBusinessInformation(this.getEntitySnapshot.businessInfo)
+    // reset business information, except for association type.
+    const businessInfo = { ...this.getEntitySnapshot.businessInfo, associationType: this.getAssociationType }
+    this.setBusinessInformation(businessInfo)
```
lewischenstudio, line 516: When you reset name, do you also need to reset the filingData array?

src/mixins/fee-mixin.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  /** Lines up promises to fetch for filing fees, falls back to empty fees.  */
+  protected async getFilingFeesFromFilingData (isFutureEffective: boolean = false): Promise<FeesIF[]> {
+    try {
+      const feePromises = this.getFilingData
+        .map(fd => this.fetchFilingFees(fd.filingTypeCode, fd.entityType, isFutureEffective))
+      return await Promise.all(feePromises)
+    } catch (error) {
+      console.log(error)
+      return this.getFilingData.map(fd => cloneDeep(EmptyFees))
```
lewischenstudio, line 73: I'm not sure what will happen if we do `cloneDeep(EmptyFees)` if it returns an error. EmptyFees is used for initialization only. Did you test it with errors?

src/views/SpecialResolution.vue
```diff
@@ -317,14 +309,35 @@ export default class SpecialResolution extends Mixins(
...
+          entityType: CorpTypeCd.COOP,
+          priority: false
+        })
+      }
+      this.setFilingData(filingData)
+      // update the current fees for the filing
+      await this.setCurrentFeesFromFilingData(this.getEffectiveDateTime.isFutureEffective)
+      // update the fee prices to display in the text
+      await this.setFeePricesFromFilingData(true)
+    }
```
lewischenstudio, line 340: I have a question sort of related to this. When we successfully changed the business name using name request, `hasBusinessNameChanged` will be true and the special resolution form is rendered (line 187). If you save and resume this filing, will it display the special resolution form? Another way to ask this is that did you update the `legalName` in `nameRequest`? I think this part is missing.

seeker25: Will be handled in: https://github.com/bcgov/entity/issues/13021 - Saving/Resuming a Draft Special Resolution

src/interfaces/special-resolution-interfaces/special-resolution-interface.ts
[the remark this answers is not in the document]
seeker25: This should be covered by this ticket: https://github.com/bcgov/entity/issues/13231

src/mixins/fee-mixin.ts
seeker25: It was reused code that already exists. It just sets an empty fee.

src/components/common/YourCompany/YourCompany.vue
seeker25: seems to be handled by the watch on `businessNameChanged`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "3.5.21",
+  "version": "3.5.22",
```
severinbeauvais, line 3: Sorry, this is now out of date.

src/mixins/fee-mixin.ts
severinbeauvais: I am not worried about `cloneDeep()` returning an error. If it does then we probably have more serious issues to deal with 😄

src/App.vue
severinbeauvais: fix confirmed

src/components/Alteration/AlterationSummary.vue
severinbeauvais: Thanks for the extra comments. Further understanding the map and reduce is an exercise left to the reader 😄

tests/unit/SpecialResolutionSummary.spec.ts
severinbeauvais: Comments confirmed, thanks.

lewischenstudio (approved):

severinbeauvais (approved):

PR #355, opened 2022-08-04
 UI for Special Resolution Form Details#12926

shabeeb-aot:
*Issue #:* /bcgov/entity#12926

*Description of changes:*
* Special resolution form 


shabeeb-aot:
I am adding will add unit test case and validation for this component

src/interfaces/special-resolution-interfaces/create-resolution-interface.ts
```diff
@@ -0,0 +1,10 @@
+import { PersonIF } from '@/interfaces'
+
+export interface CreateResolutionIF {
+  // validationDetail: ValidationDetailIF
+  resolutionConfirmed: boolean
+  resolutionDate?: string
+  resolutionText?: string
+  signingPerson?: PersonIF
+  signingDate?: string
```
severinbeauvais, line 9: It would be helpful to document, beside each date (lines 6 and 9), what format they are, eg, `// YYYY-MM-DD`.

src/interfaces/special-resolution-interfaces/person-interface.ts
```diff
@@ -0,0 +1,12 @@
+export interface PersonIF {
```
severinbeauvais, line 1: Please rename this (and EmptyPerson below) to distinguish them from OrgPersonIF and ApiPersonIF.

src/interfaces/special-resolution-interfaces/person-interface.ts
```diff
@@ -0,0 +1,12 @@
...
+  givenName: string
+  additionalName?: string
+  familyName: string
+  email?: string
+}
+
+export const EmptyPerson: PersonIF = {
+  givenName: '',
+  familyName: '',
+  additionalName: ''
```
severinbeauvais, line 11: The empty values should be the same as those returned by the API when a person object is fetched.

For example, will all of theses be empty strings? Or could they be null?

Here's what I'm trying to avoid for you: https://github.com/bcgov/business-edit-ui/blob/e7f3877a3f870a0ddba63c9524f8f97073614829/src/store/getters/state-getters.ts#L728

severinbeauvais: Maybe the optional fields should be null?

Or maybe additionalName shouldn't be optional?

src/interfaces/store-interfaces/state-model-interface.ts
```diff
@@ -38,5 +38,6 @@ export interface StateModelIF {
   validationFlags: ValidationFlagsIF
   summaryMode: boolean,
   currentFees: FeesIF,
-  feePrices: FeesIF
+  feePrices: FeesIF,
+  createResolutionStep: CreateResolutionIF
```
severinbeauvais, line 42: Let's get away from calling these "steps" (which is left over from Create UI).

This is just a data object, and could be reused by any filing.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
+<template>
+  <v-card flat id="special-resolution-summary">
+
+    <v-card flat>
```
severinbeauvais, line 4: Duplicate v-card.

Also please watch your whitespace and indents.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+
+        <HelpSection
+            class="ma-6"
+            :helpSection="helpSection"
+          />
+
+      <!-- Special Resolution Form -->
+        <section id="sample-resolution-section" class="section-container mt-10">
+          <header id="sample-resolution-header">
+            <h2>{{getCreateResolutionResource && getCreateResolutionResource.header}}</h2>
```
severinbeauvais, line 38: getCreateResolutionResource should never be falsy right?

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+            </v-card>
+
+        </section>
+      </v-card>
+    </v-card>
+  </v-card>
+</template>
+
+<script lang="ts">
+// this is a placceholder copied from AlterationSummary, Will add component when working on this page
```
severinbeauvais, line 195: Delete.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+@Component({
+  components: {
+    HelpSection,
+    DatePickerShared
+  }
+})
+export default class SpecialResolutionForm extends Mixins(
+  DateMixin
+
+) {
```
severinbeauvais, line 213: This is a short line. Move it all to the same line please.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+    resolutionDatePickerRef: DatePickerShared,
+    signatureDatePickerRef: DatePickerShared,
+    signingPersonFormRef: FormIF,
+  }
+
+  // Date properties
+  protected resolutionDateText: string = ''
+  protected signatureDateText: string = ''
+
+  protected MAX_RESOLUTION_TEXT_LENGTH: number = 1000
```
severinbeauvais, line 232: This should be
`readonly MAX_RESOLUTION_TEXT_LENGTH = 1000`

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+  // Refs
+  $refs!: {
+    resolutionDatePickerRef: DatePickerShared,
+    signatureDatePickerRef: DatePickerShared,
+    signingPersonFormRef: FormIF,
+  }
+
+  // Date properties
+  protected resolutionDateText: string = ''
+  protected signatureDateText: string = ''
```
severinbeauvais, line 230: Some versions of lint complain that you don't need to declare it's a string when it's already initialized as a string, so I take that as a best practice :)

-> Local variables should not have an explicit type when they're initialized to something with a type.

Examples:
```
protected myString = ``
protected myNumber = 123
protected myObj = null as myObjIF
```

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+  protected resolutionText: string = ''
+
+  protected signingPerson: PersonIF = null
+
+  get helpSection (): HelpSectionIF {
+    return this.getResource.changeData?.specialSpecialResolution?.helpSection
+  }
+
+  get getCreateResolutionResource (): any {
+    return this.getResource.changeData?.specialSpecialResolution?.sampleFormSection || null
```
severinbeauvais, line 242: How about falling back to `{}` so that you can still de-reference it in the template (and remove the check if falsy)?

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+
+  get previewImageSource (): string {
+    // Note: the image file path did not resolve correctly when using the require function directly.  In order
+    // to get the image path resolving correctly, needed to get the image context first and use that to build
+    // the final image file path
+
+    if (this.isJestRunning || !this.getCreateResolutionResource?.previewImagePath) { return '' }
+    const images = require.context('@/assets/images/', false, /\.png$/)
+    return images('./' + this.getCreateResolutionResource?.previewImagePath)
+  }
```
severinbeauvais, line 259: Why don't you just hard-code the path in the img tag [like this](https://github.com/bcgov/business-create-ui/blob/73cd0b9344031b50579ff7bb898551c8adf0df17/src/views/Incorporation/IncorporationAgreement.vue#L196)?

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+
+  /** The name section validity state (when prompted by app). */
+  get invalidNameSection (): boolean {
+    // add more state here
+    return (this.getComponentValidate)
+  }
+
+  /** The minimum date that can be entered (can't be earlier than incorporation date ). */
+  get resolutionDateMinStr (): string {
+    return this.dateToYyyyMmDd(new Date(this.getBusinessFoundingDate))
```
severinbeauvais, line 269: Don't use `new Date()`. Ever. It creates a date object based on the user's browser and workstation settings.

Instead, in this case, use the Date Mixin methods to convert an "api date" to a JS date object.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+  }
+
+  /** The minimum date that can be entered (can't be earlier than incorporation date ). */
+  get resolutionDateMin (): Date {
+    return this.getBusinessFoundingDate
+  }
+  /** The maximum date that can be entered (today). */
+  get resolutionDateMax (): Date {
+    const date = new Date()
```
severinbeauvais, line 278: Again, don't do this. You don't know what date you'll get.

There's a getter for current date (yyyy-mm-dd) already. (And it is updated periodically in the background.)

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+    return [
+      v => (v && v.trim().length > 0) || 'Resolution text is required',
+      v => (v && v.length <= this.MAX_RESOLUTION_TEXT_LENGTH) || 'Maximum characters exceeded',
+      v => (v && /^([\w\s$&+,:;=?@#|'<>.^*()%!-\\"]*)$/g.test(v)) || 'Invalid characters'
+    ]
+  }
+
+  /** The maximum date that can be entered (today). */
+  get resolutionDateMaxStr (): string {
+    const date = new Date()
```
severinbeauvais, line 292: As above.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+  }
+
+  /**
+   * True if date is >= the minimum (ie, today) and <= the maximum (ie, the 10th day).
+   * This is used for Vue form validation (in Date Rules above).
+   */
+  // private isBetweenDates (v: string): boolean {
+  isBetweenDates (minDate: string, maxDate: string, dateStrToValidate: string): boolean {
+    if (!dateStrToValidate) { return true }
+    return (new Date(dateStrToValidate) >= new Date(minDate)) && (new Date(dateStrToValidate) <= new Date(maxDate))
```
severinbeauvais, line 317: These 3 "new Dates" are actually OK since you're only using them for comparison.

However, I think you could just compare the dates (yyyy-mm-dd), eg.
"is 2022-08-01 < 2022-08-02"? yes
"is 2022-08-01 > 2022-08-02"? no

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+  }
+  /** Called to add a new date. */
+  async onResolutionDateSync (val: string): Promise<void> {
+    await this.$nextTick()
```
severinbeauvais, line 321: Add a comment why you need this.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+  /** called to add new resolutionDateText. */
+  protected onResolutionTextChanged (val: string) {
+    this.setResolution({
+      ...this.getCreateResolutionStep,
+      resolutionText: val
+    })
+  }
+
+  /** Validation rule for name */
+  nameRules (label, isRequired = true): Array<Function> {
```
severinbeauvais, line 337: Should be private.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+    return [
+      v => isRequired ? (!!v?.trim() || `${label} is required`) : true,
+      v => (v?.length <= 30) || 'Cannot exceed 30 characters' // maximum character count
+    ]
+  }
+
+  /** Validation rule for individual name fields */
+  firstNameRules = this.nameRules('First Name');
+  middleNameRules = this.nameRules('Middle Name', false);
+  lastNameRules = this.nameRules('Last Name');
```
severinbeauvais, line 347: Don't do this. Declare all class variables at the top with the others. Also should not need semi-colons.

severinbeauvais: Maybe these rules can be readonly also?

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+    return this.dateToYyyyMmDd(this.signatureDateMin)
+  }
+
+  /** The minimum date that can be entered (resolution date). */
+  get signatureDateMin (): Date {
+    if (this.resolutionDateText) {
+      const resolutionDate = this.yyyyMmDdToDate(this.resolutionDateText)
+      return resolutionDate
+    }
+    const date = new Date()
```
severinbeauvais, line 360: No. Use getCurrentDate.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+  /** The maximum date that can be entered (today). */
+  get signatureDateMaxStr (): string {
+    const date = this.signatureDateMax
+    const dateStr = this.dateToYyyyMmDd(date)
+    return dateStr
+  }
+
+  /** The maximum date that can be entered (today). */
+  private get signatureDateMax (): Date {
+    const date = new Date()
```
severinbeauvais, line 374: No.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+        this.isBetweenDates(this.signatureDateMinStr,
+          this.signatureDateMaxStr,
+          v) ||
+        `Date should be between ${this.dateToPacificDate(this.signatureDateMin, true)} and
+         ${this.dateToPacificDate(this.signatureDateMax, true)}`
+    ]
+  }
+  /** called to add new signature date  */
+  async onSigningDateSync (val: string): Promise<void> {
+    await this.$nextTick()
```
severinbeauvais, line 392: Add a comment why you need this.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+    this.signatureDateText = val
+    this.setResolution({
+      ...this.getCreateResolutionStep,
+      signingDate: val
+    })
+  }
+  /** called to store sining party. */
+   @Watch('signingPerson', { deep: true })
+  protected async onSigningPersonChanged (): Promise<void> {
+    await this.$nextTick()
```
severinbeauvais, line 402: Add a comment why you need this.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,463 @@
...
+    this.setResolution({
+      ...this.getCreateResolutionStep,
+      signingPerson: this.signingPerson
+    })
+  }
+   /** Set values if exist */
+   created () {
```
severinbeauvais, line 409: Created should be protected, and our code style is to place it near the top of the class (possibly just under the local getters, which are kind of like variables).

severinbeauvais: Now that I look at what this does, can't you do it during the variable initializations?

If you depend on the getter, note that these will be assigned the getter values as of when this component is created. They will not be assigned a subsequent getter value (not reactive).

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,433 @@
...
+    return this.dateToYyyyMmDd(this.apiToDate(this.getBusinessFoundingDate))
+  }
+
+  /** The minimum date that can be entered (can't be earlier than incorporation date ). */
+  get resolutionDateMin (): Date {
+    return this.apiToDate(this.getBusinessFoundingDate)
+  }
+  /** The maximum date that can be entered (today). */
+  get resolutionDateMax (): Date {
+    return this.apiToDate(this.getCurrentDate)
```
severinbeauvais, line 268: `getCurrentDate` is not in API format. Use a different method to convert yyyy-mm-dd to a Date.

severinbeauvais: Also on lines 341 and 351.

src/interfaces/special-resolution-interfaces/create-resolution-interface.ts
```diff
@@ -0,0 +1,9 @@
+import { SigningPersonIF } from '@/interfaces'
+
+export interface CreateResolutionIF {
```
severinbeauvais, line 3: Since you are reusing this IF (hence its name from Create UI), please add a comment that this IF is also using for Special Resolution filings.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -0,0 +1,432 @@
...
+
+  protected signingPerson: SigningPersonIF = null
+
+  /** Validation rule for individual name fields */
+  readonly firstNameRules = this.nameRules('First Name')
+  readonly middleNameRules = this.nameRules('Middle Name', false)
+  readonly lastNameRules = this.nameRules('Last Name')
+
+  get helpSection (): HelpSectionIF {
+    return this.getResource.changeData?.specialSpecialResolution?.helpSection
```
severinbeauvais, line 238: This is used by HelpSection component. If that component does not already handle falsy (undefined) then you should have a fallback here (like line 242 maybe?).

src/interfaces/special-resolution-interfaces/signing-person-interface.ts
```diff
@@ -0,0 +1,12 @@
...
+  additionalName?: string
+  familyName: string
+  email?: string
+}
+
+export const EmptySigningPersonIF: SigningPersonIF = {
+  givenName: '',
+  familyName: '',
+  additionalName: null
+}
```
severinbeauvais, line 12: Additional Name is optional but null.
Email is also optional. Should it be initialized here?

src/resources/SpecialResolution/CooperativeResource.ts
```diff
@@ -20,7 +20,24 @@ export const CooperativeResource: ResourceIF = {
     typeChangeInfo: 'You cannot change the business type of a Cooperative Association. You must form a new' +
     'business and dissolve this business once the new business is registered.',
     addressChangeInfo: 'To change addresses, please use the Change feature in the' +
-    ' Office Addresses list on your business dashboard.'
+    ' Office Addresses list on your business dashboard.',
+    specialSpecialResolution: {
+      helpSection: {
+        header: 'Help with Special Resolution',
+        helpText: [
+          'If your require assistance with adding a business partner please contact us.'
```
severinbeauvais, line 28: Should be "If you require".

tests/unit/SpecialResolutionForm.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  const entitySnapshot = {
+    businessInfo: {
+      legalName: 'Mock Original Name',
+      legalType: 'CP'
+    }
+  }
+
+  beforeAll(() => {
+    // init store
+    store.state.stateModel.currentJsDate = new Date('2020-03-01T16:30:00Z')
```
severinbeauvais, line 30: This may not work in GH because it will have a different date library / timezone than you. If so then you will have to mock this getter.

tests/unit/SpecialResolutionForm.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    store.state.stateModel.tombstone.currentDate = '2021-03-01'
+    store.state.stateModel.entitySnapshot = entitySnapshot
+    store.state.stateModel.createResolution = {
+      resolutionText: '',
+      signingPerson: { ...EmptySigningPersonIF },
```
severinbeauvais, line 35: It's better to hard-code data here instead of using IFs or objects, so that 2 mistakes don't cancel each other out.

tests/unit/SpecialResolutionForm.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  })
+
+  it('renders the sample-resolution-header', async () => {
+    const instructional = wrapper.find('#sample-resolution-header').text()
+    expect(instructional).toContain('Special Resolution (Form 06 COO)')
+  })
+
+  it('displays the correct help text', async () => {
+    wrapper.find('.help-btn').trigger('click')
+    await Vue.nextTick()
```
severinbeauvais, line 67: You should be able to await the trigger function (previous line) and not need this line.

src/interfaces/special-resolution-interfaces/signing-person-interface.ts
```diff
@@ -0,0 +1,12 @@
+export interface SigningPersonIF {
+  givenName: string
+  additionalName?: string
+  familyName: string
+  email?: string
+}
+
+export const EmptySigningPersonIF: SigningPersonIF = {
```
severinbeauvais, line 8: Don't name this xxxIF. It's not an IF. It's just EmptySigningPerson (object).

src/interfaces/special-resolution-interfaces/signing-person-interface.ts
shabeeb-aot: Additional Name if we didn't define null have to put additional check html
we don't have an email in design

src/resources/SpecialResolution/CooperativeResource.ts
shabeeb-aot: this is the same for SP/GP also. I copied it from Sp/GP resource file. Can I change all places?

src/interfaces/special-resolution-interfaces/signing-person-interface.ts
severinbeauvais: I'm looking for consistency between the interface and this empty object. Why are additional name and email different? I'm thinking they should be declared and initialized the same...?

tests/unit/SpecialResolutionForm.spec.ts
shabeeb-aot: it fails without await. since its a click function

src/resources/SpecialResolution/CooperativeResource.ts
severinbeauvais: Oops! Yes, please fix all 3 :)

src/interfaces/special-resolution-interfaces/signing-person-interface.ts
shabeeb-aot: my bad, email is not there removed from interface

tests/unit/SpecialResolutionForm.spec.ts
severinbeauvais: Weird. It works in other unit tests in this project.

Maybe it's because you need to wait for some async validation to complete before you can verify the result. If this is so, please add a comment here that this is why the "await nextick" is needed.

You should also await the trigger anyway since it's async.

src/interfaces/special-resolution-interfaces/signing-person-interface.ts
severinbeauvais: Bye bye, email.

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: you mean to compare strings?

severinbeauvais: Yes, since they're all yyyy-mm-dd format, right?

severinbeauvais: I leave this up to you.

severinbeauvais (approved):
I'll leave the last decisions/cleanup up to you. Looks good otherwise.

severinbeauvais:
By the way, for future PR reviews, I will wait for the other reviewers to look at it before I do. This will give others a chance to comment and should reduce the load on me.

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: As far as I heard it's better to compare with a date than string.

tests/unit/SpecialResolutionForm.spec.ts
shabeeb-aot: I am not using this value to compare now, I just used it to render correctly, So the test will not fail in this case, I will revisit if needed

lambert-alex (approved):
LGTM

tests/unit/SpecialResolutionForm.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+  })
+
+  it('renders the sample-resolution-header', async () => {
+    const instructional = wrapper.find('#sample-resolution-header').text()
+    expect(instructional).toContain('Special Resolution (Form 06 COO)')
+  })
+
+  it('displays the correct help text', async () => {
+    wrapper.find('.help-btn').trigger('click')
+    // wait untill next render to reflect click changes
```
lambert-alex, line 72: Not important but should be until not untill

lewischenstudio:
Everything looks good, but the `package.json` file didn't update the app version. I'll approve this PR once this is updated.

lewischenstudio (approved):

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: OK.

PR #357, opened 2022-08-10
 UI for Special Resolution Form Details#12926

shabeeb-aot:
*Issue #:* /bcgov/entity#12926

*Description of changes:*
* validation for special resolution form


shabeeb-aot:
Will update app version once reviewed

lewischenstudio:
Can you link this PR to the original ticket? Since you are updating the special resolution, I think you might want to know the [schema for special resolution](https://github.com/bcgov/business-schemas/blob/395e2c90b28051ce14758f0fd06ef593baa9dbf3/src/registry_schemas/schemas/special_resolution.json#L11), where "resolution" is a required field (Create UI sets is required too). I am updating the special resolution schema in this [PR](https://github.com/bcgov/business-schemas/pull/103) to include the new fields like "resolutionText" etc, and either this or another ticket should make it required.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -309,20 +311,22 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
...
     })
+    this.validate()
   }
 
   /** Validation rule for name */
   private nameRules (label, isRequired = true): Array<Function> {
     return [
       v => isRequired ? (!!v?.trim() || `${label} is required`) : true,
-      v => (v?.length <= 30) || 'Cannot exceed 30 characters' // maximum character count
+      v => v?.trim() ? (v?.length <= 30) || 'Cannot exceed 30 characters' : true // maximum character count
```
lewischenstudio, line 329: v.trim() removes spaces before/after input, but space is a character long. The schema will likely not pass this.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -368,6 +372,7 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
       ...this.getcreateResolution,
       signingDate: val
     })
+    this.validate()
```
lewischenstudio, line 375: I'm not sure why we need to do validation here.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -376,10 +381,30 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
       ...this.getcreateResolution,
       signingPerson: this.signingPerson
     })
+    await this.$nextTick()
```
lewischenstudio, line 384: This looks like a duplicate when there is already `await this.$nextTick()` inside validate().

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: We need this to set the component state valid. so when there is change component state will set as valid or not

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: without this here it is not reflecting on the change of this object, this will happen if you enter one character in ```signingPerson``` object.
I hope there is 2 tick happen, since its updating on watch

severinbeauvais: I see that validate() already waits for async stuff to happen. You probably just need to await validate here (and delete the extra await nextTick).

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -96,7 +96,6 @@
                   </label>
                 </v-col>
                 <v-col cols="12" sm="9">
-                  <v-form ref="resolutionTextFormRef">
                     <v-textarea ref="resolutionTextRef"
```
lambert-alex, line 99: v-textarea is now indented one more space than necessary

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -130,7 +127,6 @@
                   <label class="resolution-signature-vcard-title">Signing Party</label>
                 </v-col>
                 <v-col cols="12" sm="9" class="pt-4 pt-sm-0">
-                  <v-form ref="signingPersonFormRef">
                     <div class="form__row three-column">
```
lambert-alex, line 130: Same as above comment

lambert-alex (commented):
Looks good should there be some added unit tests for some of the new functionality?

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -241,17 +243,25 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
...
   get documentURL (): string {
-    const docUrl = sessionStorage.getItem('BASE_URL') +
+    /**
+     * In session is stored the BASE_URL with business ID
+     * Here to acccess asset folder, need to remove business ID from URL
+     */
+    const baseUrlwithId = sessionStorage.getItem('BASE_URL')
+    const businessId = sessionStorage.getItem('BUSINESS_ID')
+
+    const baseUrlwithoutId = baseUrlwithId.replace(new RegExp(`${businessId}/?`), '')
```
shabeeb-aot, line 255: is there any other way to get only base URL without business-id?
in session storage, it is set along with business, Need to access the files folder in public folder, 
so remove business id from base URL and using it.
Please let me know is there any other way to access

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: Seems like on the first tick it updates the values and reflect `formValid` current value for all the components, So we need nextTick on validating function.

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: Why do you need to access files in the public folder? Or, why can't you do this the same way that Create UI does it?

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: What you have here is TWO await-nextTick: one here and one in validate().

Did you try awaiting validate() and deleting this line?

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-  <v-card flat id="special-resolution-summary">
+  <v-card flat id="create-special-resolution-form">
```
severinbeauvais, line 2: The pattern used in our UIs is to name this the same as the component name.

Also, this isn't a form (it's a card) so don't call it that.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -130,37 +127,35 @@
...
-                        ref="signingPersonMiddleNameRef"
-                        filled
-                        class="item"
-                        label="Middle Name (Optional)"
-                        id="person__middle-name"
-                        v-model="signingPerson.additionalName"
-                        :rules="middleNameRules"
+                  <div class="form__row three-column">
+                    <v-text-field
+                      ref="signingPersonGivenNameRef"
```
severinbeauvais, line 132: Don't assign a ref if you don't need it. Same with all other components below.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -208,16 +204,19 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
   @Getter getBusinessFoundingDate!: string
   @Getter getCurrentDate!: string
   @Getter getCurrentJsDate!: string
-
-  @Action setResolution!: ActionBindingIF
   @Getter getcreateResolution!: CreateResolutionIF
   @Getter getComponentValidate!: boolean
+  @Getter getCreateResolutionFromValid!: boolean
```
severinbeauvais, line 209: Typo.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -208,16 +204,19 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
...
+
+  @Action setResolution!: ActionBindingIF
+  @Action setValidComponent!: ActionBindingIF
 
   // Refs
   $refs!: {
     resolutionDatePickerRef: DatePickerShared,
     signatureDatePickerRef: DatePickerShared,
     signingPersonFormRef: FormIF,
+    specialResolutionFrom: FormIF,
```
severinbeauvais, line 219: I think you only need to declare the date picker refs.

src/components/SpecialResolution/SpecialResolutionForm.vue
```diff
@@ -376,10 +389,30 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
...
       signingPerson: this.signingPerson
     })
+    await this.$nextTick()
+    this.validate()
   }
-  /** Set values if exist */
+   /** called to store component validity to store. */
+   protected async validate () {
+     // wait to reflect validation state
+     await this.$nextTick()
```
severinbeauvais, line 398: This will do nothing unless you await this method.

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: Users want to download PDF, which is located in public/file folder. I create UI  we are using only BASEURL, but session it store only BASE_URL (without BusinessId), in edit UI, we are store BASE_URL with business ID
create UI
Screen Shot 2022-08-10 at 2 00 33 PM
In edit UI
Screen Shot 2022-08-10 at 2 01 00 PM

src/components/SpecialResolution/SpecialResolutionForm.vue
lewischenstudio: Alright.

lewischenstudio:
The [schema for special resolution](https://github.com/bcgov/business-schemas/blob/395e2c90b28051ce14758f0fd06ef593baa9dbf3/src/registry_schemas/schemas/special_resolution.json#L11) will use "resolution" instead of "resolutionText" and "signatory" instead of "signingPerson". Can you address these fields in this ticket or create another ticket for this issue? @argush3 is asking whether "resolutionConfirmed" is really required. Can you let us know why this is required in the schema?

argush3:
Actually, I don't believe `resolutionConfirmed` needs to be in the schema.  We've had discussions like this in the past.  We've never had a column for the confirmation in the database so there is no need to add this to the schema.  The FE can choose to add that value in the filing json and use it for FE purposes but doesn't use it.  

So unless there is a business requirement that we need to save the confirmation as a specific field in the database, I don't think we need `resolutionConfirmed` in the schema.

shabeeb-aot:
I can update the `resolution`
@argush3 I think we may need this for save and resume, As per design we have to persist the checkbox state.
https://projects.invisionapp.com/share/FY12X9VTMPD3#/screens/468822077
can you check and confirm

argush3:
@shabeeb-aot I'm saying that you can still add `resolutionConfirmed` to the filing json and use it to persist the state on the UI for resuming drafts.  We don't want `resolutionConfirmed` in the schema as the BE doesn't do anything with it.

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: Here is why the Edit UI base URL includes the business id.

https://github.com/bcgov/business-edit-ui/blob/6071429797646b3a9296414bf4da61fd040cb24a/src/utils/fetchConfig.ts#L133

Maybe you can use `process.env.BASE_URL` directly as per this code.

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: When I am removing validation is not working properly. 
I could not figure it out exactly. Am I missing something? 
with 2 await
[not-working2.webm](https://user-images.githubusercontent.com/55506249/184027210-37271f12-ef40-472b-9679-759adbf390ae.webm)
with one await
[not-working1.webm](https://user-images.githubusercontent.com/55506249/184027214-02741736-bf29-4702-9825-e2ba1d9d301b.webm)
with one await
[working1.webm](https://user-images.githubusercontent.com/55506249/184027184-19ed8e81-9bb4-4f8c-935a-a92d03505c3d.webm)

severinbeauvais:
> Argus is asking whether "resolutionConfirmed" is really required. Can you let us know why this is required in the schema?

Does this property need to be in the schema, or is it only used to save a flag for the UI.

But actually, I think we _should_ save whether the user has confirmed something, in the filing JSON, even if the API doesn't need. It's a record of what happened.

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: Yes, you are missing that you need to `await this.validate()` otherwise the inner async stuff won't work.

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: The ID looks OK but doesn't match the component filename. Would it make sense to name the component "CreateSpecialResolution" or "SpecialResolution" instead?

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: it didn't work for me :(
[notworking3.webm](https://user-images.githubusercontent.com/55506249/184029911-7179c1de-5e2f-408a-9c0f-a9ca6973aecf.webm)

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: we already have "SpecialResolution" for outer page, i will rename to CreateSpecialResolution

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: This is the code in your video:

image

You are still not awaiting the validate method.

Try this:
line 387: delete it
line 388: `await this.validate()`

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: Sounds good.

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: with this code also not working
```
   @Watch('signatory', { deep: true })
  protected async onsignatoryChanged (): Promise<void> {
    this.setResolution({
      ...this.getcreateResolution,
      signatory: this.signatory
    })

    await this.validate()
  }
```
[notworking4.webm](https://user-images.githubusercontent.com/55506249/184032176-ee8277eb-7398-424d-9894-8f87a4eb7388.webm)

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: updated

src/interfaces/special-resolution-interfaces/create-resolution-interface.ts
```diff
@@ -4,7 +4,7 @@ import { SigningPersonIF } from '@/interfaces'
 export interface CreateResolutionIF {
   resolutionConfirmed: boolean
```
lewischenstudio, line 5: The question becomes in a way that once we submit "resolutionConfirmed" to BE, then next time loading the filing will use this property from BE. If we want to check for whether resolution form has been changed, then we can perhaps use a getter from the entitySnapshot and resolution properties, like "hasBusinessNameChanged". I'm not sure if that is the way to go, since we don't have special resolution for the entitySnapshot yet. I'm kind of confused as to what this property is actually doing.

shabeeb-aot: I can make this property as optional, I am not sure where we are taking entitySnapshot, If it's from draft , we can reset it from there. I think after above discussion we will have this value only if user clicks on save and resume later

lewischenstudio:
> > Argus is asking whether "resolutionConfirmed" is really required. Can you let us know why this is required in the schema?
> 
> Does this property need to be in the schema, or is it only used to save a flag for the UI.
> 
> But actually, I think we _should_ save whether the user has confirmed something, in the filing JSON, even if the API doesn't need. It's a record of what happened.

I agree, and we are using entitySnapshot. I just looked the invision and found out that we want to have check the checkbox for "Confirm Special Resolution". If this property "resolutionConfirmed" is for this checkbox, I think we don't have to send it to BE, and we need something to track this checkbox for validation before clicking on "File and Pay", which is probably being handled inside the form, I think.

src/interfaces/special-resolution-interfaces/create-resolution-interface.ts
lewischenstudio: `this.parseSpecialResolutionFiling(filing, businessSnapshot)` is the one to set entity for specialResolution. We can add 'confirmed' to `ResolutionsIF` and initialize it like "hasBusinessNameChanged". We don't have to if the validation is handled inside the form. But when we click "File and Pay", if the confirmation checkbox is not checked, the UI should take us to the checkbox component.

shabeeb-aot:
> > > Argus is asking whether "resolutionConfirmed" is really required. Can you let us know why this is required in the schema?
> > 
> > 
> > Does this property need to be in the schema, or is it only used to save a flag for the UI.
> > But actually, I think we _should_ save whether the user has confirmed something, in the filing JSON, even if the API doesn't need. It's a record of what happened.
> 
> I agree, and we are using entitySnapshot. I just looked the invision and found out that we want to have check the checkbox for "Confirm Special Resolution". If this property "resolutionConfirmed" is for this checkbox, I think we don't have to send it to BE, and we need something to track this checkbox for validation before clicking on "File and Pay", which is probably being handled inside the form, I think.

Agree we may need to validate checkbox and remove it before sending to BE. We need to send this to the draft.
As per the design, we need to persist the status of the checkbox on " save" and "save and resume later".Hope  both are same since we don't have any button to save only

Screen Shot 2022-08-10 at 4 19 34 PM

lewischenstudio:
> export interface CreateResolutionIF {

I remember last time I worked on this checkbox, the user needs to check off it again every time it is resumed from draft. Can you verify with the design team whether we need to check off the confirmation every time we resume the filing, or we check off once and it will stay checked?

shabeeb-aot:
yeah, I can check with them again. We have another ticket for that 
https://app.zenhub.com/workspaces/relationships-team-space-61435c47e483c4000f08e9f6/issues/bcgov/entity/13289

src/interfaces/special-resolution-interfaces/create-resolution-interface.ts
shabeeb-aot: okey i will remove `resolutionConfirmed ` from this file and add it in `ResolutionsIF `

src/components/SpecialResolution/SpecialResolutionForm.vue
lewischenstudio: I found the validation method in this component works like a charm: [link](https://github.com/bcgov/business-edit-ui/blob/main/src/components/common/YourCompany/CompanyName/CorrectNameRequest.vue). Hope this can give a little help.

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -368,18 +371,39 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
...
+    this.validate()
   }
-  /** Set values if exist */
+   /** called to store component validity to store. */
+   protected async validate () {
+     // wait to reflect validation state
+     await this.$nextTick()
+     // date validation
+     const isResolutionDateValid = this.$refs?.resolutionDatePickerRef?.isDateValid()
+     const isSignatureDateValid = this.$refs?.signatureDatePickerRef?.isDateValid()
```
lewischenstudio, line 392: I pulled the PR and found out there is an error in `isDatevalid()`. If it's not in your end, then it's fine.

src/interfaces/special-resolution-interfaces/create-resolution-interface.ts
lewischenstudio: This will make sure the schema is correct upon saving the draft. The question is how do we deal with the confirmation checkbox. I see there is another ticket for this issue, so we can address this issue in that ticket. I want to hear from other reviewers about this.

src/interfaces/alteration-interfaces/resolutions-interface.ts
```diff
@@ -2,5 +2,6 @@
 export interface ResolutionsIF {
   date: string,
   id: number,
-  type: string
+  type: string,
+  confirmed:boolean,
```
lewischenstudio, line 6: I think we should remove "confirmed" here and use "resolutionConfirmed" in "CreateResolutionIF". What you had in the previous commit was the correct. When we save draft, we will remove "resolutionConfirmed" from `buildSpecialResolutionFiling`: `specialResolution: {... `. This will ensure the schema is correct and enable validation for the checkbox (with additional small work).

src/interfaces/special-resolution-interfaces/create-resolution-interface.ts
severinbeauvais: My ears are burning, lol.

In other filings, we save some certify checkbox flags in the filing JSON even though the API doesn't use it.

In other filings, there is a requirement to NOT restore the checkbox when a draft is resumed. I know this filing is different.

As for where to store this flag, I think it's up to you. If you think it fits well in the Resolution object then use that. Some other filings save the flag in the header object. If it looks right / feels right then I'm OK with it :)

lewischenstudio: Agreed. I withdraw the idea to use it in `ResolutionIF`. We can use this flag here and remove it from the payload.

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: Faiz said he tried this and it didn't work. Can this conversation be resolved? There's another conversation discussing this anyway.

lewischenstudio: I'll resolve this conversation as we can carry on the nextTick discussion below.

lewischenstudio:
You will need to remove "resolutionConfirmed" before sending the payload in either this ticket or another one: [here](https://github.com/bcgov/business-edit-ui/blob/6071429797646b3a9296414bf4da61fd040cb24a/src/mixins/filing-template-mixin.ts#L286-L332). The schema for "specialResolution" is [here](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/special_resolution.json). Aside from this and the nextTick issue, everything else is looking good.

shabeeb-aot:
> yeah, I can check with them again. We have another ticket for that https://app.zenhub.com/workspaces/relationships-team-space-61435c47e483c4000f08e9f6/issues/bcgov/entity/13289
> 
> The design team confirm user must re-check
> 
> Screen Shot 2022-08-10 at 4 59 53 PM

Yui replied we need to persist check box in draft
Screen Shot 2022-08-11 at 9 24 20 AM

severinbeauvais:
> You will need to remove "resolutionConfirmed" before sending the payload in either this ticket or another one: [here](https://github.com/bcgov/business-edit-ui/blob/6071429797646b3a9296414bf4da61fd040cb24a/src/mixins/filing-template-mixin.ts#L286-L332). The schema for "specialResolution" is [here](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/special_resolution.json). Aside from this and the nextTick issue, everything else is looking good.

Usually, schemas allow extra properties without error.

src/components/SpecialResolution/SpecialResolutionForm.vue
shabeeb-aot: In this link, we have all values in local property and form validation is emitted to different components. 
But in our case, there is no submit/done button or we dont have any parent to emit. we need to save component status into store. SO we cant use like link :(

lewischenstudio:
> Yui replied we need to persist check box in draft

In this case, we do need to send "resolutionConfirmed" to BE for saving the draft.

shabeeb-aot:
> > Yui replied we need to persist check box in draft
> 
> In this case, we do need to send "resolutionConfirmed" to BE for saving the draft.

yeah , I updated the PR with `resolutionConfirmed`

lewischenstudio:
Please also don't forget to update the app version 🙂

lambert-alex (approved):
LGTM, as long as the other reviewers are good with it

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -241,17 +238,21 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
   get getCreateResolutionResource (): any {
     return this.getResource.changeData?.specialSpecialResolution?.sampleFormSection || {}
   }
-
+  /** download URL fot pdf file */
```
severinbeauvais, line 241: typo

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -368,18 +371,42 @@ export default class SpecialResolutionForm extends Mixins(DateMixin) {
...
+     const isResolutionDateValid = this.$refs?.resolutionDatePickerRef?.isDateValid()
+     const isSignatureDateValid = this.$refs?.signatureDatePickerRef?.isDateValid()
+
+     const isFormValid = this.formValid && isResolutionDateValid && isSignatureDateValid
+     // settign component validity flag
+     this.setValidComponent({ key: 'isValidCreateSpecialResolution', value: isFormValid })
+   }
+
+   /** Set values if exist
+    * while coming back from summary page this form need to show existign values.
```
severinbeauvais, line 403: typo

src/interfaces/alteration-interfaces/resolutions-interface.ts
```diff
@@ -2,5 +2,5 @@
 export interface ResolutionsIF {
   date: string,
   id: number,
-  type: string
+  type: string,
```
severinbeauvais, line 5: Comma is not needed at end.

Actually, commas aren't needed for any property in an interface.

src/components/SpecialResolution/SpecialResolutionForm.vue
severinbeauvais: **Re: await issue**

Faiz and I looked at this together and tested it. We found that, if validate() is not awaited, it returns as soon as it hits its "await nextTick", and the statement after its call executes. This is risky code if there is a statement after the call to validate(), since validate() hasn't finished running. To resolve this, all calls to validate() are awaited (even though they cause more work for the framework and execution).

**Re: multiple nextTick issue**

A nextTick() is needed for the store action (set) to complete, and a second nextTick() is needed for the component validations to complete. The second nextTick already had a comment explaining it. A comment was added to the first nextTick to explain it.

I think this is a bit weird but (a) it's apparently needed for the reasons stated, and (b) this way each method handles what it needs to know about -- specifically, a method doesn't need to worry about another method's concerns.

 I have no further concerns with these issues and am resolving this conversation.

severinbeauvais (approved):
Looks good to me.

As there are no other outstanding issues, this is ready to merge.

PR #365, opened 2022-08-29
13021 - Special Resolution - Drafts

seeker25:
*Issue #:* /bcgov/entity#13021

*Description of changes:*
Special Resolution Drafts

Waiting on:

https://github.com/bcgov/bcrs-shared-components/pull/146
https://github.com/bcgov/business-filings-ui/pull/381


seeker25:
I don't see these in the schema (https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/name_request.json), but not sure if we want to add them? Any thoughts? 

image

image



The other issue is this data can be stale (NR could have expired or changed since..)  - Maybe saving just the NR and fetching it with the draft is better?


Example data:
image

src/mixins/filing-template-mixin.ts
```diff
@@ -90,6 +91,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
   @Action setNewResolutionDates!: ActionBindingIF
   @Action setFileNumber!: ActionBindingIF
   @Action setHasPlanOfArrangement!: ActionBindingIF
+  @Action setResolution!: ActionBindingIF
```
severinbeauvais, line 94: It would be nice if the action (and mutation) was named similarly to the getter.

src/services/legal-services.ts
```diff
@@ -1,9 +1,10 @@
 import { axios } from '@/utils/'
 import { AddressesIF, AlterationFilingIF, BusinessInformationIF, ChgRegistrationFilingIF, ConversionFilingIF,
-  CorrectionFilingIF, NameTranslationIF, OrgPersonIF, ResolutionsIF, ShareStructureIF, SpecialResolutionFilingIF }
+  CorrectionFilingIF, NameTranslationIF, OrgPersonIF, ResolutionsIF, SpecialResolutionFilingIF }
   from '@/interfaces/'
 import { RoleTypes } from '@/enums'
 
+import { ShareStructureIF } from '@bcrs-shared-components/interfaces/'
```
severinbeauvais, line 7: Yes, you've now seen that we don't re-export the shared interfaces in this UI. This was to avoid some cyclic imports at one time. I'm not sure if that still exists but I'm totally OK with each component importing local or shared imports (as long as we don't do it both ways simultaneously).

src/store/actions/actions-model.ts
```diff
@@ -270,10 +271,14 @@ export const setNaics: ActionIF = ({ commit }, naics: NaicsIF): void => {
   commit('mutateNaics', naics)
 }
 
-export const setResolution = ({ commit }, resolution: CreateResolutionIF) => {
+export const setResolution = ({ commit }, resolution: SpecialResolutionIF) => {
   commit('mutateResolution', resolution)
```
severinbeauvais, line 275: It would be nice if the state variable was named the same.

src/store/getters/state-getters.ts
```diff
@@ -927,7 +927,7 @@ export const getCompletingParty = (state: StateIF): CompletingPartyIF => {
 }
 
 /** The Create Special Resolution object. */
-export const getcreateResolution = (state: StateIF): CreateResolutionIF => {
+export const getCreateResolution = (state: StateIF): SpecialResolutionIF => {
   return state.stateModel.createResolution
```
severinbeauvais, line 931: Hmm, might be nice to have 1 name for the same thing instead of 3 or 4!

My suggestion: similar to its interface ;)

src/store/state/state-model.ts
```diff
@@ -135,7 +135,11 @@ export const stateModel: StateModelIF = {
   feePrices: [cloneDeep(EmptyFees)],
   createResolution: {
```
severinbeauvais, line 136: This name is historical: it was taken from Create UI, in which this data is actually "created". Here I think it makes more sense to just call it what it is: special resolution.

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -300,23 +308,30 @@ export default class CreateSpecialResolution extends Mixins(DateMixin) {
...
-  isBetweenDates (minDate: string, maxDate: string, dateStrToValidate: string): boolean {
+  private isValidDateRange (minDate: Date, maxDate: Date, dateStrToValidate: string): boolean {
     if (!dateStrToValidate) { return true }
-    return (new Date(dateStrToValidate) >= new Date(minDate)) && (new Date(dateStrToValidate) <= new Date(maxDate))
+    // only compare year/month/day (ignore time)
+    let date = new Date(dateStrToValidate)
+    date = new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
+    const minDay = new Date(minDate.getFullYear(), minDate.getMonth(), minDate.getDate())
+    const maxDay = new Date(maxDate.getFullYear(), maxDate.getMonth(), maxDate.getDate())
+    return (date >= minDay && date <= maxDay)
```
severinbeauvais, line 318: I'm a bit nervous when I see `Date()` used outside of the date mixin (in which much thought/testing was given to things). It's probably OK here because you're comparing results.

I'm a bit nervous that you're mixing UTC and non-UTC dates here. You may get unexpected results for the day if it's after 4pm or 5pm Pacific (depending on daylight time).

I'd prefer to see date methods in Date Mixin.

You may also find that you can use existing Date Mixin methods to compare dates (eg, convert to yyyy-mm-dd and then just apply string >= and <=).

Right now I don't have that warm, fuzzy feeling that this code works correctly in all cases.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "3.5.29",
+  "version": "3.5.30",
```
severinbeauvais, line 3: Please go to 3.6.0 due to new imports.

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -384,40 +390,45 @@ export default class CreateSpecialResolution extends Mixins(DateMixin) {
...
+    this.signatory = this.getCreateResolution.signatory ||
+      {
+        givenName: '',
+        familyName: '',
+        additionalName: null
+      }
+    this.resolution = this.getCreateResolution.resolution || ''
+    this.resolutionDateText = this.getCreateResolution.resolutionDate || ''
+    this.signingDate = this.getCreateResolution.signingDate || ''
+  }
```
severinbeauvais, line 431: FYI, getting the data here (from getters) works because this component isn't created until after the data is loaded. I think this assumption is solid, but if there's any reason the getter might have a different value after the component is created/mounted then you'll need to watch the getter.

No action needed here, just FYI.

src/interfaces/resource-interfaces/sample-form-interface.ts
```diff
@@ -0,0 +1,8 @@
+/** Interface to define the resource sample form. */
+
+export interface SampleFormIF {
+    header?: string
+    text?: string
+    downloadDocLabel?: string
+    downloadDocPath?: string
+}
```
severinbeauvais, line 8: This seems pretty specific (or rather, not global to any old sample form). Can you come up with a more specific name?

src/mixins/filing-template-mixin.ts
```diff
@@ -290,18 +292,27 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
...
         legalType: this.getEntitySnapshot.businessInfo.legalType,
-        nrNumber: this.getEntitySnapshot.businessInfo.nrNumber,
-        associationType: this.getAssociationType
+        nrNumber: this.getEntitySnapshot.businessInfo.nrNumber
       },
       specialResolution: {
-        ...this.getcreateResolution
+        ...this.getCreateResolution
+      },
+      alteration: {
```
severinbeauvais, line 300: This looks a bit strange: an alteration object next to the special resolution object. I would have expected one to be nested under the other (or the alteration properties under special resolution?), unless there's a good reason why it's like this.

src/mixins/filing-template-mixin.ts
```diff
@@ -290,18 +292,27 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
...
+        memorandumFileKey: 'test',
+        memorandumFileName: 'test'
       }
-
     }
 
     // Apply NR / business name / business type change to filing
     if (this.getNameRequestNumber || this.hasBusinessNameChanged || this.hasBusinessTypeChanged) {
-      filing.alteration.nameRequest = this.getNameRequest
+      filing.changeOfName = { nameRequest: this.getNameRequest }
```
severinbeauvais, line 315: Another object at the top level of the filing?

src/mixins/filing-template-mixin.ts
```diff
@@ -698,25 +709,27 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
     this.setEntitySnapshot(entitySnapshot)
 
     // store Entity Type
-    this.setEntityType(filing.alteration.business?.legalType || entitySnapshot.businessInfo.legalType)
+    this.setEntityType(filing.alteration?.business?.legalType || entitySnapshot.businessInfo.legalType)
```
severinbeauvais, line 712: Yes, the first part is to try to read the initial draft (that Filings UI created I think?).

src/mixins/filing-template-mixin.ts
```diff
@@ -698,25 +709,27 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
...
 
     // store Business Information
     this.setBusinessInformation({
       ...filing.business,
-      ...filing.alteration.business
+      ...filing.alteration.business,
```
severinbeauvais, line 717: Why conditionally-chain alteration everywhere but here?

src/mixins/filing-template-mixin.ts
```diff
@@ -1139,7 +1155,8 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
    * @param filing the filing to parse
    */
   private storeStaffPayment (
-    filing: AlterationFilingIF | CorrectionFilingIF | ConversionFilingIF | ChgRegistrationFilingIF
+    filing: AlterationFilingIF | CorrectionFilingIF | ConversionFilingIF |
+            ChgRegistrationFilingIF | SpecialResolutionFilingIF
```
severinbeauvais, line 1159: Not needed unless you also use `buildStaffPayment()`, which you don't, but maybe you should?

src/components/SpecialResolution/CreateSpecialResolution.vue
seeker25: This is directly copied from `EffectiveDateTime` 

I'd rather use a date library personally, moment is depreciated, maybe something like `date-fns`.

src/mixins/filing-template-mixin.ts
seeker25: Looks like it was it's own type of filing:

https://github.com/bcgov/business-schemas/blob/6b2268753999371b8731ef6a95218570c4b25412/src/registry_schemas/schemas/filing.json#L69

src/components/SpecialResolution/CreateSpecialResolution.vue
severinbeauvais: Modern JS should not need a date/time library. The main problem is devs not understanding how UTC time works --hence our Date Mixin. We can talk more about this if you like; I'm willing to change my mind :)

How about leaving EffectiveDateTime.vue alone and moving your code to the Date Mixin? This should result in an overall code simplification/improvement.

src/mixins/filing-template-mixin.ts
severinbeauvais: So is `alteration` left over from when the code was cut & pasted? Or does the API require several top-level objects like this?

Unfortunately our schemas don't cover this case; they mostly cover individual objects [like Special Resolution itself]()https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/special_resolution.json).

src/mixins/filing-template-mixin.ts
seeker25: Payload:

```
{
    "filing": {
        "header": {
            "name": "specialResolution",
            "date": "2022-08-02",
            "certifiedBy": "test test"
        },
        "business": {
            "legalType": "CP",
            "identifier": "CP1002552",
            "legalName": "TRAV COOP 2",
            "foundingDate": "2022-05-09T00:00:00+00:00"
        },
        "specialResolution": {
            "signatory": {
                "givenName": "givenName",
                "familyName": "familyName",
                "additionalName": "additionalName"
            },
            "resolution": "None",
            "signingDate": "2022-08-02",
            "resolutionDate": "2022-08-02",
            "resolutionConfirmed": true
        },
        "alteration": {
            "business": {
                "legalType": "CP"
            },
            "cooperativeAssociationType": "CSC",
            "rulesFileKey": "test",
            "rulesFileName": "testUrl",
            "memorandumFileKey": "test",
            "memorandumFileName": "test"
        },
        "changeOfName": {
            "nameRequest": {
                "legalType": "CP",
                "legalName": "SUPER COOOP COOP",
                "nrNumber": "NR 4192073",
                "expiry": "2022-10-25T06:59:00+00:00",
                "status": "APPROVED",
                "requestType": "CHG",
                "applicant": {
                    "fullName": "BCREGTEST Bena TEST",
                    "fullAddress": "3545 Winter Dr, Ottawa, ON, A5Z 2B8, CA",
                    "phoneNumber": "2506463431",
                    "emailAddress": "hallo@gmail.com"
                }
            }
        }
    }
}
```

severinbeauvais: Yes, exactly -- why are there 3 sub-objects (not including header and business)? I think all other filings have a single sub-object (named according to `header.name`).

severinbeauvais: I mean, this could be "3 filings in one", but I haven't seen this before.

argush3: yes, this is 3 filings in one.  it won't always have 3 filings but it's possible depending on what was changed.

image

argush3: I think the SR filing will always exist along with at least one of CoN or alteration.  Unless it's a change that doesn't require a filing.

severinbeauvais: I see! I definitely tagged the right dev -- thanks Argus!

I see the filing code in this PR always adds the alteration object to the filing. Should it be conditional (like the changeOfName object)?

argush3: yes, alteration should be conditional.

seeker25: if alteration is conditional.. that means the rules and memorandum are optional

severinbeauvais: Is changing rules and memorandum optional for the user?

seeker25: not 100% sure, I think so

We have two more tickets to cover those both.. I think Megan is working on it
We also have another ticket for filing + pay

package.json
```diff
@@ -14,25 +14,25 @@
...
+    "@bcrs-shared-components/folio-number": "1.0.3",
     "@bcrs-shared-components/help-business-number": "1.0.0",
-    "@bcrs-shared-components/interfaces": "1.0.39",
-    "@bcrs-shared-components/nature-of-business": "1.0.8",
-    "@bcrs-shared-components/share-structure": "2.0.0",
-    "@bcrs-shared-components/staff-payment": "1.0.35",
+    "@bcrs-shared-components/interfaces": "1.0.40",
+    "@bcrs-shared-components/nature-of-business": "1.0.10",
+    "@bcrs-shared-components/share-structure": "2.0.1",
+    "@bcrs-shared-components/staff-payment": "2.0.10",
```
severinbeauvais, line 35: Did you test all these?

(I know that some of these are layout changes.)

If it was me, I'd only change the ones I needed (and I'd look at their history and test accordingly).

seeker25: I tested the Major's, the patch versions - I left mostly

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -104,7 +104,7 @@
                               :counter="MAX_RESOLUTION_TEXT_LENGTH"
                               v-model="resolution"
                               :rules="resolutionRules"
-                              @change="onresolutionChanged"
+                              @change="onResolutionChanged"
```
severinbeauvais, line 107: This works but I prefer to be explicit that this is a function call, eg.
`@change="onResolutionChanges($event)`

Doesn't _have_ to be changed, but if you then please also update lines 88, 173, etc.

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -300,23 +297,27 @@ export default class CreateSpecialResolution extends Mixins(DateMixin) {
    * True if date is >= the minimum (ie, today) and <= the maximum (ie, the 10th day).
    * This is used for Vue form validation (in Date Rules above).
    */
-  isBetweenDates (minDate: string, maxDate: string, dateStrToValidate: string): boolean {
+  private isValidDateRange (minDate: Date, maxDate: Date, dateStrToValidate: string): boolean {
     if (!dateStrToValidate) { return true }
-    return (new Date(dateStrToValidate) >= new Date(minDate)) && (new Date(dateStrToValidate) <= new Date(maxDate))
+    // only compare year/month/day (ignore time)
+    const date = new Date(dateStrToValidate) // Input is in the format of MM dd, yyyy
+    return (date >= minDate && date <= maxDate)
```
severinbeauvais, line 304: "minDate" and "maxDate" are correctly in UTC (according to the Date Mixin method used to create them) but I think "date" is set according to the browser's local time offset.

Easy solution: use `yyyyMmDdToDate(dateStrToValidate)`.

src/interfaces/resource-interfaces/sample-form-interface.ts
severinbeauvais: Sorry, I meant the interface name should be specific to its use -- it's not a generic sample form, is it?

Please revert the property name changes.

src/mixins/filing-template-mixin.ts
severinbeauvais: Discussed and resolved in next conversation.

src/mixins/filing-template-mixin.ts
severinbeauvais: Now that I understand that this is a "multi filing", I'm OK with the multiple filing objects.

I leave it to you to make objects conditional as needed.

src/components/SpecialResolution/CreateSpecialResolution.vue
severinbeauvais: I'm OK with the new code.

src/components/SpecialResolution/CreateSpecialResolution.vue
seeker25: For signatory:

minDate is in PST
date is in PST
maxDate arrive in PST
image


For Resolution Date:

minDate is UTC ( doesn't matter it's getting changed )
date is in PST
maxDate is in PST
image

src/mixins/filing-template-mixin.ts
severinbeauvais: Looks good.

src/components/SpecialResolution/CreateSpecialResolution.vue
seeker25: `   const date = this.yyyyMmDdToDate(dateStrToValidate) // Input is in the format of MM dd, yyyy`

results in 

`date: null`

package.json
severinbeauvais: Please add a comment to your ticket for QA to pay special attention to the shared components you updated.

src/components/SpecialResolution/CreateSpecialResolution.vue
seeker25: Changing timeclock on my computer, and retesting date

severinbeauvais: Here are the different types of dates in our UIs:
1. "yyyy-mm-dd" are assumed to be Pacific date
2. "yyyy-mm-ddThh:mm:ss.xxxZ" are assumed to be UTC (ISO format)
3. "yyyy-mm-ddThh:mm:ss.xxx+00.00" are assumed to be UTC (API format)
4. Date objects, you have to careful with
   - if a Date Mixin method was used to create it, it should be in UTC internally (and display according to the browser's timezone)
   - if `Date()` was used to create it, it reverse-offsets according to the browser's timezone and stores internally as "UTC", but you can see here the problem that might occur if the browser's time offset is incorrect/spoofed

severinbeauvais: > ` const date = this.yyyyMmDdToDate(dateStrToValidate) // Input is in the format of MM dd, yyyy`
> 
> results in
> 
> `date: null`

Ah, `yyyMmDdToDate` doesn't support your internal format.

severinbeauvais: PS When you say a Date object is "in PST", I believe it's actually stored as UTC and displayed in your browser's timezone.

severinbeauvais: > Changing timeclock on my computer, and retesting date

Try a different timezone, too.

severinbeauvais:
> I don't see some of the fields were using in the schema (https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/name_request.json), but not sure if we want to add them? Any thoughts?

Don't muck with them if you don't have to. It would be good to have a Tech Debt ticket to go in and clean this up.

src/components/SpecialResolution/CreateSpecialResolution.vue
seeker25: image

severinbeauvais: ^^ using `const date = new Date(dateStrToValidate)` and your workstation/browser set to different timezones?

Explanation: In your example, JS magically knows the correct UTC offset from the string "Aug 30, 2022"... I assume it uses your workstation/browser offset for this, which might differ. However, we want a fixed offset -- always Pacific -- for displayed dates.

seeker25:
https://github.com/bcgov/entity/issues/13445

src/components/SpecialResolution/CreateSpecialResolution.vue
seeker25: dateStrToValidate is coming from `DatePickerShared`

package.json
seeker25: Done, should be in the ticket.

src/interfaces/resource-interfaces/sample-form-interface.ts
seeker25: Ok, i'll fix this up

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -299,9 +299,11 @@ export default class CreateSpecialResolution extends Mixins(DateMixin) {
    */
   private isValidDateRange (minDate: Date, maxDate: Date, dateStrToValidate: string): boolean {
     if (!dateStrToValidate) { return true }
-    // only compare year/month/day (ignore time)
-    const date = new Date(dateStrToValidate) // Input is in the format of MM dd, yyyy
-    return (date >= minDate && date <= maxDate)
+    // Input is in the format of MM dd, yyyy - only compare year/month/day (ignore time)
+    const utcDateStr = new Date(dateStrToValidate + ' 00:00 UTC').toISOString().split('T')[0]
+    const pstDate = this.yyyyMmDdToDate(utcDateStr)
+    console.log(pstDate)
```
severinbeauvais: Does `utcDateStr` in UTC have a time of 00:00 regardless of your workstation's time and time offset?

Does `pstDate` in UTC have a time of 00:00 ?

If so then I believe this will work correctly.

(Please remove console.log.)

src/interfaces/resource-interfaces/special-resolution-sample-form-interface.ts
```diff
@@ -0,0 +1,8 @@
+/** Interface to define the special resolution sample form. */
+
+export interface SpecialResolutionSampleFormIF {
+    header?: string
+    text?: string
+    label?: string
+    path?: string
+}
```
severinbeauvais, line 8: Looks good. For now it's specific to Special Resolutions. If this changes then we can look again what to call this or whatever.

src/components/SpecialResolution/CreateSpecialResolution.vue
severinbeauvais: Closing this thread. This is continued in another conversation below.

src/components/SpecialResolution/CreateSpecialResolution.vue
seeker25: 1. Yes
2.   yyyyMmDdToDate (dateStr: string): Date {
 -  `Converts a date string (YYYY-MM-DD) to a Date object at 12:00:00 am Pacific time.`
 - `   * @example 2021-11-22 -> 2021-11-22T08:00:00.00Z`

severinbeauvais: Then this doesn't work, since you're comparing utcDateStr with time 00:00 with pstDate with time 08:00.

seeker25: This is alot of reinventing the wheel, when a library could easily take care of most of this. 

Luckily for us, I think we're mostly using PST, even though there are technically 3 timezones in BC:

PDT, MST, MDT

severinbeauvais: Did you look at using `createUtcDate()` (in Date Mixin) for creating a Date to compare?

seeker25: No, I just want to get this PR merged

severinbeauvais: I got confused there for a sec.

So `pstDate` is UTC (with Pacific offset) and so are `minDate` and maxDate` ? Then this is fine.

The business decided that displayed dates/times would be in Pacific timezone (PST/PDT) so that's all we have to worry about.

severinbeauvais (approved):
LGTM

@argush3 Any further comments or gtg?

@lewischenstudio Will you be reviewing this PR?

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -300,23 +297,28 @@ export default class CreateSpecialResolution extends Mixins(DateMixin) {
    * True if date is >= the minimum (ie, today) and <= the maximum (ie, the 10th day).
    * This is used for Vue form validation (in Date Rules above).
    */
-  isBetweenDates (minDate: string, maxDate: string, dateStrToValidate: string): boolean {
+  private isValidDateRange (minDate: Date, maxDate: Date, dateStrToValidate: string): boolean {
     if (!dateStrToValidate) { return true }
-    return (new Date(dateStrToValidate) >= new Date(minDate)) && (new Date(dateStrToValidate) <= new Date(maxDate))
+    // Input is in the format of MM dd, yyyy - only compare year/month/day (ignore time)
+    const utcDateStr = new Date(dateStrToValidate + ' 00:00 UTC').toISOString().split('T')[0]
```
seeker25, line 303: I'll revisit this with a comparison to `createUtcDate` for the File and Pay ticket: 

https://github.com/bcgov/entity/issues/13231

There's some backend changes that need to be done in here:
https://github.com/bcgov/entity/issues/13459

To provide the most recent filing date

severinbeauvais: FYI,

"Complex dates" (eg, yyyy-mm-ddThh:mm:ss.xxx+00:00) in the BE are UTC.

"Simple dates" (eg, yyyy-mm-dd) should be Pacific dates, but some might not be.

src/mixins/filing-template-mixin.ts
```diff
@@ -290,24 +292,36 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
...
         legalType: this.getEntitySnapshot.businessInfo.legalType,
-        nrNumber: this.getEntitySnapshot.businessInfo.nrNumber,
-        associationType: this.getAssociationType
+        nrNumber: this.getEntitySnapshot.businessInfo.nrNumber
       },
       specialResolution: {
-        ...this.getcreateResolution
+        ...this.getSpecialResolution
+      },
+      alteration: {
```
argush3, line 300: so I think this should be conditional like change of name.  for example, if there was only a change of name, then you would have a SR + CoN filing.  Is this taken care of already and I'm just seeing it?

seeker25: I've made it conditional

argush3 (approved):

PR #372, opened 2022-09-07
12690 Edit UI: Conversion filing - remove Business contact component

lewischenstudio:
*Issue #:* /bcgov/entity#12690

*Description of changes:*
- conditionally removed component `BusinessContactInfo`
- updated validation
- updated the build method and the parse method

### DIFF
--- a/src/components/common/YourCompany/YourCompany.vue
+++ b/src/components/common/YourCompany/YourCompany.vue
@@ -281,14 +281,16 @@
       <OfficeAddresses :invalidSection="invalidAddressSection" />
     </div>
 
-    <v-divider class="mx-4 my-1" />
-
     <!-- Business Contact Information -->
-    <div id="contact-info-section" class="section-container" :class="{'invalid-section': invalidContactSection}">
-      <BusinessContactInfo
-        :invalidSection="invalidContactSection"
-      />
-    </div>
+    <template v-if="!isFirmConversionFiling">
+      <v-divider class="mx-4 my-1" />
+
+      <div id="contact-info-section" class="section-container" :class="{'invalid-section': invalidContactSection}">
+        <BusinessContactInfo
+          :invalidSection="invalidContactSection"
+        />
+      </div>
+    </template>
 
     <!-- Folio Information (all except SP or GP) -->
     <template v-if="isPremiumAccount && !isEntityTypeFirm">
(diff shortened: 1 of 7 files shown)


seeker25 (approved):
LGTM

lambert-alex (approved):
LGTM

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -281,14 +281,16 @@
...
 
-    <v-divider class="mx-4 my-1" />
-
     <!-- Business Contact Information -->
-    <div id="contact-info-section" class="section-container" :class="{'invalid-section': invalidContactSection}">
-      <BusinessContactInfo
-        :invalidSection="invalidContactSection"
-      />
-    </div>
+    <template v-if="!isFirmConversionFiling">
```
severinbeauvais, line 285: I think it will be better for us in the future if we INCLUDE the filings this applies to instead of EXCLUDING the fillings it doesn't.

Otherwise, this looks great.

severinbeauvais (approved):
Looks good but please change that logic., and then you can merge it.

src/components/common/YourCompany/YourCompany.vue
lewischenstudio: This component applies to everything else except for `isFirmConversionFiling`. If we want to include the filings this applies, then the code will be as long as this: `v-if="isAlterationFiling || isFirmChangeFiling || isFirmCorrectionFiling || isSpecialResolutionFiling || isBenCorrectionFiling"`, given that I didn't miss anything here. I find it too long, so I used the exclusion case. I agree to use inclusion case for many scenarios, but I think it looks better in this special case. What do you think?

severinbeauvais: From experience, I can tell you it is easier to understand and manage if we include instead of exclude.

You can create a local getter to capture the logic, eg, `get showBusinessContactInformation (): boolean ...`.

lewischenstudio: Yeah. That's a good point. Let me add this local getter.

lewischenstudio: I have added `showBusinessContactInformation` in my latest commit. I just noticed there is `showChangeBusinessType`, so this approach "coincide" with the existing local getter pretty well (I couldn't find a better word for it 😆).

PR #379, opened 2022-09-14
13553 Edit UI: BEN IA correction should prompt for Completing Party

lewischenstudio:
*Issue #:* /bcgov/entity#13553

*Description of changes:*
- add Completing Party component (for Client Error only)
    - place above Detail component
- unit tests

### DIFF
--- a/src/views/Correction/BenCorrection.vue
+++ b/src/views/Correction/BenCorrection.vue
@@ -27,6 +27,10 @@
 
     <Articles class="mt-10" />
 
+    <template v-if="isClientErrorCorrection">
+      <CompletingParty class="mt-10" sectionNumber="1." validate="true" />
+    </template>
+
     <Detail
       class="mt-10"
       sectionNumber="1."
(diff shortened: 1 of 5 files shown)


src/views/Correction/BenCorrection.vue
```diff
@@ -27,6 +27,10 @@
 
     <Articles class="mt-10" />
 
+    <template v-if="isClientErrorCorrection">
+      <CompletingParty class="mt-10" sectionNumber="1." validate="true" />
```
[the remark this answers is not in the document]
lewischenstudio: Thanks. It's now fixed. Same as in `FmCorrection`.

lambert-alex (approved):
LGTM

src/views/Correction/FmCorrection.vue
```diff
@@ -28,7 +28,7 @@
 
     <StaffPayment
       class="mt-10"
-      :sectionNumber="isClientErrorCorrection ? '4.' : '2.'"
+      :sectionNumber="isClientErrorCorrection ? '4.' : '3.'"
```
severinbeauvais, line 31: I think the previous code was actually correct: if this is not a client error correction then both Completing Party and Certify Section are not rendered, so StaffPayment is indeed the second section.

severinbeauvais: I also wondering why the Certify Section is not displayed here but it IS displayed in a BEN correction? Please confirm with Mihai.

lewischenstudio: You are right. Certify will be displayed for client error, so StaffPayment is the second section. I will change it back in my this PR: https://github.com/bcgov/business-edit-ui/pull/380. I will confirm with Mihai during today's standup and add the condition to Certify if we want to hide Certify for staff error for BEN IA correction.

PR #384, opened 2022-10-04
13723 Edit UI - Review and Confirm step shows missing required info when there is none missing

lewischenstudio:
*Issue #:* /bcgov/entity#13723

*Description of changes:*
- make truthy special resolution validation flag in `parseChangeRegFiling`

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -93,7 +93,6 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
   @Action setHasPlanOfArrangement!: ActionBindingIF
   @Action setSpecialResolution!: ActionBindingIF
   @Action setCorrectionStartDate!: ActionBindingIF
-  @Action setSpecialResolutionConfirmStateValidity!: ActionBindingIF
 
   /** The default (hard-coded first line) correction detail comment. */
   public get defaultCorrectionDetailComment (): string {
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -93,6 +93,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
   @Action setHasPlanOfArrangement!: ActionBindingIF
   @Action setSpecialResolution!: ActionBindingIF
   @Action setCorrectionStartDate!: ActionBindingIF
+  @Action setSpecialResolutionConfirmStateValidity!: ActionBindingIF
 
   /** The default (hard-coded first line) correction detail comment. */
   public get defaultCorrectionDetailComment (): string {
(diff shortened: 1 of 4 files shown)


lambert-alex (commented):
When I try this locally it does not fix the problem. Also could you add a version number please?

src/mixins/filing-template-mixin.ts
```diff
@@ -866,6 +867,9 @@ export default class FilingTemplateMixin extends Mixins(DateMixin, EnumMixin) {
     // store Business Start Date
     this.setCorrectionStartDate(filing.changeOfRegistration.startDate || null)
 
+    // default this flag to true for change filing
+    this.setSpecialResolutionConfirmStateValidity(true)
```
severinbeauvais, line 871: How about just initializing this flag to True in state-models.ts (like the other ReviewCertify flags) ?  If there is a special reason this flag needs to be initialized as False then it should be its component's responsibility to do that.

Also, please update the app version :)

lewischenstudio: I think your way is better. This flag is only used in the `CreateSpecialResolutionSummary`, which sets it to false onload, and initialize it to true in the state model works.

lewischenstudio:
> When I try this locally it does not fix the problem. Also could you add a version number please?

This was working on my local env, but I updated the repo with the new initialization. Can you pull and try again? App version is now added.

lambert-alex:
> > When I try this locally it does not fix the problem. Also could you add a version number please?
> 
> This was working on my local env, but I updated the repo with the new initialization. Can you pull and try again? App version is now added.

I tried it again and I can file now! I do get a unable to save filing error I think that might be unrelated to your fix.

Screenshot 2022-10-04

lewischenstudio:
> > > When I try this locally it does not fix the problem. Also could you add a version number please?
> > 
> > 
> > This was working on my local env, but I updated the repo with the new initialization. Can you pull and try again? App version is now added.
> 
> I tried it again and I can file now! I do get a unable to save filing error I think that might be unrelated to your fix.
> 

This is most likely a schema error. Can you test the change filing again with another SP/GP business number? I'm not entirely sure what causes this, and I was able to file a change filing.

lambert-alex:
> > > > When I try this locally it does not fix the problem. Also could you add a version number please?
> > > 
> > > 
> > > This was working on my local env, but I updated the repo with the new initialization. Can you pull and try again? App version is now added.
> > 
> > 
> > I tried it again and I can file now! I do get a unable to save filing error I think that might be unrelated to your fix.
> 
> This is most likely a schema error. Can you test the change filing again with another SP/GP business number? I'm not entirely sure what causes this, and I was able to file a change filing.

Yes it is a schema error, I figured out the error I guess Province is a required field in the schema even though it is optional

lambert-alex (approved):

PR #386, opened 2022-10-06
13554/13776 Edit UI: Add Documents Delivery to miscellaneous filings

lewischenstudio:
*Issue #:* /bcgov/entity#13554 and /bcgov/entity#13776

*Description of changes:*
- editable business contact information
- updated schema for correction
- updated correction filing validation getter
- preload completing party for firm corrections

### DIFF
--- a/src/components/common/YourCompany/BusinessContactInfo.vue
+++ b/src/components/common/YourCompany/BusinessContactInfo.vue
@@ -3,7 +3,7 @@
     :contactLabel="getResource.contactLabel"
     :businessContact="getBusinessContact"
     :originalBusinessContact="originalContact"
-    :hasBusinessContactInfoChange="hasContactInfoChanged"
+    :hasBusinessContactInfoChange="hasBusinessContactInfoChange"
     :editLabel="editLabel"
     :editedLabel="editSavedLabel"
     :disableActionTooltip="isFirmChangeFiling"
@@ -35,7 +35,6 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
   @Getter getEntitySnapshot!: EntitySnapshotIF
   @Getter getResource!: ResourceIF
   @Getter getBusinessId!: string
-  @Getter hasContactInfoChanged!: boolean
 
   // Global setters
   @Action setBusinessContact!: ActionBindingIF
@@ -51,6 +50,13 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
     return this.getEntitySnapshot?.authInfo?.contact
   }
 
+  /** True if there are changes between current Contact and original Contact Info. */
+  get hasBusinessContactInfoChange (): boolean {
+    return this.getBusinessContact?.email !== this.originalContact?.email ||
+      this.getBusinessContact?.phone !== this.originalContact?.phone ||
+      this.getBusinessContact?.extension !== this.originalContact?.extension
+  }
+
   /** On Contact Info Change event, updates auth db and store. */
   protected async onContactInfoChange (contactInfo: ContactPointIF): Promise<void> {
     // do nothing if contact info was not changed
@@ -60,6 +66,7 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
       if (
         this.isAlterationFiling ||
         this.isFirmChangeFiling ||
+        this.isCorrectionFiling ||
         this.isSpecialResolutionFiling
       ) {
         await AuthServices.updateContactInfo(contactInfo, this.getBusinessId)
--- a/src/components/common/YourCompany/BusinessContactInfo.vue
+++ b/src/components/common/YourCompany/BusinessContactInfo.vue
@@ -3,10 +3,9 @@
     :contactLabel="getResource.contactLabel"
     :businessContact="getBusinessContact"
     :originalBusinessContact="originalContact"
-    :hasBusinessContactInfoChange="hasBusinessContactInfoChange"
+    :hasBusinessContactInfoChange="hasContactInfoChanged"
     :editLabel="editLabel"
     :editedLabel="editSavedLabel"
-    :disableActions="isCorrectionFiling"
     :disableActionTooltip="isFirmChangeFiling"
     :invalidSection="invalidSection"
     :optionalPhone="isAlterationFiling || isFirmChangeFiling"
@@ -36,6 +35,7 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
   @Getter getEntitySnapshot!: EntitySnapshotIF
   @Getter getResource!: ResourceIF
   @Getter getBusinessId!: string
+  @Getter hasContactInfoChanged!: boolean
 
   // Global setters
   @Action setBusinessContact!: ActionBindingIF
@@ -49,13 +49,6 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
   /** The original Contact data. */
   get originalContact (): ContactPointIF {
     return this.getEntitySnapshot?.authInfo?.contact
-  }
-
-  /** True if there are changes between current Contact and original Contact Info. */
-  get hasBusinessContactInfoChange (): boolean {
-    return this.getBusinessContact?.email !== this.originalContact?.email ||
-      this.getBusinessContact?.phone !== this.originalContact?.phone ||
-      this.getBusinessContact?.extension !== this.originalContact?.extension
   }
 
   /** On Contact Info Change event, updates auth db and store. */
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -126,6 +126,7 @@ export default class FilingTemplateMixin extends DateMixin {
         nrNumber: this.getNameRequestNumber
       },
       correction: {
+        legalType: this.getEntityType,
         business: {
           identifier: this.getBusinessId
         },
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -386,7 +386,8 @@ export const hasCorrectionDataChanged = (state: StateIF): boolean => {
       havePeopleAndRolesChanged(state) ||
       hasShareStructureChanged(state) ||
       areProvisionsRemoved(state) ||
-      haveNewResolutionDates(state)
+      haveNewResolutionDates(state) ||
+      hasContactInfoChanged(state)
     )
   }
 
@@ -395,9 +396,9 @@ export const hasCorrectionDataChanged = (state: StateIF): boolean => {
       hasBusinessNameChanged(state) ||
       hasBusinessStartDateChanged(state) ||
       hasNaicsChanged(state) ||
-      hasBusinessStartDateChanged(state) ||
       haveOfficeAddressesChanged(state) ||
-      havePeopleAndRolesChanged(state)
+      havePeopleAndRolesChanged(state) ||
+      hasContactInfoChanged(state)
     )
   }
 
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -386,8 +386,7 @@ export const hasCorrectionDataChanged = (state: StateIF): boolean => {
       havePeopleAndRolesChanged(state) ||
       hasShareStructureChanged(state) ||
       areProvisionsRemoved(state) ||
-      haveNewResolutionDates(state) ||
-      hasContactInfoChanged(state)
+      haveNewResolutionDates(state)
     )
   }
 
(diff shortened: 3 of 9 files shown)


src/components/common/YourCompany/BusinessContactInfo.vue
```diff
@@ -6,7 +6,6 @@
     :hasBusinessContactInfoChange="hasBusinessContactInfoChange"
     :editLabel="editLabel"
     :editedLabel="editSavedLabel"
-    :disableActions="isCorrectionFiling"
```
severinbeauvais, line 9: Is this for ticket 13776?

lambert-alex: Yes it looks like it, I just did it for 13776

src/mixins/filing-template-mixin.ts
```diff
@@ -589,7 +590,7 @@ export default class FilingTemplateMixin extends DateMixin {
     ))
 
     // store current Business Contact
-    this.setBusinessContact({ ...entitySnapshot.authInfo.contact })
+    this.setBusinessContact(filing.correction.contactPoint)
```
severinbeauvais, line 593: The other parsing functions in here get the business contact from the entity snapshot, which I think is OK here, too.
Basically, don't restore the business contact from the contact point object. It's special in that you always want to look at the latest data (not from a draft filing).

src/components/common/YourCompany/BusinessContactInfo.vue
lewischenstudio: Ticket 13776 overlaps with this ticket, and I mentioned it in planning yesterday.

src/mixins/filing-template-mixin.ts
severinbeauvais: Also this code might fail on the initial correction draft since it probably doesn't contain a contact point object.

lewischenstudio: There is this problem for contactPoint. We need to send the contactPoint json to `AuthServices.updateContactInfo` API when we update it in the component, which also updates the entitysnapshot. Then it can't tell whether contactPoint is changed or not. So I use `filing.correction` as the reference here.

src/views/Correction/FmCorrection.vue
```diff
@@ -97,6 +103,21 @@ export default class FmCorrection extends Mixins(CommonMixin, FeeMixin, FilingTe
...
+        mailingAddress: {
+          addressCity: mailingAddress?.addressCity,
+          addressCountry: mailingAddress?.addressCountry,
+          addressRegion: mailingAddress?.addressRegion,
+          postalCode: mailingAddress?.postalCode,
+          streetAddress: mailingAddress?.streetAddress,
+          streetAddressAdditional: mailingAddress?.streetAddressAdditional
+        }
+      } as CompletingPartyIF)
+
```
severinbeauvais, line 120: First, I think it's a risky assumption that the completing party's address is the same as the business office.

Second, the other filings that manage completing party data don't do this.

Why are you doing this?

src/mixins/filing-template-mixin.ts
lewischenstudio: I can add a condition to use entitysnapshot if it doesn't contain a contact point object, but this contact point is mandatory according to schema. I'm working on it.

src/views/Correction/FmCorrection.vue
lewischenstudio: This is part of the ticket's requirement: "For corrections triggered by clients, need to display the Completing Party, name, email and address, editable, pre-populated with the logged in user values.". Maybe I am missing something here.

src/components/common/YourCompany/BusinessContactInfo.vue
severinbeauvais: How are you both working on the same ticket? You should always put your name on the tickets you're working on.

lambert-alex: I'll close 13776

src/views/Correction/FmCorrection.vue
severinbeauvais: How do the other filings in this UI handle this requirement?

lambert-alex: Will this not prefill for both client triggered corrections and staff triggered corrections?

lewischenstudio: The other filings in this UI doesn't render pre-populated completing party with the logged in user values.

src/components/common/YourCompany/BusinessContactInfo.vue
lambert-alex: > How are you both working on the same ticket? You should always put your name on the tickets you're working on, to prevent this.

Different tickets but they had an overlap with each other

src/mixins/filing-template-mixin.ts
severinbeauvais: I think business contact needs to be fetched from snapshot every time -- never from a draft. Otherwise someone could have updated the business contact separately and then this info would be wrong.

PS Watch you don't confuse business contact (object stored in auth db) and contact point (object in the correction filing).

What I noticed here is that this filing handles things differently from the other filings, and that's unexpected.

src/views/Correction/FmCorrection.vue
severinbeauvais: Why doesn't this code prefill things? Maybe it needs to be fixed.

https://github.com/bcgov/business-edit-ui/blob/4b3892439710e413f6e0cb98d29b27d8bd9b8cdb/src/App.vue#L534

src/mixins/filing-template-mixin.ts
lewischenstudio: Okay. I will load it from the entity snapshot, but there will be a minor problem when it comes to saving the draft (unless we don't include `this.isCorrectionFiling` in `onContactPointChange()` in `BusinessContactInfo.vue`). This is issue and why I wanted to use the filing.

src/components/common/YourCompany/BusinessContactInfo.vue
```diff
@@ -67,6 +65,7 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
       if (
         this.isAlterationFiling ||
         this.isFirmChangeFiling ||
+        this.isCorrectionFiling ||
```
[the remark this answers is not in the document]
lewischenstudio: If we include `this.isCorrectionFiling` here, then UI will not detect contact point being changed from resuming a draft.

src/mixins/filing-template-mixin.ts
severinbeauvais: What's the problem with saving the draft?

src/views/Correction/FmCorrection.vue
lewischenstudio: Yeah, this code doesn't load completing party.

severinbeauvais: Does it work in other filings and is only broken for correction filings? Or only when staff?

src/store/getters/state-getters.ts
```diff
@@ -386,7 +386,8 @@ export const hasCorrectionDataChanged = (state: StateIF): boolean => {
       havePeopleAndRolesChanged(state) ||
       hasShareStructureChanged(state) ||
       areProvisionsRemoved(state) ||
-      haveNewResolutionDates(state)
+      haveNewResolutionDates(state) ||
+      hasContactInfoChanged(state)
```
severinbeauvais, line 390: Remove this (and on line 401 below).

Change the business contact info does not create a correction filing.

src/mixins/filing-template-mixin.ts
lewischenstudio: Changing it back. Sorry for the confusion.

src/store/getters/state-getters.ts
lewischenstudio: Removing it.

src/views/Correction/FmCorrection.vue
lewischenstudio: This doesn't work with other filings either. This code doesn't load anything into `setCompletingParty`, and I think we can ask about it during tomorrow's standup.

severinbeauvais: Does this code work if you're logged in as not staff?

lewischenstudio: Let me try. Hold on.

severinbeauvais: I don't know where we'd get the name to pre-populate if we are logged in as staff -- a good question to ask tomorrow.

lewischenstudio: I logged in as a client user and got it tested on a change filing. It actually loads the pre-populated values for the completing party component, which is not editable for client users.

severinbeauvais: OK, so we need to figure out what we want when logged in as staff (and whether we can do it).

lewischenstudio: I think I didn't read this line `if (!this.isRoleStaff)` in the code below:
https://github.com/bcgov/business-edit-ui/blob/4b3892439710e413f6e0cb98d29b27d8bd9b8cdb/src/App.vue#L665-L669

You're right. I'll mention it during the standup.

lewischenstudio: > Will this not prefill for both client triggered corrections and staff triggered corrections?

The code is now restored/removed, but that code will prefill for either client or staff triggered corrections. We'll see whether we should prefill correction filing for users logged in as staff.

lewischenstudio:
I suggest that we don't merge this ticket until we figure out the prefill condition for completing party, hopefully tomorrow during standup.

src/store/getters/state-getters.ts
```diff
@@ -395,7 +395,6 @@ export const hasCorrectionDataChanged = (state: StateIF): boolean => {
       hasBusinessNameChanged(state) ||
       hasBusinessStartDateChanged(state) ||
       hasNaicsChanged(state) ||
-      hasBusinessStartDateChanged(state) ||
```
severinbeauvais, line 398: Why did you remove this?

lewischenstudio: It's a duplicate from line 396.

src/views/Correction/FmCorrection.vue
severinbeauvais: Wait - "client or staff triggered corrections" can only be done by staff. (I'm not sure if "SBC Staff" will ever do corrections, but if they do then they have non-idir login so this code may inadvertently use their info. This is another question to ask.)

Does this mean we cannot prefill any of this data for a correction?

src/store/getters/state-getters.ts
severinbeauvais: Good catch :)

src/views/Correction/FmCorrection.vue
lewischenstudio: We currently cannot prefill any of this for
1. a correction filing
2. any other type of filings, i.e. change filing, if logged in as a staff user.

lewischenstudio: It was confirmed from Sarah that from the business point of view, completing party in any type of filing should be left empty if the logged in user is not the owner/account user. So in our case, completing party is not pre-filled for staff users, and the current/original implementation is correct.

src/components/common/YourCompany/BusinessContactInfo.vue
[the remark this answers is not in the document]
severinbeauvais: > For consistency, could you please update YourCompany.vue::showBusinessContactInformation() with the same conditions?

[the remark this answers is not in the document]
lewischenstudio: I missed this spot, and it's now fixed.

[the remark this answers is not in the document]
severinbeauvais: Thanks, looks good.

tests/unit/state-getters.spec.ts
```diff
@@ -357,6 +363,10 @@ describe('SP/GP correction getters', () => {
       },
       shareStructure: {
         shareClasses: []
+      },
+      authInfo: {
+        contact: {},
+        folioNumber: ''
```
severinbeauvais, line 369: Do we still need these changes?

severinbeauvais (approved):
LGTM

tests/unit/state-getters.spec.ts
lewischenstudio: It was used earlier, but not anymore. They are now reverted.

PR #390, opened 2022-10-07
sp/gp sbc flow - make completing party editable

kialj876:
Signed-off-by: Kial Jinnah <kialjinnah@gmail.com>

*Issue #:* /bcgov/entity#13822

*Description of changes:*

### DIFF
--- a/src/components/common/CompletingParty.vue
+++ b/src/components/common/CompletingParty.vue
@@ -44,6 +44,7 @@ export default class CompletingParty extends Vue {
   // store getters
   @Getter getCompletingParty!: CompletingPartyIF
   @Getter isRoleStaff!: boolean
+  @Getter isSbcStaff!: boolean
 
   // store actions
   @Action setCompletingParty!: ActionBindingIF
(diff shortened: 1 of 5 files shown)


src/components/common/CompletingParty.vue
```diff
@@ -58,6 +60,18 @@ export default class CompletingParty extends Vue {
...
+  get isSbcStaff (): boolean {
+    const currentAccount = sessionStorage.getItem(SessionStorageKeys.CurrentAccount)
+    if (!currentAccount) return false
+    try {
+      return JSON.parse(currentAccount).accountType === AccountTypes.STAFF_SBC
+    } catch (error) {
+      console.log('Error parsing current account =', error)
+      return false
+    }
+  }
```
severinbeauvais, line 73: Most of this code is already in App.vue

I think that all you need is a new getter like this in state-getters.ts:
```
/** Whether the current account is SBC Staff. */
export const isPremiumAccount = (state: StateIF): boolean => {
  return (state.stateModel.accountInformation.accountType === AccountTypes.SBC_STAFF)
}
```

src/components/common/CompletingParty.vue
```diff
@@ -58,6 +60,18 @@ export default class CompletingParty extends Vue {
     return (this.validate && !this.completingPartyValid)
   }
 
+  // FUTURE: pull this from shared auth composable (after vue3 upgrade)
+  get isSbcStaff (): boolean {
+    const currentAccount = sessionStorage.getItem(SessionStorageKeys.CurrentAccount)
+    if (!currentAccount) return false
+    try {
+      return JSON.parse(currentAccount).accountType === AccountTypes.STAFF_SBC
```
severinbeauvais, line 68: Weird that you're not getting an error here with the wrong AccountTypes enum (STAFF_SBC vs SBC_STAFF).

src/components/common/CompletingParty.vue
severinbeauvais: https://github.com/bcgov/business-edit-ui/blob/e1db511ab234c9b6a4abbd27f180c270d502c432/src/App.vue#L713

kialj876: ah nice okay will update

src/components/common/CompletingParty.vue
kialj876: ah I switched the enum name name after testing it in the UI (I had it different in the create ui)

severinbeauvais (approved):
LGTM

PR #397, opened 2022-10-19
Update NOB shared component

lambert-alex:
*Issue #:* /bcgov/entity#13875

*Description of changes:*
 - Reduce Tech debt update NOB shared component from 1.1.1 -> 1.2.0
 - Tested with Conversion and change filings (two places the edit UI uses NOB)
 - version 3.8.8

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-edit-ui",
-  "version": "3.8.7",
+  "version": "3.8.8",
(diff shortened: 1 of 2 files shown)


severinbeauvais (approved):
So it works fine on initial load?

lambert-alex:
> So it works fine on initial load?

Yes it work the emit triggers the onValidchanged and sets the valid component.

PR #400, opened 2022-10-20
13924 Validation and state changes to SP/GP filings

lambert-alex:
*Issue #:* /bcgov/entity#13924

*Description of changes:*


src/enums/componentsReviewCertify/alterationReviewCertify.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
lambert-alex, line 1: The Each of the filings now have their own review certify enums

src/interfaces/store-interfaces/state-interfaces/flags-review-certify-interfaces/alteration-flags-review-certify.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
lambert-alex, line 1: Each of the flags are now separated preventing incorrect scroll behavior for incomplete filings

src/store/actions/actions-model.ts
```diff
@@ -70,32 +70,76 @@ export const setCertifyState: ActionIF = ({ commit }, certifyState: CertifyIF):
   commit('mutateCertifyState', certifyState)
 }
 
-export const setCertifyStateValidity: ActionIF = ({ commit }, validity): void => {
-  commit('mutateCertifyStateValidity', validity)
+export const setAlterationCertifyStateValidity: ActionIF = ({ commit }, validity): void => {
```
lambert-alex, line 73: A down side of this is now the state is larger and there is lots of code repetition because each of the filings need their own mutations and actions instead of having it more generalized

src/store/getters/state-getters.ts
```diff
@@ -503,15 +490,26 @@ export const hasConversionDataChanged = (state: StateIF): boolean => {
...
-    return (
-      getFlagsCompanyInfo(state).isValidCompanyName &&
-      getFlagsCompanyInfo(state).isValidNameTranslation &&
-      getFlagsCompanyInfo(state).isValidOrgPersons &&
-      getFlagsCompanyInfo(state).isValidShareStructure &&
-      getFlagsReviewCertify(state).isValidDetailComment &&
-      getFlagsReviewCertify(state).isValidCertify &&
-      getFlagsReviewCertify(state).isValidStaffPayment
-    )
+    if (isClientErrorCorrection(state)) {
```
lambert-alex, line 493: I separated Ben corrections into staff and client similar to how firm corrections are.

src/store/mutations/mutations-model.ts
```diff
@@ -60,44 +60,88 @@ export const mutateEffectiveDateTimeString = (state: StateIF, dateTime: string)
...
 }
 
 export const mutateCertifyState = (state: StateIF, certifyState: CertifyIF) => {
   state.stateModel.certifyState = certifyState
   if (!state.stateModel.tombstone.ignoreChanges) mutateHaveUnsavedChanges(state, true)
 }
 
-export const mutateCertifyStateValidity = (state: StateIF, validity: boolean) => {
-  state.stateModel.validationFlags.flagsReviewCertify.isValidCertify = validity
+export const mutateAlterationCertifyStateValidity = (state: StateIF, validity: boolean) => {
```
lambert-alex, line 72: Same issues the corresponding actions

src/store/state/state-model.ts
```diff
@@ -50,17 +50,55 @@ export const stateModel: StateModelIF = {
       isValidAssociationType: true,
       isValidCreateSpecialResolution: true
     },
-    flagsReviewCertify: {
```
lambert-alex, line 53: flagsReviewCertify is now deleted and has been replaced with smaller a state models that are specific to each filing

src/mixins/common-mixin.ts
```diff
@@ -38,16 +38,15 @@ export default class CommonMixin extends Vue {
   async validateAndScroll (flags: object, components: object): Promise<boolean> {
     // Create an array of the _ordered_ validity flags
     const validFlagArray = Object.keys(flags).map(key => flags[key])
-
     // Find the _first_ corresponding component that is invalid
     const component = document.getElementById(components[validFlagArray.indexOf(false)])
 
     // If there is an invalid component, scroll to it
     if (component) {
```
lambert-alex, line 45: What was happening before is there would be an incomplete component on a filing for example take a specialResolution with an invalid component certify, since flags-review-certify-interface.ts was general the order
would set  validFlagArray would contain a false component however it would set the detail component as the false one then this line would not evaluate to true as there is no detail component on a specialResolution. So true would be returned.

src/mixins/common-mixin.ts
```diff
@@ -38,16 +38,15 @@ export default class CommonMixin extends Vue {
...
       await this.scrollToTop(component)
-      return false
     }
-    return true
+    // If any componets are not valid, then validation fails.
+    return !validFlagArray.includes(false)
```
lambert-alex, line 49: Moving return false and having evaluate here separately would be a quick fix that would partially fix the issue. Needing no other changes

lambert-alex:
I am looking for input on what I should do. Right now I am stuck because the way I set up the store each filing will need to set validity separately, currently the individual components set validity because it was common between the filings before.

I was thinking of setting components as valid in the views.

lewischenstudio:
I am considering whether there is a simple solution to this issue. The initial state is set by [`flagsReviewCertify`](https://github.com/bcgov/business-edit-ui/blob/e697d718768da1515bcd1b832673110d0396af85/src/store/state/state-model.ts#L53-L63), and each component will set that state to false upon rendering the step for the first time. Is it possible that some component didn't update the relative key-value in this state object `flagsReviewCertify` even though the component is valid and finished?

src/mixins/common-mixin.ts
severinbeauvais: I'm still thinking that it's OK for un-rendered components to be in the array as long as their value is true (valid), which is how the array is initialized in state-model.ts.

The question is, which un-rendered component is invalid? Maybe we should fix that problem instead.

(Is there another issue with the components being in different orders? That shouldn't matter much.)

src/store/actions/actions-model.ts
severinbeauvais: If we go this route then I'd like to see it generalized again ... possibly using the list of components for each filing in their resource file?

src/store/getters/state-getters.ts
```diff
@@ -448,21 +450,6 @@ export const hasSpecialResolutionDataChanged = (state: StateIF): boolean => {
   )
 }
 
-/** Whether the special resolution filing is valid. Does NOT include:
```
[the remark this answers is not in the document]
severinbeauvais: ^^ Actually, it depends on whether SRs use the correction-type validation or the list-type validation.

src/store/getters/state-getters.ts
severinbeauvais: I think I already did this in yesterday's PR.

severinbeauvais (commented):
On the plus side, you now have a great view of our validation architecture 🥴

src/store/getters/state-getters.ts
lambert-alex: Right I think you did, I haven't update my local in a day or two.

src/store/actions/actions-model.ts
lambert-alex: Okay that makes sense. Sort of like using any interface but then having the code generalized in the store?

src/store/getters/state-getters.ts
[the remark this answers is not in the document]
lambert-alex: SRs are using list-type validation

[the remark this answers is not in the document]
severinbeauvais: Then you're right: this getter is not needed.

lambert-alex:
This issue was fixed in PR [401](https://github.com/bcgov/business-edit-ui/pull/401)

PR #408, opened 2022-10-31
13896 Pages should gray out after clicking Save and Resume Later and File Now, to avoid multiple clicks

lambert-alex:
*Issue #:* /bcgov/entity#13896

*Description of changes:*

Note to consistently see this error you have to set throttling in dev tools

 - Upon clicking Save and Resume and File Now removed setSaving to false because if the page is slow to redirect then the buttons become reenabled.
 - App version = 3.8.13 -> 3.8.14
 - Replated PR [Create UI](https://github.com/bcgov/business-create-ui/pull/463)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -825,10 +825,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
         const error = new Error('Missing Payment Token or Filing ID')
         this.$root.$emit('save-error-event', error)
       }
-      this.setIsFilingPaying(false)
     }
-
-    this.setIsSaving(false)
   }
 }
 </script>
--- a/src/App.vue
+++ b/src/App.vue
@@ -801,6 +801,8 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
     // if filing is not a draft, proceed with payment
     if (!isDraft && filingComplete) {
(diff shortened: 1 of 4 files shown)


src/App.vue
```diff
@@ -825,10 +825,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
         const error = new Error('Missing Payment Token or Filing ID')
         this.$root.$emit('save-error-event', error)
       }
-      this.setIsFilingPaying(false)
```
lambert-alex, line 828: Upon loading a page isFilingPaying, isSaving and isSavingResuming are false so we don't need to set them to false for situations where we leave the page.

severinbeauvais: I'm thinking you should add your comment to the code, so that it doesn't look like someone forgot to add the line.

Also in Actions.vue.

lewischenstudio (approved):
LGTM

src/App.vue
lambert-alex: Will do!

ChrisGabel23 (approved):

severinbeauvais (approved):
Looks good. Thanks for the comment updates.

src/App.vue
lambert-alex: Will do

PR #410, opened 2022-11-04
Feature 13968 Certify block must be editable for SBC staff on the cha…

ritvick:
…nge filing

*Issue #:* /bcgov/entity/issues/13968

*Description of changes:*
Certify block must be editable for SBC staff on the change filing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-edit-ui",
-  "version": "3.8.14",
+  "version": "3.8.15",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "business-edit-ui",
-      "version": "3.8.14",
+      "version": "3.8.15",
       "dependencies": {
         "@babel/compat-data": "^7.19.1",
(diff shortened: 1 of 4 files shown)


ritvick:
@severinbeauvais / other devs,  is this the right place to make this change? or shall we have another parameter passed to Certify component isSbcStaff, and add the logic inside that shared component?
What do you think?

severinbeauvais (approved):
Looks good but please update the app version.

severinbeauvais:
> @severinbeauvais / other devs, is this the right place to make this change? or shall we have another parameter passed to Certify component isSbcStaff, and add the logic inside that shared component? What do you think?

I think this is fine. At the moment, the Certify component doesn't need to know what kind of staff it is, so just using a single prop is fine. It's good to keep each component as independent/dumb/simple as possible :)

lambert-alex (approved):
LGTM

lambert-alex (approved):

severinbeauvais (commented):
This code change looks fine to me, but please update the app version in package.json and then run `npm i` and commit both package files.

severinbeauvais (approved):

PR #411, opened 2022-11-07
Folio Number bug fix

lekshmimallika-aot:
*Issue #:* /bcgov/entity#13871

*Description of changes:*

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -1137,9 +1137,10 @@ export default class FilingTemplateMixin extends DateMixin {
   private buildFolioNumber (filing: AlterationFilingIF | ChgRegistrationFilingIF | SpecialResolutionFilingIF): any {
     const fn = this.getFolioNumber
     const tfn = this.getTransactionalFolioNumber
+    filing.header.folioNumber = fn
 
     // if a Transactional Folio Number was entered then override the Folio Number
-    if (tfn !== fn) {
+    if (tfn && tfn !== fn) {
       filing.header.folioNumber = tfn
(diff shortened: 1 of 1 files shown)


severinbeauvais (commented):
Too easy! :)

Please also update the app version and commit the package files.

src/mixins/filing-template-mixin.ts
```diff
@@ -1137,9 +1137,10 @@ export default class FilingTemplateMixin extends DateMixin {
   private buildFolioNumber (filing: AlterationFilingIF | ChgRegistrationFilingIF | SpecialResolutionFilingIF): any {
     const fn = this.getFolioNumber
     const tfn = this.getTransactionalFolioNumber
+    filing.header.folioNumber = fn
```
severinbeauvais, line 1140: Why is this needed? I think it's already set in the "build" functions above.

lekshmimallika-aot: IMO since this functions responsibility is building the folio Number it should happen here

severinbeauvais: Well now it's in two places.

lekshmimallika-aot: I see that folio number is set from different places like staff payment, in the beginning of building a filing etc. If this method is supposed to set only transactional folio number it would be intuitive if the method name is like buildTransactionalFolioNumber or  setTransactionalFolioNumber.

severinbeauvais (approved):

src/mixins/filing-template-mixin.ts
severinbeauvais: Yes, I agree. You can change it if you want.

PR #444, opened 2023-02-15
15291 hide sole proprietor change button

jonathan-longe:
*Issue #:* /bcgov/entity#15291

The business has requested that non-staff users should not be able to modify the sole proprietor where the sole proprietor is an organization.  This PR hides the appropriate change button and updates the related help text.

### DIFF
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -283,7 +283,7 @@
 
               <!-- org-person we haven't touched: -->
               <div v-else class="actions mr-4">
-                <span class="edit-action" v-if="showSoleProprietorChangeButton">
+                <span class="edit-action" v-if="hideChangeButtonForSoleProps">
                   <v-btn
                     text color="primary"
                     :id="`officer-${index}-edit-btn`"
@@ -376,7 +376,7 @@ import OrgPerson from './OrgPerson.vue'
 import { CommonMixin, OrgPersonMixin } from '@/mixins/'
 import { IsSame } from '@/utils/'
 import { OrgPersonIF } from '@/interfaces/'
-import { showSoleProprietorChangeButton } from '@/store/getters'
+import { hideChangeButtonForSoleProps } from '@/store/getters'
 
 @Component({
   components: {
@@ -415,7 +415,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   @Getter isFirmConversionFiling!: boolean
   @Getter isFirmCorrectionFiling!: boolean
   @Getter isRoleStaff!: boolean
-  @Getter showSoleProprietorChangeButton!: boolean
+  @Getter hideChangeButtonForSoleProps!: boolean
 
   /** V-model for dropdown menus. */
   protected dropdown: Array<boolean> = []
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -283,7 +283,7 @@
 
               <!-- org-person we haven't touched: -->
               <div v-else class="actions mr-4">
-                <span class="edit-action">
+                <span class="edit-action" v-if="showSoleProprietorChangeButton">
                   <v-btn
                     text color="primary"
                     :id="`officer-${index}-edit-btn`"
@@ -376,6 +376,7 @@ import OrgPerson from './OrgPerson.vue'
 import { CommonMixin, OrgPersonMixin } from '@/mixins/'
 import { IsSame } from '@/utils/'
 import { OrgPersonIF } from '@/interfaces/'
+import { showSoleProprietorChangeButton } from '@/store/getters'
 
 @Component({
   components: {
@@ -414,6 +415,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   @Getter isFirmConversionFiling!: boolean
   @Getter isFirmCorrectionFiling!: boolean
   @Getter isRoleStaff!: boolean
+  @Getter showSoleProprietorChangeButton!: boolean
 
   /** V-model for dropdown menus. */
   protected dropdown: Array<boolean> = []
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -283,7 +283,7 @@
 
               <!-- org-person we haven't touched: -->
               <div v-else class="actions mr-4">
-                <span class="edit-action" v-if="hideChangeButtonForSoleProps">
+                <span class="edit-action" v-if="!hideChangeButtonForSoleProps">
                   <v-btn
                     text color="primary"
                     :id="`officer-${index}-edit-btn`"
@@ -376,7 +376,6 @@ import OrgPerson from './OrgPerson.vue'
 import { CommonMixin, OrgPersonMixin } from '@/mixins/'
 import { IsSame } from '@/utils/'
 import { OrgPersonIF } from '@/interfaces/'
-import { hideChangeButtonForSoleProps } from '@/store/getters'
 
 @Component({
   components: {
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -1060,13 +1060,15 @@ export const getSpecialResolutionConfirmValid = (state: StateIF): boolean => {
 export const getRestoration = (state: StateIF): RestorationStateIF => {
   return state.stateModel.restoration
 }
+
 /** Returns true when users can change the sole proprietor (SP).
  * Restricts the ability of non-staff users from changing the sole
- * proprietor when the SP is a non-individual.  This restriction has been
+ * proprietor when the SP is an organization.  This restriction has been
  * added to ensure that changes made in business-edit-ui are also updated by
  * staff in COLIN */
 export const showSoleProprietorChangeButton = (state: StateIF, getters): boolean => {
-  const isProprietor = getters.getOrgPeople?.[0].roles?.[0]?.roleType === RoleTypes.PROPRIETOR
-  const isOrganization = getters.getOrgPeople?.[0]?.officer?.partyType === PartyTypes.ORGANIZATION
-  return !(!getters.isRoleStaff && getters.isSoleProp && isOrganization && isProprietor)
+  const isProprietor = getters.getOrgPeople[0]?.roles[0]?.roleType === RoleTypes.PROPRIETOR
+  const isOrganization = getters.getOrgPeople[0]?.officer?.partyType === PartyTypes.ORGANIZATION
+  const isDba = isProprietor && isOrganization
+  return getters.isRoleStaff || !isDba
 }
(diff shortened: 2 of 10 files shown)


tests/unit/PeopleAndRoles.spec.ts
```diff
@@ -469,4 +469,49 @@ describe('People And Roles component for Change of Registration', () => {
     expect(wrapper.vm.$data.activeIndex).toBe(NaN)
     expect(wrapper.vm.$data.isAddingEditingOrgPerson).toBe(true)
   })
+
+  it('change button is not visible to users for SP where the sole proprietor is a non-individual', () => {
```
severinbeauvais, line 473: You can be specific here: a non-individual must be an organization.

src/store/getters/state-getters.ts
```diff
@@ -1029,3 +1060,13 @@ export const getSpecialResolutionConfirmValid = (state: StateIF): boolean => {
 export const getRestoration = (state: StateIF): RestorationStateIF => {
   return state.stateModel.restoration
 }
+/** Returns true when users can change the sole proprietor (SP).
+ * Restricts the ability of non-staff users from changing the sole
+ * proprietor when the SP is a non-individual.  This restriction has been
+ * added to ensure that changes made in business-edit-ui are also updated by
+ * staff in COLIN */
+export const showSoleProprietorChangeButton = (state: StateIF, getters): boolean => {
+  const isProprietor = getters.getOrgPeople?.[0].roles?.[0]?.roleType === RoleTypes.PROPRIETOR
```
severinbeauvais, line 1069: Is this a better syntax then `getOrgPeople(state)`?

I think you don't need the extra space between the conditional object and the array index -- I think this works too: `getOrgPeople[0]?.roles[0]?.roleType`.

src/store/getters/state-getters.ts
```diff
@@ -1029,3 +1060,13 @@ export const getSpecialResolutionConfirmValid = (state: StateIF): boolean => {
...
+  const isOrganization = getters.getOrgPeople?.[0]?.officer?.partyType === PartyTypes.ORGANIZATION
+  return !(!getters.isRoleStaff && getters.isSoleProp && isOrganization && isProprietor)
```
severinbeauvais, line 1071: Please simplify this negative-negative logic. It may be better to use a variable for the second part (isSoleProp && isOrganization && isProprietor) and then return "isRoleStaff && isNewVariable".

Another simplification: if this is a SP then the officer must necessarily be a proprietor. (That's the definition of Proprietor. Likewise, a Partner only exists for a GP.)

src/store/getters/state-getters.ts
```diff
@@ -1029,3 +1060,13 @@ export const getSpecialResolutionConfirmValid = (state: StateIF): boolean => {
 export const getRestoration = (state: StateIF): RestorationStateIF => {
   return state.stateModel.restoration
 }
+/** Returns true when users can change the sole proprietor (SP).
```
severinbeauvais, line 1063: Add blank line above.

src/store/getters/state-getters.ts
```diff
@@ -1029,3 +1060,13 @@ export const getSpecialResolutionConfirmValid = (state: StateIF): boolean => {
...
+ * Restricts the ability of non-staff users from changing the sole
+ * proprietor when the SP is a non-individual.  This restriction has been
+ * added to ensure that changes made in business-edit-ui are also updated by
+ * staff in COLIN */
+export const showSoleProprietorChangeButton = (state: StateIF, getters): boolean => {
```
severinbeauvais, line 1068: Since this getter is only used in 1 place, can it moved to that component (or its mixin)?

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -283,7 +283,7 @@
 
               <!-- org-person we haven't touched: -->
               <div v-else class="actions mr-4">
-                <span class="edit-action">
+                <span class="edit-action" v-if="showSoleProprietorChangeButton">
```
severinbeauvais, line 286: As on line 299 and 300, what do you think about labelling this "canEdit" ?

I'm thinking "separation of concerns" -- this template doesn't need to know about SP or Proprietor or that the edit label is "Change". KISS here, and in the script you can explain the reason for the logic.
This works well with moving the getter from the store to this component (or its mixin) -- the store doesn't need to ko

tests/unit/PeopleAndRoles.spec.ts
jonathan-longe: Good call, I'll change it.

src/store/getters/state-getters.ts
jonathan-longe: With regards to your first question, yes!  I think it's always better to write pure functions where all dependencies are injected.  Testing functions where the dependencies are not injected is more difficult and refactoring is more difficult.

With regards to your second point, good to know, I'll fix it.

src/store/getters/state-getters.ts
jonathan-longe: Good points, I'll simplify it as you propose.

src/store/getters/state-getters.ts
jonathan-longe: Will do

src/store/getters/state-getters.ts
jonathan-longe: My recommendation would be that all business logic belongs in the store.  Placing the logic in a component -- even if it's only used once -- means that we're more likely to create a duplicate when we need identical logic for another component.   I'm not crazy about the idea of extending mixins after you said that they weren't supported in Vue3.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
jonathan-longe: Let's hold off deciding on an appropriate name until we've decided on the appropriate location -- getter vs component.

src/store/getters/state-getters.ts
severinbeauvais: I might be wrong on the Vue3 mixin thing. Maybe they just need to be handled differently.

I generally agree with your point, but the "should I show this edit button" logic is very specific to the ListPeopleAndRoles component, so I still think it should be moved there instead of in the store. Not a deal breaker though.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Hehehe

src/store/getters/state-getters.ts
jonathan-longe: Let's discuss the business logic issue further.  It would be great to have a clear vision of the end goal.

src/store/getters/state-getters.ts
jonathan-longe: I really like the simplification you proposed!

src/store/getters/state-getters.ts
severinbeauvais: I'd like to see a code architecture where it's easy to create / drop in a new component without having to know much about the architecture, and it's easy to create a new thing (eg, filing) by simply putting existing components on a new page (view).

I'd like to see a relatively "dumb" UI, with as little business logic as possible.

And I like the idea of "independent" components... Maybe call them "smart" components (vs dumb), where most of what they do is localized and their interface to the rest of the system is clear / well defined. (I know this seems to go against the "store does everything" model. But maybe there's nice middle ground.)

I like the "tier" idea and separation of concerns, which to me keeps things independent and well defined. (It also makes unit testing easier.)

What do you think?

jonathan-longe: I think the architecture that I wrote for Roadsafety is what you're describing.  Form components (made up of one or more questions) could dropped into any new form and the only wiring required was the store location and the display options.  They had local smarts.  For example, they could display a spinner when busy but the component logic was encapsulated. The parent didn't need to know how the inner workings behaved and new fields could be added the shared component without changing the parent. I'll tee-up a meeting for us to discuss the implementation details.

severinbeauvais: Jonathan and I discussed this larger architectural question in various 1:1 meetings (esp in the context of the new Consent for Continuation Out filing, and re: app startup in App.vue, and mixins vs store modules).

Anyway, I'm OK with the implementation above since it's a temporary change, and long-term we'll figure out our ideal model so let's not get side tracked by that right now.

Resolving this conversation.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -283,7 +283,7 @@
 
               <!-- org-person we haven't touched: -->
               <div v-else class="actions mr-4">
-                <span class="edit-action">
+                <span class="edit-action" v-if="hideChangeButtonForSoleProps">
```
severinbeauvais, line 286: If NOT hide ?

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -376,6 +376,7 @@ import OrgPerson from './OrgPerson.vue'
 import { CommonMixin, OrgPersonMixin } from '@/mixins/'
 import { IsSame } from '@/utils/'
 import { OrgPersonIF } from '@/interfaces/'
+import { hideChangeButtonForSoleProps } from '@/store/getters'
```
severinbeauvais, line 379: Not needed.

src/store/getters/state-getters.ts
```diff
@@ -1029,3 +1060,15 @@ export const getSpecialResolutionConfirmValid = (state: StateIF): boolean => {
 export const getRestoration = (state: StateIF): RestorationStateIF => {
   return state.stateModel.restoration
 }
+
+/** Returns true when users can change the sole proprietor (SP).
```
severinbeauvais, line 1064: Switch the logic :)

tests/unit/PeopleAndRoles.spec.ts
```diff
@@ -469,4 +469,49 @@ describe('People And Roles component for Change of Registration', () => {
...
   })
+
+  it('change button is not visible to users for SP where the sole proprietor is an organization', () => {
+    store.state.stateModel.tombstone.keycloakRoles = ['user']
+    store.state.stateModel.tombstone.entityType = 'SP'
+    store.state.stateModel.peopleAndRoles.orgPeople = [
+      {
+        officer: {
+          partyType: 'organization',
+          id: '0'
```
severinbeauvais, line 480: You don't need the id for the getter, so prob better to leave it out. Also below.

severinbeauvais (approved):
GTG when you've sorted the logic.

Remember to update the app version in both PRs.

Would be nice to have other review approvals :)

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
jonathan-longe: Ah, good thinking!

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
jonathan-longe: Fixed

src/store/getters/state-getters.ts
jonathan-longe: Will do

tests/unit/PeopleAndRoles.spec.ts
jonathan-longe: Nice, good catch

severinbeauvais (approved):

severinbeauvais (commented):
Looks good but this PR is against main branch. If you intend to fix build 3.9.15 then you'll have to rebase.

jonathan-longe:
> Looks good but this PR is against main branch. If you intend to fix build 3.9.15 then you'll have to rebase.

Pull #448 was created to hotfix build 3.9.15; this PR was created to ensure the hotfix was also incorporated into the main branch.

PR #449, opened 2023-02-22
15081 add restoration extension and conversion

jonathan-longe:
*Issue #:* /bcgov/entity#15081

This PR makes some minor changes to the YourCompany.vue component to support restoration extension and conversion as shown in the screenshots below:

During a limited restoration extension, no changes to the name block are permitted.
image

During a limited restoration conversion to full, users can change the entity's name.
image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -160,7 +160,7 @@ import { ConfirmDialog as ConfirmDialogShared } from '@bcrs-shared-components/co
 import * as Views from '@/views/'
 import * as Dialogs from '@/dialogs/'
 import { AuthServices, LegalServices } from '@/services/'
-import { CommonMixin, DateMixin, FilingTemplateMixin } from '@/mixins/'
+import { CommonMixin, FilingTemplateMixin } from '@/mixins/'
 import { FilingDataIF, ActionBindingIF, ConfirmDialogType, FlagsReviewCertifyIF, FlagsCompanyInfoIF,
   AlterationFilingIF, ChgRegistrationFilingIF, ConversionFilingIF, SpecialResolutionFilingIF }
   from '@/interfaces/'
@@ -170,6 +170,7 @@ import { ComponentsCompanyInfo, ComponentsReviewCertify, RouteNames } from '@/en
 import { FeeSummaryActions } from '@bcrs-shared-components/enums/'
 import { getEntityDashboardBreadcrumb, getMyBusinessRegistryBreadcrumb, getRegistryDashboardBreadcrumb,
   getStaffDashboardBreadcrumb } from '@/resources/BreadCrumbResources'
+import DateUtilities from '@/services/date-utilities'
 
 @Component({
   components: {
@@ -184,9 +185,13 @@ import { getEntityDashboardBreadcrumb, getMyBusinessRegistryBreadcrumb, getRegis
     SbcFeeSummary,
     ...Dialogs,
     ...Views
-  }
+  },
+  mixins: [
+    CommonMixin,
+    FilingTemplateMixin
+  ]
 })
-export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMixin) {
+export default class App extends Vue {
   // Refs
   $refs!: {
     confirm: ConfirmDialogType
@@ -444,7 +449,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
   /** Fetches and stores the current JS date. */
   private async updateCurrentJsDate (): Promise<void> {
-    const jsDate = await this.getServerDate()
+    const jsDate = await DateUtilities.getServerDate()
     this.setCurrentJsDate(jsDate)
   }
 
@@ -494,7 +499,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
     this.resetFlags()
 
     // set current date from "real time" date from server
-    this.setCurrentDate(this.dateToYyyyMmDd(this.getCurrentJsDate))
+    this.setCurrentDate(DateUtilities.dateToYyyyMmDd(this.getCurrentJsDate))
 
     // get and store keycloak roles
     try {
--- a/src/components/Restoration/YourCompanySummary.vue
+++ b/src/components/Restoration/YourCompanySummary.vue
@@ -5,7 +5,7 @@
       <v-row no-gutters>
         <v-col cols="9">
           <img class="header-icon" src="@/assets/images/currency-usd-circle.svg">
-          <label class="summary-title">Your Company</label>
+          <label class="summary-title pl-2">Your Company</label>
         </v-col>
       </v-row>
     </div>
@@ -19,7 +19,7 @@
           </v-col>
           <v-col cols="8" class="mt-n1" v-if="isLimitedExtendRestorationFiling">
             <div class="font-weight-bold">Limited Restoration Extension</div>
-            <div>Extension Time: {{ getFormattedExpiryText }}</div>
+            <div>Extension Time: {{ getFormattedExpiryText() }}</div>
           </v-col>
           <v-col cols="8" class="mt-n1" v-if="isLimitedConversionRestorationFiling">
             <div class="font-weight-bold">Conversion to Full Restoration</div>
@@ -58,40 +58,6 @@ export default {
   display: flex;
   background-color: $BCgovBlue5O;
   padding: 1.25rem;
-}
-
-.summary-title {
-  padding-left: 0.5rem;
-}
-
-.company-name {
-  font-size: 1.5rem;
-}
-
-.actions {
-  position: absolute;
-  right: 0;
-
-  #btn-delete-restoration {
-    margin-top: -6px;
-  }
-}
-
-.inner-col-1 {
-  // adjustment to make this inner container column the same width as the outer columns
-  // ie, decrease width by 1/2 container margin + padding
-  flex: 0 0 calc(25% - 1.5rem);
-}
-
-.inner-col-2 {
-  // adjustment to make this inner container column the same width as the outer columns
-  // ie, increase width by 1/2 container margin + padding
-  flex: 0 0 calc(75% + 1.5rem);
-  max-width: calc(75% + 1.5rem);
-}
-
-#effective-date-text {
-  color: $gray7;
 }
 
 // hide first v-divider
--- a/src/components/Restoration/YourCompanySummary.vue
+++ b/src/components/Restoration/YourCompanySummary.vue
@@ -1,0 +1,101 @@
+<template>
+  <v-card flat id="restoration-summary">
+    <!-- Section Header -->
+    <div class="summary-header px-4 mb-2 rounded-t">
+      <v-row no-gutters>
+        <v-col cols="9">
+          <img class="header-icon" src="@/assets/images/currency-usd-circle.svg">
+          <label class="summary-title">Your Company</label>
+        </v-col>
+      </v-row>
+    </div>
+    <!-- Business Name -->
+    <template>
+      <v-divider class="mx-4" />
+      <div class="section-container business-name-summary">
+        <v-row no-gutters>
+          <v-col cols="3">
+            <label><strong>Restoration Type</strong></label>
+          </v-col>
+          <v-col cols="8" class="mt-n1" v-if="isLimitedExtendRestorationFiling">
+            <div class="font-weight-bold">Limited Restoration Extension</div>
+            <div>Extension Time: {{ getFormattedExpiryText }}</div>
+          </v-col>
+          <v-col cols="8" class="mt-n1" v-if="isLimitedConversionRestorationFiling">
+            <div class="font-weight-bold">Conversion to Full Restoration</div>
+            <div>[TODO - Applicant's relationship: Director, Shareholder]</div>
+          </v-col>
+        </v-row>
+        <v-row no-gutters class="mt-3" v-if="hasFileNumber">
+          <v-col cols="3">
+            <label><strong>Approval Type</strong></label>
+          </v-col>
+          <v-col cols="8" class="mt-n1">
+            <div class="font-weight-bold">Approved by Court Order</div>
+            <div>Court Order Number: {{ getFileNumber }}</div>
+          </v-col>
+        </v-row>
+      </div>
+    </template>
+  </v-card>
+</template>
+
+<script lang="ts">
+import { mapGetters } from 'vuex'
+
+export default {
+  computed: {
+    ...mapGetters(['isLimitedConversionRestorationFiling', 'isLimitedExtendRestorationFiling',
+      'hasFileNumber', 'getFileNumber', 'getFormattedExpiryText'])
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+.summary-header {
+  display: flex;
+  background-color: $BCgovBlue5O;
+  padding: 1.25rem;
+}
+
+.summary-title {
+  padding-left: 0.5rem;
+}
+
+.company-name {
+  font-size: 1.5rem;
+}
+
+.actions {
+  position: absolute;
+  right: 0;
+
+  #btn-delete-restoration {
+    margin-top: -6px;
+  }
+}
+
+.inner-col-1 {
+  // adjustment to make this inner container column the same width as the outer columns
+  // ie, decrease width by 1/2 container margin + padding
+  flex: 0 0 calc(25% - 1.5rem);
+}
+
+.inner-col-2 {
+  // adjustment to make this inner container column the same width as the outer columns
+  // ie, increase width by 1/2 container margin + padding
+  flex: 0 0 calc(75% + 1.5rem);
+  max-width: calc(75% + 1.5rem);
+}
+
+#effective-date-text {
+  color: $gray7;
+}
+
+// hide first v-divider
+.v-divider:first-of-type {
+  display: none;
+}
+</style>
--- a/src/components/common/YourCompany/NameTranslations/NameTranslation.vue
+++ b/src/components/common/YourCompany/NameTranslations/NameTranslation.vue
@@ -30,11 +30,7 @@
       </v-col>
 
       <!-- Actions -->
-      <v-col cols="2" class="mt-n2" v-if="
-      !hasNameTranslationChange &&
-      !isSummaryMode &&
-      !isLimitedConversionRestorationFiling &&
-      !isLimitedExtendRestorationFiling">
+      <v-col cols="2" class="mt-n2" v-if="isEditNameTranslationButtonVisible">
         <div class="actions mr-4">
           <v-btn
             class="correct-name-translation"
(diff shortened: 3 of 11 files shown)


jonathan-longe:
This PR is incomplete.  I'm not clear how to view the YourCompany summary as shown in the ticket.

src/store/getters/state-getters.ts
```diff
@@ -91,13 +91,13 @@ export const isRestorationFiling = (state: StateIF): boolean => {
 }
 
 /** Whether the current filing is a Limited Extension Restoration. */
-export const isLimitedExtendRestorationFiling = (state: StateIF): boolean => {
-  return (getRestoration(state).type === RestorationTypes.LTD_EXTEND)
+export const isLimitedExtendRestorationFiling = (state: StateIF, getters): boolean => {
+  return (getters.getRestoration.type === RestorationTypes.LTD_EXTEND)
```
jonathan-longe, line 95: Converting to pure functions as suggested by Vuex documentation.

src/components/common/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -30,7 +30,11 @@
       </v-col>
 
       <!-- Actions -->
-      <v-col cols="2" class="mt-n2" v-if="!hasNameTranslationChange && !isSummaryMode">
+      <v-col cols="2" class="mt-n2" v-if="
+      !hasNameTranslationChange &&
+      !isSummaryMode &&
+      !isLimitedConversionRestorationFiling &&
+      !isLimitedExtendRestorationFiling">
```
jonathan-longe, line 37: I'm following an existing pattern, but this needs refactoring as it will become increasingly messy and difficult to read as the number of filings increase.  Also, it's impossible to add a new filing type without risking making inadvertent changes to functionality that's already working.

jonathan-longe:
Adding a summary component for the restoration filing

image

image

src/App.vue
```diff
@@ -444,7 +449,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
   /** Fetches and stores the current JS date. */
   private async updateCurrentJsDate (): Promise<void> {
-    const jsDate = await this.getServerDate()
+    const jsDate = await DateUtilities.getServerDate()
```
jonathan-longe, line 452: Moving date utilities from a Mixin to a Service.  Since these were written as pure functions they're easy to move around.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -0,0 +1,101 @@
...
+          <v-col cols="3">
+            <label><strong>Restoration Type</strong></label>
+          </v-col>
+          <v-col cols="8" class="mt-n1" v-if="isLimitedExtendRestorationFiling">
+            <div class="font-weight-bold">Limited Restoration Extension</div>
+            <div>Extension Time: {{ getFormattedExpiryText }}</div>
+          </v-col>
+          <v-col cols="8" class="mt-n1" v-if="isLimitedConversionRestorationFiling">
+            <div class="font-weight-bold">Conversion to Full Restoration</div>
+            <div>[TODO - Applicant's relationship: Director, Shareholder]</div>
```
jonathan-longe, line 26: This block isn't part of the current ticket, but I'm adding a placeholder to be replaced at a later date.

src/mixins/date-mixin.ts
```diff
@@ -1,252 +1,145 @@
...
-export default class DateMixin extends CommonMixin {
+@Component({
+  mixins: [
+    CommonMixin
+  ]
+})
+export default class DateMixin extends Vue {
   /**
-   * Fetches and returns the web server's current date (in UTC).
+   * DEPRECATED Fetches and returns the web server's current date (in UTC).
```
jonathan-longe, line 16: The date-mixin now becomes a wrapper for the date-utilities service.  As a service, I can use these functions in the store getters whereas as a mixin, I cannot.

src/services/date-utilities.ts
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * Returns the number of months remaining when given an expiry date string
+   * like this: monthDiffToToday('2023-12-31')
+   * if needed, today's date can be passed as the second parameter
+   */
+  static monthDiffToToday (dateToStr: string, dateFrom = new Date()): number {
+    const dateTo = this.mmmDdYyyyToDate(dateToStr)
+    return dateTo.getMonth() - dateFrom.getMonth() +
+      (12 * (dateTo.getFullYear() - dateFrom.getFullYear()))
+  }
```
jonathan-longe, line 256: This is the only new function that's been added to date-utilities.ts

src/store/getters/state-getters.ts
```diff
@@ -1072,3 +1078,15 @@ export const hideChangeButtonForSoleProps = (state: StateIF, getters): boolean =
...
+  return state.stateModel.restoration?.expiry
+}
+
+export const getFormattedExpiryText = (state, getters): string => {
+  if (getters.getExpiryDateString) {
+    return DateUtilities.monthDiffToToday(getters.getExpiryDateString) + ' months, expires on ' +
+      DateUtilities.yyyyMmDdToPacificDate(getters.getExpiryDateString)
+  }
+  return '[no expiry date]'
+}
```
jonathan-longe, line 1092: I'd like an opportunity to refactor this and create a module for Restorations.  These getters, and others, would then be moved to the restoration module's getters.

tests/unit/state-getters.spec.ts
```diff
@@ -406,8 +408,8 @@ describe('SP/GP correction getters', () => {
     naicsDescription: 'NAICS description'
   }
   const newAddress = {
-    deliveryAddress: { postalCode: 'V8V 8V8' },
-    mailingAddress: { postalCode: 'V8V 8V8' }
+    deliveryAddress: {postalCode: 'V8V 8V8'},
+    mailingAddress: {postalCode: 'V8V 8V8'}
```
jonathan-longe, line 412: Oh, looks like my editor made these changes without me noticing.  I'll fix them.

src/store/getters/state-getters.ts
severinbeauvais: I'm pleased to have a suggestion on this, but what is the type of `getters.getRestoration`?

src/components/common/YourCompany/NameTranslations/NameTranslation.vue
severinbeauvais: I'd prefer to see a computed property for anything more than trivial logic -- such as this. As a bonus, you can document the getter to clarify what it's doing.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -360,6 +360,8 @@ export default class YourCompany extends Mixins(
   @Getter isRestorationFiling!: boolean
   @Getter getEntityType!: CorpTypeCd
   @Getter getAssociationType!: CoopTypes
+  @Getter isLimitedExtendRestorationFiling!: boolean
+  @Getter isLimitedConversionRestorationFiling!: boolean
```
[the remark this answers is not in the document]
jonathan-longe: Good catch, I'll remove it.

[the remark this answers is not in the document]
severinbeauvais: I think you still need `isLimitedExtendRestorationFiling`, just not the other one.

src/store/getters/state-getters.ts
jonathan-longe: That's a good question!  If it's okay with you, I'd like to change the way we populate the store with the restoration filing.  At present, the `parseRestorationFiling()` method, set individual values with methods like: `setRestorationType()` and `setRestorationExpiry()`.  What I'd prefer to do, is load the following three resources into the store and use getters to extract the data we need.  

- restoration filing
- entity snapshot
- state filing

This approach greatly reduces the complexity of populating the store and it gives us more flexibility to get whatever data we need -- both now and in the future.

src/components/common/YourCompany/YourCompany.vue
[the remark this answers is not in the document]
jonathan-longe: Haha, you're right.  I should have checked what was being used.  My bad!

src/components/common/YourCompany/NameTranslations/NameTranslation.vue
jonathan-longe: That's a good idea.  I'll create a getter to replace that logic.

src/store/getters/state-getters.ts
severinbeauvais: Sure, go for it. (Don't get bogged down.)

src/components/common/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -224,6 +220,14 @@ export default class NameTranslation extends Mixins(CommonMixin) {
     return this.draftTranslations.filter(x => x.action !== ActionTypes.REMOVED)
   }
 
+  /** Returns true when user can edit the name translations. */
+  get isEditNameTranslationButtonVisible (): boolean {
+    return (!this.hasNameTranslationChange &&
+    !this.isSummaryMode &&
+    !this.isLimitedConversionRestorationFiling &&
+    !this.isLimitedExtendRestorationFiling)
+  }
```
severinbeauvais, line 229: What do you think about this style instead?
```
return (
 condition1 &&
 condition2 ...
)
```

src/store/getters/state-getters.ts
jonathan-longe: What do you think about creating a separate ticket for the work so it's visible and gets approval before proceeding?   Is that worthwhile? There may be other refactoring tasks that are higher priority.

severinbeauvais: I think we wouldn't prioritize a refactoring ticket right now, but if I found some easy changes that moved towards our vision, I'd probably do them as part of existing work.

src/components/common/YourCompany/NameTranslations/NameTranslation.vue
jonathan-longe: I was reluctant to mess with the logic.  Are you suggesting removing the `!` in front of each condition?  I just noticed that I should change the method's name and comment to make clear that the get method returns true when the button should be hidden -- the logic is correct but the wording is incorrect.

severinbeauvais: The style, not the logic.

severinbeauvais: If you hate the logic, you could do this:
```
if (a || b || c) return true
return false
```

jonathan-longe: Ah, got it, thanks for the clarification.  I've changed both the format and the cleaned up the logic so it's not this and not that etc.  Let me know if that works for you.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -0,0 +1,101 @@
...
+  </v-card>
+</template>
+
+<script lang="ts">
+import { mapGetters } from 'vuex'
+
+export default {
+  computed: {
+    ...mapGetters(['isLimitedConversionRestorationFiling', 'isLimitedExtendRestorationFiling',
+      'hasFileNumber', 'getFileNumber', 'getFormattedExpiryText'])
```
severinbeauvais, line 49: I know you like these, but we lose typing this way. Thoughts?

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -0,0 +1,101 @@
...
+@import '@/assets/styles/theme.scss';
+
+.summary-header {
+  display: flex;
+  background-color: $BCgovBlue5O;
+  padding: 1.25rem;
+}
+
+.summary-title {
+  padding-left: 0.5rem;
```
severinbeauvais, line 64: After much UI cleanup in the past, I favour using the [Vuetify spacing classes](https://vuetifyjs.com/en/styles/spacing/) right in the template for whitespace and general layout. This is way easier to understand and tweak than needing to look up the CSS for the `summary-title` and all other classes. Thoughts?

src/components/common/YourCompany/NameTranslations/NameTranslation.vue
```diff
@@ -30,7 +30,7 @@
       </v-col>
 
       <!-- Actions -->
-      <v-col cols="2" class="mt-n2" v-if="!hasNameTranslationChange && !isSummaryMode">
+      <v-col cols="2" class="mt-n2" v-if="isHideEditNameTranslationButton">
```
severinbeauvais, line 33: Maybe this property should be renamed :)

src/services/date-utilities.ts
severinbeauvais: Using `new Date()` is dangerous because that returns the browser's date, which might be in a different timezone or wrong/spoofed. Could you make that second parameter required instead?

src/store/getters/state-getters.ts
severinbeauvais: If it fits within the estimate for this ticket, go for it.

(An estimate of 3 should equal 1-3 days of work.)

src/components/Restoration/YourCompanySummary.vue
jonathan-longe: Oh, right, I meant to call that out. The `@Getter` syntax is only available if we're using the Class API (thanks for providing a name to that pattern).  Since the Class API isn't available in Vue3 and we have upgrade prior to December 2023, I thought I'd use the Options API for any new components.  What do you think?

src/components/common/YourCompany/NameTranslations/NameTranslation.vue
jonathan-longe: Shoot, you're right, I reversed it when it was correct before

src/components/Restoration/YourCompanySummary.vue
jonathan-longe: Yes, I agree, I prefer that too.  I'll change it.

src/services/date-utilities.ts
jonathan-longe: Dangerous? For critical stuff, I would agree.  In this case, the downside risk is the number of months remaining is incorrect which is, I would suggest, of no consequence.  The way it's written a developer can override it if its critical.  I have added a second parameter to `getFormattedExpiryText()` so we can force today's date to a static value for testing etc. I'll push that change shortly.  We should chat about how we handle today's date application wide.  I'm not sure I understand the benefit of pulling today's date from the server.

src/components/Restoration/YourCompanySummary.vue
severinbeauvais: I don't know, so let's just leave this.

src/services/date-utilities.ts
severinbeauvais: Your current usage of this function may be low risk, but a future developer may reuse it without my caveat above.

Could you default the second parameter to `getCurrentDate` or `getCurrentJsDate` (store getters) instead?

jonathan-longe: The date-utilities service can't reference the store.  This is by design.

I can't see how we can trust the date on the front-end.  Even if we make a call from our server with the current date, the user could change it.   I'm not sure what we're trying to protect by getting the current date and time from the server.  If a user must file a document before midnight, it's the server's date and time that counts.  If the user is late filing, the server returns a business rule error and we gracefully handle that. What am I missing?

severinbeauvais: We have some "future effective" filings, where the user (usually a lawyer/paralegal) specifies an exact date and time to file an IA, for example. We must ensure that when the user says "noon Pacific" that it isn't skewed by a timezone difference or whatever.

I'm thinking that, if you make that second parameter required, we pass the buck to the caller to ensure a valid value, instead of tucking a "dangerous" `new Date()` in this class. Thoughts?

jonathan-longe: In our call yesterday, I think we agreed that there's nothing dangerous about `new Date()` or relying on the browser's date and time.  The [JavaScript documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) includes no warnings on its use.  We must rely on the API for any current date and time validations.

jonathan-longe:
@severinbeauvais let me know if you're okay approving this PR.  I'm happy to make further revisions if you wish.

src/services/date-utilities.ts
severinbeauvais: Yes, you're right: using `new Date()` isn't dangerous, but assuming it returns a valid current date/time could cause issues (even though I have yet to come up with a solid example for you).

How about this -- require `dateFrom` above, and in the caller go ahead and use `new Date()` for its values. That satisfies me that the main logic (in the parent) pays attention to what it's doing, and THIS function doesn't make any assumptions.

src/store/getters/state-getters.ts
severinbeauvais: Was a decision made re: this conversation?

jonathan-longe: With the time remaining in the sprint and the other tickets assigned to me, I don't think I have time to dig into this at present.

src/mixins/date-mixin.ts
JazzarKarim: This is cool. I like it!

src/services/date-utilities.ts
JazzarKarim: I've actually implemented a method that does the exact same thing in Business Create UI. That's the downside of not having Mixins shared 😟
https://github.com/bcgov/business-create-ui/blob/5e7967885de598886fa6eb3782054e73243d45ac/src/mixins/date-mixin.ts#L260

JazzarKarim: Now, there's going to be inconsistency between Edit UI and Create UI. How should we proceed?

jonathan-longe: > Yes, you're right: using `new Date()` isn't dangerous, but assuming it returns a valid current date/time could cause issues (even though I have yet to come up with a solid example for you).

I think we need to come up with those examples of where `new Date()` could cause problems for our application.  Once we know the problem we're trying to solve, the rest becomes so much easier.  I've designed `monthDiffToToday` so it's nice and clean and the parent doesn't have to provide the current date unless it's a test that requries a static date.  If the browser's current time is using the wrong time zone, the function will off by plus or minus 12 hours when calculating the number of months remaining.  I think we have bigger issues to solve, but once you can cite those examples we'll know for sure.

jonathan-longe: > Now, there's going to be inconsistency between Edit UI and Create UI. How should we proceed?

That's a very good point Karim.  The best way to solve the duplication is to merge the repos.  I can understand that was a good reason to keep the repos separate, now I think there are many more benefits to being able to share each other's good work.

severinbeauvais: I still don't like `new Date()` in a utility function (would prefer higher-level code to make that decision), but you're in this with me so let's merge this and evaluate as we go.

severinbeauvais: > ... The best way to solve the duplication is to merge the repos. I can understand that was a good reason to keep the repos separate, now I think there are many more benefits to being able to share each other's good work.

... Which isn't something we're going to solve in this PR. If the 2 functions are relatively close then let's not worry about them right now. (I think there are other Date Mixin differences anyway.)

JazzarKarim (approved):
Looks good to me 😄

ketaki-deodhar (approved):

severinbeauvais (approved):
👍

PR #450, opened 2023-03-07
14975 - Add Applicant Information (Extension and Conversion)

ketaki-deodhar:
*Issue #:* /bcgov/entity#14975

*Description of changes:*
Add applicant Information component for extension and Conversion

- new role type (applicant)
- allow add (same rules as full/limited resto)
- allow edit or delete
- validation

### DIFF
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -25,8 +25,8 @@
     <div
       v-if="currentPeopleAndRoles.length > 0"
       id="people-roles-list"
+      class="section-container"
       :class="{
-        'section-container': !isSummaryView,
         'invalid-section': isComponentInvalid
       }"
     >
@@ -79,7 +79,9 @@
                   <p class="people-roles-title mb-1" :class="{ 'removed': wasRemoved(orgPerson)}">
                     {{ formatName(orgPerson) }}
                   </p>
-                  <p class="info-text mb-1 people-roles-email" :class="{ 'removed': wasRemoved(orgPerson)}">
+                  <p v-if="!(isLimitedConversionRestorationFiling || isLimitedExtendRestorationFiling)"
+                    class="info-text mb-1 people-roles-email"
+                    :class="{ 'removed': wasRemoved(orgPerson)}">
                     {{ orgPerson.officer.email }}
                   </p>
                   <p
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -413,6 +413,8 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   @Getter isFirmChangeFiling!: boolean
   @Getter isFirmConversionFiling!: boolean
   @Getter isFirmCorrectionFiling!: boolean
+  @Getter isLimitedExtendRestorationFiling!: boolean
+  @Getter isLimitedConversionRestorationFiling!: boolean
   @Getter isRoleStaff!: boolean
 
   /** V-model for dropdown menus. */
@@ -472,6 +474,9 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
     if (this.isFirmCorrectionFiling) {
       // cannot remove proprietor/partner
       return false
+    }
+    if (this.isLimitedConversionRestorationFiling || this.isLimitedExtendRestorationFiling) {
+      return true
     }
     return false // should never happen
   }
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -128,24 +128,22 @@
               </template>
             </v-col>
 
-            <v-col cols="12" sm="3" :class="{ 'removed': wasRemoved(orgPerson)}">
+            <v-col cols="12" sm="3" :class="{ 'removed': wasRemoved(orgPerson)}" v-if="showEmailColumn">
               <p>{{ orgPerson.officer.email }}</p>
             </v-col>
 
             <!-- Roles -->
-            <v-col cols="12" sm="2" :class="{ 'removed': wasRemoved(orgPerson)}">
-              <template v-if="isBenBcCccUlcCorrectionFiling">
-                <!-- Warning if orgPerson has no roles -->
-                <div v-if="orgPerson.roles.length > 0">
-                  <v-col v-for="(role, index) in orgPerson.roles" :key="index" class="col-roles">
-                    <span class="info-text small-text break-spaces pr-2">{{ role.roleType }}</span>
-                  </v-col>
-                </div>
-                <div v-else>
-                  <v-icon color="red darken-3">mdi-alert</v-icon>
-                  <span class="warning-text small-text">Missing Role</span>
-                </div>
-              </template>
+            <v-col cols="12" sm="2" :class="{ 'removed': wasRemoved(orgPerson)}" v-if="showRolesColumn">
+              <!-- Warning if orgPerson has no roles -->
+              <div v-if="orgPerson.roles.length > 0">
+                <v-col v-for="(role, index) in orgPerson.roles" :key="index" class="col-roles">
+                  <span class="info-text small-text break-spaces pr-2">{{ role.roleType }}</span>
+                </v-col>
+              </div>
+              <div v-else>
+                <v-icon color="red darken-3">mdi-alert</v-icon>
+                <span class="warning-text small-text">Missing Role</span>
+              </div>
             </v-col>
 
             <!-- Actions Buttons -->
@@ -409,6 +407,10 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   /** Whether OrgPersons list is valid. */
   @Prop({ default: true }) readonly validOrgPersons!: boolean
 
+  @Prop({ default: true }) readonly showDeliveryAddressColumn!: boolean
+  @Prop({ default: true }) readonly showRolesColumn!: boolean
+  @Prop({ default: false }) readonly showEmailColumn!: boolean
+
   // Store getter
   @Getter getOrgPeople!: OrgPersonIF[]
   @Getter isAlterationFiling!: boolean
@@ -425,14 +427,12 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   protected dropdown: Array<boolean> = []
 
   /** Headers for the person table. */
-  get tableHeaders () {
-    return [
-      'Name',
-      'Mailing Address',
-      !(this.isLimitedExtendRestorationFiling || this.isLimitedConversionRestorationFiling) ? 'Delivery Address' : '',
-      (this.isLimitedExtendRestorationFiling || this.isLimitedConversionRestorationFiling) ? 'Email' : '',
-      this.isBenBcCccUlcCorrectionFiling ? 'Roles' : ''
-    ]
+  get tableHeaders (): Array<string> {
+    const headers = ['Name', 'Mailing Address']
+    if (this.showDeliveryAddressColumn) headers.push('Delivery Address')
+    if (this.showRolesColumn) headers.push('Roles')
+    if (this.showEmailColumn) headers.push('Email Address')
+    return headers
   }
 
   /** This component's validity state (for error styling). */
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -115,7 +115,7 @@
             </v-col>
 
             <!-- Delivery Address -->
-            <v-col cols="12" sm="3" :class="{ 'removed': wasRemoved(orgPerson)}">
+            <v-col cols="12" sm="3" :class="{ 'removed': wasRemoved(orgPerson)}" v-if="showDeliveryAddressColumn">
               <template
                 v-if="hasRoleDirector(orgPerson) || hasRoleProprietor(orgPerson) || hasRolePartner(orgPerson)"
               >
@@ -126,6 +126,10 @@
                   <DeliveryAddress class="peoples-roles-delivery-address" :address="orgPerson.deliveryAddress" />
                 </template>
               </template>
+            </v-col>
+
+            <v-col cols="12" sm="3" :class="{ 'removed': wasRemoved(orgPerson)}">
+              <p>{{ orgPerson.officer.email }}</p>
             </v-col>
 
             <!-- Roles -->
@@ -425,7 +429,8 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
     return [
       'Name',
       'Mailing Address',
-      'Delivery Address',
+      !(this.isLimitedExtendRestorationFiling || this.isLimitedConversionRestorationFiling) ? 'Delivery Address' : '',
+      (this.isLimitedExtendRestorationFiling || this.isLimitedConversionRestorationFiling) ? 'Email' : '',
       this.isBenBcCccUlcCorrectionFiling ? 'Roles' : ''
     ]
   }
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -287,7 +287,7 @@
 
               <!-- org-person we haven't touched: -->
               <div v-else class="actions mr-4">
-                <span class="edit-action">
+                <span class="edit-action" v-if="!hideChangeButtonForSoleProps">
                   <v-btn
                     text color="primary"
                     :id="`officer-${index}-edit-btn`"
@@ -424,6 +424,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   @Getter isLimitedExtendRestorationFiling!: boolean
   @Getter isLimitedConversionRestorationFiling!: boolean
   @Getter isRoleStaff!: boolean
+  @Getter hideChangeButtonForSoleProps!: boolean
 
   /** V-model for dropdown menus. */
   protected dropdown: Array<boolean> = []
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -41,7 +41,7 @@
           <span>{{ title }}</span>
         </v-col>
         <!-- Spacer Column for Actions -->
-        <v-col lg="1"></v-col>
+        <v-col class="actions-width"></v-col>
       </v-row>
 
       <!-- List Content -->
@@ -149,7 +149,7 @@
             </v-col>
 
             <!-- Actions Buttons -->
-            <v-col v-if="!isSummaryView" class="pr-0">
+            <v-col v-if="!isSummaryView" class="actions">
               <!-- org-person that was replaced: -->
               <div v-if="wasReplaced(orgPerson)" class="actions pr-5">
                 <span class="undo-action">
@@ -624,18 +624,27 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
     right: 0;
     margin-top: -0.5rem;
 
-    .v-btn {
-      min-width: 0.5rem;
-    }
+    @extend .actions-width;
+    margin-top: -8px;
 
     .v-btn + .v-btn {
       margin-left: 0.5rem;
+    }
+
+    .more-actions-btn {
+      padding: 0;
+      min-width: 28px;
     }
   }
 
   .dropdown-action {
     border-left: 1px solid $gray3;
   }
+}
+
+.actions-width {
+  min-width: 150px;
+  max-width: 150px;
 }
 
 .summary-view {
@@ -683,10 +692,5 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   .badges .v-icon {
     margin-top: 4px;
   }
-}
-
-.actions-width {
-  min-width: 110px;
-  max-width: 110px;
 }
 </style>
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -25,8 +25,8 @@
     <div
       v-if="currentPeopleAndRoles.length > 0"
       id="people-roles-list"
-      class="section-container"
       :class="{
+        'section-container': !isSummaryView,
         'invalid-section': isComponentInvalid
       }"
     >
@@ -79,7 +79,7 @@
                   <p class="people-roles-title mb-1" :class="{ 'removed': wasRemoved(orgPerson)}">
                     {{ formatName(orgPerson) }}
                   </p>
-                  <p v-if="!(isLimitedConversionRestorationFiling || isLimitedExtendRestorationFiling)"
+                  <p v-if="showEmail"
                     class="info-text mb-1 people-roles-email"
                     :class="{ 'removed': wasRemoved(orgPerson)}">
                     {{ orgPerson.officer.email }}
@@ -415,6 +415,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
 
   // Store getter
   @Getter getOrgPeople!: OrgPersonIF[]
+  @Getter getResource!: ResourceIF
   @Getter isAlterationFiling!: boolean
   @Getter isBenBcCccUlcCorrectionFiling!: boolean
   @Getter isCorrectionFiling!: boolean
@@ -521,6 +522,11 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
       return this.formatFullName(orgPerson.officer)
     }
     return ''
+  }
+
+  /** flag to show email under name. */
+  get showEmail (): string {
+    return this.getResource.showEmail
   }
 
   /**
@@ -677,5 +683,10 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   .badges .v-icon {
     margin-top: 4px;
   }
+}
+
+.actions-width {
+  min-width: 110px;
+  max-width: 110px;
 }
 </style>
--- a/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
@@ -79,7 +79,7 @@
                   <p class="people-roles-title mb-1" :class="{ 'removed': wasRemoved(orgPerson)}">
                     {{ formatName(orgPerson) }}
                   </p>
-                  <p v-if="showEmail"
+                  <p v-if="showEmailUnderName"
                     class="info-text mb-1 people-roles-email"
                     :class="{ 'removed': wasRemoved(orgPerson)}">
                     {{ orgPerson.officer.email }}
@@ -412,10 +412,10 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   @Prop({ default: true }) readonly showDeliveryAddressColumn!: boolean
   @Prop({ default: true }) readonly showRolesColumn!: boolean
   @Prop({ default: false }) readonly showEmailColumn!: boolean
+  @Prop({ default: true }) readonly showEmailUnderName!: string
 
   // Store getter
   @Getter getOrgPeople!: OrgPersonIF[]
-  @Getter getResource!: ResourceIF
   @Getter isAlterationFiling!: boolean
   @Getter isBenBcCccUlcCorrectionFiling!: boolean
   @Getter isCorrectionFiling!: boolean
@@ -522,11 +522,6 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
       return this.formatFullName(orgPerson.officer)
     }
     return ''
-  }
-
-  /** flag to show email under name. */
-  get showEmail (): string {
-    return this.getResource.showEmail
   }
 
   /**
--- a/src/components/common/PeopleAndRoles/OrgPerson.vue
+++ b/src/components/common/PeopleAndRoles/OrgPerson.vue
@@ -393,6 +393,8 @@ export default class OrgPerson extends Mixins(CommonMixin, OrgPersonMixin) {
   @Getter isBenBcCccUlcCorrectionFiling!: boolean
   @Getter isFirmCorrectionFiling!: boolean
   @Getter isFirm!: boolean
+  @Getter isLimitedExtendRestorationFiling!: boolean
+  @Getter isLimitedConversionRestorationFiling!: boolean
   @Getter isRoleStaff!: boolean
 
   // Local variables
@@ -537,6 +539,9 @@ export default class OrgPerson extends Mixins(CommonMixin, OrgPersonMixin) {
     if (this.isFirmCorrectionFiling) {
       // cannot remove proprietor/partner
       return false
+    }
+    if (this.isLimitedConversionRestorationFiling || this.isLimitedExtendRestorationFiling) {
+      return true
     }
     return false // should never happen
   }
--- a/src/components/common/PeopleAndRoles/OrgPerson.vue
+++ b/src/components/common/PeopleAndRoles/OrgPerson.vue
@@ -99,32 +99,56 @@
                       Business or Corporation is Unregistered in B.C.
                     </a>
 
-                    <v-card v-if="wasBusinessSelectedFromLookup" outlined class="message-box rounded-0 mt-6">
-                      <p>
-                        <strong>Important:</strong> If the addresses shown below are out of date, you may
-                        update them here. The updates are applicable only to this registration.
-                      </p>
-                    </v-card>
+                    <div v-if="wasBusinessSelectedFromLookup">
+                      <v-card outlined class="message-box rounded-0 mt-6">
+                        <p>
+                          <strong>Important:</strong> If the addresses shown below are out of date, you may
+                          update them here. The updates are applicable only to this registration.
+                        </p>
+                      </v-card>
+                    </div>
+                    <div v-else>
+                      <div v-if="isProprietor">
+                        <p class="info-text mt-6 pt-0 mb-0">
+                          To add a registered B.C. business or corporation as the Proprietor, enter the name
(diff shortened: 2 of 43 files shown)


src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -413,20 +421,21 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
...
-      'Delivery Address',
-      this.isBenBcCccUlcCorrectionFiling ? 'Roles' : ''
-    ]
+  get tableHeaders (): Array<string> {
+    const headers = ['Name', 'Mailing Address']
+    if (this.showDeliveryAddressColumn) headers.push('Delivery Address')
+    if (this.showRolesColumn) headers.push('Roles')
+    if (this.showEmailColumn) headers.push('Email Address')
+    return headers
   }
```
ketaki-deodhar, line 440: Show only applicable columns

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -474,6 +483,9 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
       // cannot remove proprietor/partner
       return false
     }
+    if (this.isLimitedConversionRestorationFiling || this.isLimitedExtendRestorationFiling) {
+      return true
+    }
```
ketaki-deodhar, line 488: add extension and conversion to can remove org-person function

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -99,32 +99,56 @@
...
-                        v-if="!isRoleStaff"
-                        class="mt-6"
-                        :helpSection="getResource.changeData.orgPersonInfo.helpSection"
-                      />
-                    </template>
+                          v-if="!isRoleStaff"
+                          class="mt-6"
+                          :helpSection="getResource.changeData.orgPersonInfo.helpSection"
+                        />
+                    </div>
```
ketaki-deodhar, line 152: refactor to show correct text based on if Proprietor, Partner or else

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -124,6 +124,62 @@
...
+        </v-btn>
+        <p v-if="!hasApplicant" class="error-text small-text mt-5 mb-0">
+          You must have one applicant
+        </p>
+        <p v-if="!haveRequiredAddresses" class="error-text small-text mt-5 mb-0">
+          A applicant address is missing or incorrect
+        </p>
+        </div>
+      </article>
+
```
ketaki-deodhar, line 182: Add text and buttons

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -456,9 +527,15 @@ export default class PeopleAndRoles extends Mixins(CommonMixin, DateMixin, OrgPe
...
-      const thisPerson = (id !== undefined) && cloneDeep(this.originalParties.find(x => +x.officer.id === +id))
+      const regexExp = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/gi
+
+      let thisPerson
+      if (id !== undefined && regexExp.test(id)) {
+        thisPerson = cloneDeep(this.originalParties.find(x => x.officer.id === id))
+      } else {
+        // get a copy of original person from original IA
+        thisPerson = cloneDeep(this.originalParties.find(x => +x.officer.id === +id))
+      }
```
ketaki-deodhar, line 539: We can't reliably convert a UUID to a JavaScript/TypeScript number. The number type is a floating-point type and becomes potentially imprecise. Hence, compared as above (String)

src/views/Restoration.vue
```diff
@@ -208,6 +222,14 @@ export default class Restoration extends Vue {
...
       const entitySnapshot = await this.fetchEntitySnapshot()
 
+      // set Applicant info in entitySnapshot
+      const stateFiling = entitySnapshot.businessInfo.stateFiling
+      const filing = stateFiling && await LegalServices.fetchFiling(stateFiling)
+      entitySnapshot.orgPersons = filing.restoration.parties as OrgPersonIF
+      entitySnapshot.orgPersons.forEach(o => {
+        o.officer.id = uuidv4()
+      })
+
```
ketaki-deodhar, line 232: **fetchDirectors** or **fetchParties** Legal API endpoint return **business/entity** parties. 
For extension/conversion, correct parties need to be fetched from 'stateFiling' (applicant role) and set in entitySnapshot. 
Also, need to assign a randomly created 'id' using UUIDV4 as 'stateFiling' data does not return id for the orgPerson.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
jonathan-longe: Good thinking!

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -124,6 +124,62 @@
...
+        <ul>
+          <li>
+            <v-icon>mdi-circle-small</v-icon>
+            <span>An individual</span>
+          </li>
+          <li>
+            <v-icon>mdi-circle-small</v-icon>
+            <span>A business or a corporation</span>
+          </li>
+        </ul>
```
jonathan-longe, line 145: Is it possible that adding this code to `PeopleAndRoles` might be mixing concerns?  If this code is relevant to Restoration filings only, can we place it in separate component and call it from the parent?

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -79,7 +79,9 @@
                   <p class="people-roles-title mb-1" :class="{ 'removed': wasRemoved(orgPerson)}">
                     {{ formatName(orgPerson) }}
                   </p>
-                  <p class="info-text mb-1 people-roles-email" :class="{ 'removed': wasRemoved(orgPerson)}">
+                  <p v-if="!(isLimitedConversionRestorationFiling || isLimitedExtendRestorationFiling)"
+                    class="info-text mb-1 people-roles-email"
+                    :class="{ 'removed': wasRemoved(orgPerson)}">
                     {{ orgPerson.officer.email }}
```
jonathan-longe, line 85: Is there a way to decouple this?  If `ListPeopleAndRoles` is a common component, it would be better not to add logic that's specific to a restoration filing.  It's not ideal, but perhaps we could add a getter called something like, `showOfficerEmail` and have the business logic in the getter instead?

jonathan-longe (commented):
Looks good @ketaki-deodhar Let me chat with @severinbeauvais to see if there's an easy way to decouple the restoration specific business logic from the common components.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
ketaki-deodhar: I agree. I was on the fence while adding this as well.

severinbeauvais: If possible, this logic should be abstracted (ie, not so much whether this is not a resto filing, but whether this is displayed or not), and then controlled by a resource file property.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -25,8 +25,8 @@
     <div
       v-if="currentPeopleAndRoles.length > 0"
       id="people-roles-list"
+      class="section-container"
```
severinbeauvais, line 28: Can you provide before and after screenshots in edit view and summary view, please?

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: It looks like the 4 columns above are equal widths -- but the actions column should be narrower (fixed width).

src/components/common/PeopleAndRoles/OrgPerson.vue
```diff
@@ -99,32 +99,56 @@
...
-                        enter the name or incorporation number.
-                      </p>
-
-                      <p class="info-text mt-6 pt-0 mb-0">
-                        If you are adding a company that is not legally required to register in B.C.
-                        such as a bank or a railway, use the manual entry form. All other types of
-                        businesses cannot be {{ isProprietor ? 'the Proprietor' : 'a partner' }}.
-                      </p>
-
+                    <div v-if="wasBusinessSelectedFromLookup">
```
severinbeauvais, line 102: If this div (which is rendered) does nothing except hold logic then I suggest changing it to a template (which isn't rendered).

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: If we do this then we should do it for several such sections/articles, which I think should be in a separate cleanup ticket, not this ticket.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -132,6 +188,9 @@
           :activeIndex="activeIndex"
           :validate="getComponentValidate"
           :validOrgPersons="validOrgPersons"
+          :showDeliveryAddressColumn="!(isLimitedExtendRestorationFiling || isLimitedConversionRestorationFiling)"
+          :showRolesColumn="isBenBcCccUlcCorrectionFiling"
+          :showEmailColumn="isLimitedExtendRestorationFiling || isLimitedConversionRestorationFiling"
```
severinbeauvais, line 193: This is a good example of a sub-component (the list) not needing to know what type of entity/filing this is. Ultimately we might move these rules to a resource file, but not today. Looks good.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Why are you looking at the UUID like this? I don't remember other code where we need to check anything more than "does the id exist?".

On line 533, can't you simply do `if (!id)` ? (If it helps, you can check `isNaN(id)` to see whether it's a number or not.)

src/views/Restoration.vue
```diff
@@ -208,6 +222,14 @@ export default class Restoration extends Vue {
       // fetch entity snapshot
       const entitySnapshot = await this.fetchEntitySnapshot()
 
+      // set Applicant info in entitySnapshot
+      const stateFiling = entitySnapshot.businessInfo.stateFiling
+      const filing = stateFiling && await LegalServices.fetchFiling(stateFiling)
+      entitySnapshot.orgPersons = filing.restoration.parties as OrgPersonIF
```
severinbeauvais, line 228: 1. You need to handle the case where filing, or restoration, or parties is undefined/null.

2. `orgPersons` should be of type array.

3. This code will assign ALL people from the state filing into the local store. I think you should verify that you have found the applicant and only add that one to the store. Thoughts?

Suggestion:
``

tests/unit/Restoration.spec.ts
```diff
@@ -193,7 +222,21 @@ describe('Restoration component - edit page', () => {
     // create a Local Vue and install router on it
     const localVue = createLocalVue()
     localVue.use(VueRouter)
-    const router = mockRouter.mock()
+    const router: any = mockRouter.mock()
+    const store: any = getVuexStore()
```
severinbeauvais, line 226: You don't have to specify type if the functions return a type.

If a function doesn't return a type then you should do, for example: `const x = getX() as number` (or whatever type).

tests/unit/Restoration.spec.ts
```diff
@@ -208,17 +251,35 @@ describe('Restoration component - edit page', () => {
     wrapper.destroy()
   })
 
-  it('renders the page correctly', () => {
+  it('renders the page correctly', async () => {
```
severinbeauvais, line 254: Async is only needed if you have an awaited function.

severinbeauvais (commented):
Looks pretty good to me. This was a complicated one -- nice job 👍 

Remember to update the app version when you're close to merging this PR.

JazzarKarim (approved):
Looks pretty good to me after updates from other comments! Some learning for me also 👍

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
ketaki-deodhar: Before:

image

After:

image

severinbeauvais: Please test this with a change filing (in which the org/person information is not in its own section).

For example:

image

ketaki-deodhar: Good catch. I tried to test as much as possible but will test this scenario and make sure it works

jonathan-longe (approved):
Good work @ketaki-deodhar .  I understand @severinbeauvais point of view.  It's better to handle the coupling in a separate refactoring ticket.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Also, I highly recommend the use of positive logic (ie, in which scenarios do you want to display this block?) instead of negative logic (ie, for which filings does this NOT apply). The reasons are:
- easier to understand
- easier to find what _should_ be displayed instead of what _shouldn't_ be displayed -- if we add a filing type then curently this would automatically be displayed; it would be better to not see it and then need to add a condition to show it

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
ketaki-deodhar: Instead of changing style in component, handling this in Restoration.vue

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
ketaki-deodhar: isNaN(id) works. Thanks Sev! New learning for me :)

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
ketaki-deodhar: Fixed width for actions column

image

src/views/Restoration.vue
ketaki-deodhar: I refactored as you suggested. Please let me know if looks okay.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Did you test ChangeSummary (change of registration filing) and ConversionSummary (record conversion filing) to verify that they display correctly?

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -415,6 +415,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
 
   // Store getter
   @Getter getOrgPeople!: OrgPersonIF[]
+  @Getter getResource!: ResourceIF
```
severinbeauvais, line 418: 1. I don't see this type being imported...?

2. Using a resource here, while most other conditions are passed in as props, is mixing architectures. It would be better to have just 1 architecture in this component. I suggest passing in a prop for the condition you need, and in the parent you can look at the resource file flag and pass in the prop.

src/resources/Change/SP.ts
```diff
@@ -7,6 +7,7 @@ export const SpChangeResource: ResourceIF = {
   contactLabel: 'Business',
   displayName: GetCorpFullDescription(CorpTypeCd.SOLE_PROP),
   addressLabel: 'Business Addresses',
+  showEmail: true,
```
severinbeauvais, line 10: In the future, when you add a new resource like this, you might set the default value to get existing behaviour, so that you only have to impact the filings you are working on. Here, you have impacted (though low risk) some unrelated filings.

On the other hand, I do like resources/props to ENABLE a feature, as you have done here.

Also, you have already implemented this, so no change required. (Just be aware for next time, to reduce risk and testing impact.)

src/store/state/resource-model.ts
```diff
@@ -8,5 +8,6 @@ export const resourceModel: ResourceIF = {
   displayName: '',
   addressLabel: '',
   filingData: null,
-  certifyClause: ''
+  certifyClause: '',
+  showEmail: null
```
severinbeauvais, line 12: Not really required since you made it an optional property.

src/views/Restoration.vue
```diff
@@ -27,18 +27,23 @@
...
+          <v-row no-gutters class="mt-4 section-container">
+            <v-col cols="12">
+              <ListPeopleAndRoles
+                :isSummaryView="true"
+                :showDeliveryAddressColumn="false"
+                :showRolesColumn="false"
+                :showEmailColumn="true"
+              />
+            </v-col>
+          </v-row>
```
severinbeauvais, line 46: Why make this a single column in a single row? Can it be a simple div?

src/views/Restoration.vue
```diff
@@ -225,10 +231,17 @@ export default class Restoration extends Vue {
       // set Applicant info in entitySnapshot
       const stateFiling = entitySnapshot.businessInfo.stateFiling
       const filing = stateFiling && await LegalServices.fetchFiling(stateFiling)
-      entitySnapshot.orgPersons = filing.restoration.parties as OrgPersonIF
-      entitySnapshot.orgPersons.forEach(o => {
-        o.officer.id = uuidv4()
-      })
+      if (!filing) {
+        throw new Error(`Invalid fetched stateFiling = ${this.getBusinessId}`)
+      } else {
```
severinbeauvais, line 236: You don't need the else since the thrown error will return from this code block.

src/views/Restoration.vue
```diff
@@ -225,10 +231,17 @@ export default class Restoration extends Vue {
...
+        const parties = filing.restoration.parties
+        // filter applicant from fetched parties
+        const applicant = parties.filter(
+          orgPerson => orgPerson.roles.some(role => role.roleType === RoleTypes.APPLICANT)
+        )
+        console.log('applicant', applicant)
```
severinbeauvais, line 242: Remove debugging code before final commit.

src/views/Restoration.vue
```diff
@@ -225,10 +231,17 @@ export default class Restoration extends Vue {
...
       const stateFiling = entitySnapshot.businessInfo.stateFiling
       const filing = stateFiling && await LegalServices.fetchFiling(stateFiling)
-      entitySnapshot.orgPersons = filing.restoration.parties as OrgPersonIF
-      entitySnapshot.orgPersons.forEach(o => {
-        o.officer.id = uuidv4()
-      })
+      if (!filing) {
+        throw new Error(`Invalid fetched stateFiling = ${this.getBusinessId}`)
+      } else {
+        const parties = filing.restoration.parties
```
severinbeauvais, line 237: `restoration` could still be undefined or null. Try this:
`const parties = filing.restoration?.parties || []`

src/views/Restoration.vue
```diff
@@ -225,10 +231,17 @@ export default class Restoration extends Vue {
...
+        // filter applicant from fetched parties
+        const applicant = parties.filter(
```
severinbeauvais, line 239: Variables that are arrays should be plural.

src/views/Restoration.vue
```diff
@@ -225,10 +231,17 @@ export default class Restoration extends Vue {
...
+          orgPerson => orgPerson.roles.some(role => role.roleType === RoleTypes.APPLICANT)
```
severinbeauvais, line 240: You could also use `find` to get the first applicant in the array. It might simplify a few things for you.

src/views/Restoration.vue
```diff
@@ -271,6 +284,27 @@ export default class Restoration extends Vue {
     this.$nextTick(() => this.setHaveUnsavedChanges(false))
   }
 
+  // build applicant orgPerson and assign id (uuid)
+  private parseApplicantOrgPerson (applicant: OrgPersonIF[]): OrgPersonIF[] {
```
severinbeauvais, line 288: On line 243, you pass in a single object (applicant[0]). but here you declare the formal parameter as an array.

src/views/Restoration.vue
```diff
@@ -271,6 +284,27 @@ export default class Restoration extends Vue {
...
+    let applicantOrgPerson: Array<OrgPersonIF> = []
```
severinbeauvais, line 289: You could make this a const since you don't reassign it.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
ketaki-deodhar: Yes. screenshots to compare from DEV and my LOCAL.

Change Filing
DEV:

image

 LOCAL:

image


Correction Filing
DEV:

image

LOCAL:

image

src/views/Restoration.vue
ketaki-deodhar: yes. you are right . Done

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Awesome! Good job. Please add a "Test Notes" comment to your ticket to call this out as something you changed but should look the same (for UX and QA to verify).

PS Record Conversion is not the same as Correction, but the screenshots are correct 👍

severinbeauvais: You can resolve this conversation if you like. I'm satisfied 😀

src/views/Restoration.vue
ketaki-deodhar: Yes

src/resources/Change/SP.ts
ketaki-deodhar: Although this is low impact, I will change the implementation based on your suggestion (^^ comment). I think it is unnecessary to change unrelated files and can be handled with default value and prop..

severinbeauvais: Sounds good.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
ketaki-deodhar: sounds good! thanks

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: PS Thanks for changing the section titlebar/style. I didn't notice that the screenshot doesn't match the uxpin, but I'm sure your new version is good 👍

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: I want to see what you've changed here.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -413,20 +422,21 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   @Getter isFirmChangeFiling!: boolean
   @Getter isFirmConversionFiling!: boolean
   @Getter isFirmCorrectionFiling!: boolean
+  @Getter isLimitedExtendRestorationFiling!: boolean
+  @Getter isLimitedConversionRestorationFiling!: boolean
   @Getter isRoleStaff!: boolean
   @Getter hideChangeButtonForSoleProps!: boolean
```
ketaki-deodhar, line 429: @severinbeauvais, somehow I do not see your comment captured in screenshot below. The change was to not add ResourceIF in this component and instead use prop and let parent use resource file to ge t the value

image

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Give me a few minutes to review your latest code. I'll update this conversation accordingly.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: This component looks good to me. You can resolve this conversation.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
ketaki-deodhar: Here is the screenshot with changes:

image

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -1,146 +1,214 @@
...
-        <label class="font-weight-bold pl-2">{{ orgPersonLabel }}</label>
-      </div>
-
-      <!-- Instructional people and roles text (base corrections only)-->
-      <article v-if="isBenBcCccUlcCorrectionFiling" class="section-container">
-        This application must include the following:
+    <!-- Restoration conversion and extension add buttons -->
+    <div v-if="(isLimitedConversionRestorationFiling || isLimitedExtendRestorationFiling)">
+      <article>
+        <header>
```
severinbeauvais, line 6: @jonathan-longe Ketaki and I discussed this earlier today and I approve this "duplication".

Part of this is due to the different display style needed, and part of it is because of the different applicant rules, between filings.

The alternative would be a wrapper to handle the style differences, and still some conditionals (possibly resource driven) to handle the rest. It still wouldn't be super-clean, and we don't know what the future looks like because you are still designing it :)

What do you think -- is this OK for now?

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -1,146 +1,214 @@
...
+            [{ roleType: RoleTypes.APPLICANT, appointmentDate: appointmentDate }],
+            PartyTypes.ORGANIZATION
+          )"
+        >
+          <v-icon>mdi-domain-plus</v-icon>
+          <span>Add a Business or Corporation</span>
+        </v-btn>
+        <!-- <p v-if="!hasApplicant" class="error-text small-text mt-5 mb-0">
+          You must have one applicant
+        </p> -->
```
severinbeauvais, line 56: Ketaki, why is this commented out -- not needed or not working as desired?

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -340,6 +419,11 @@ export default class PeopleAndRoles extends Mixins(CommonMixin, DateMixin, OrgPe
     return this.dateToYyyyMmDd(this.getCurrentJsDate)
   }
 
+  /** flag to show email under name. */
+  get showEmailUnderName (): string {
```
severinbeauvais, line 423: Return type should be boolean.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -383,7 +467,8 @@ export default class PeopleAndRoles extends Mixins(CommonMixin, DateMixin, OrgPe
     this.currentOrgPerson.actions = actions
 
     // for firms, use business lookup initially
-    if (this.isPartnership || this.isSoleProp) {
+    if (this.isPartnership || this.isSoleProp || isLimitedExtendRestorationFiling ||
```
severinbeauvais, line 470: Update comment on line 469.

src/resources/Correction/BC.ts
```diff
@@ -8,6 +8,7 @@ export const BcCorrectionResource: ResourceIF = {
   displayName: null, // not used
   entityType: CorpTypeCd.BC_COMPANY,
   addressLabel: 'Registered Office',
+
```
severinbeauvais, line 11: Undo this. Also in next file.

src/views/Restoration.vue
```diff
@@ -27,7 +27,22 @@
 
         <YourCompanySummary class="mt-10" />
 
-        <CurrentDirectors class="mt-10" />
+        <!-- Applicant list -->
+        <v-card id="people-and-roles-vcard" flat class="mt-6">
+          <!-- Header -->
+          <div class="section-container header-container">
+            <v-icon color="appDkBlue">mdi-account-multiple-plus</v-icon>
+            <label class="font-weight-bold pl-2">{{ orgPersonLabel }} Informaton</label>
```
severinbeauvais, line 35: Typo.

src/views/Restoration.vue
```diff
@@ -207,6 +225,26 @@ export default class Restoration extends Vue {
 
       // fetch entity snapshot
       const entitySnapshot = await this.fetchEntitySnapshot()
+      console.log('entitySnapshot', entitySnapshot)
+
+      // set Applicant info in entitySnapshot
```
severinbeauvais, line 230: Comment doesn't relate to next line (or block) of code.

src/views/Restoration.vue
```diff
@@ -207,6 +225,26 @@ export default class Restoration extends Vue {
...
+      const stateFiling = entitySnapshot.businessInfo.stateFiling
+
+      const filing = stateFiling && await LegalServices.fetchFiling(stateFiling)
+      if (!filing) {
+        throw new Error(`Invalid fetched stateFiling = ${this.getBusinessId}`)
+      }
+      console.log('filing', filing)
```
severinbeauvais, line 237: Remove debugging code before merging.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: LGTM. Jonathan, your conversation to resolve when you are satisfied.

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: OK, this component no longer looks at the resource file. Now it uses props for the new logic. 👍

src/views/Restoration.vue
severinbeauvais: If no parties were provided, does this filing fail gracefully? Specifically, can the user enter a new applicant if there is no previous one? (You could temporarily hack this code to test it locally.)

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
ketaki-deodhar: I will remove once confirm we do not need it with Ethan.

src/views/Restoration.vue
```diff
@@ -207,6 +225,26 @@ export default class Restoration extends Vue {
...
+
+      const parties = filing.restoration?.parties || []
+
+      // find applicant from fetched parties
+      const applicants = parties.find(
+        orgPerson => orgPerson.roles.some(role => role.roleType === RoleTypes.APPLICANT)
+      )
+
+      // set applciant orgPerson
```
severinbeauvais, line 246: Typo.

src/views/Restoration.vue
severinbeauvais: There's still a problem here. On line 247, you pass an array that may have 0 or 1 or more applicants (hey, you never know).

Here (eg, line 294), you treat that array as a single item. JavaScript isn't giving you an error (and may even work if there is a single item in the array) but it's still incorrect code.

I recommend you use "find" to find the first applicant in the parties list, and then only add it to the entity snapshot if "find" found one.

If "find" found no items, how does the rest of the code handle it ... can the user add a new applicant?

Ref: https://www.w3schools.com/jsref/jsref_find.asp

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Does this assume we have the original applicant? What if the API didn't provide it, or the user deleted it -- does this code still function correctly? (fail gracefully)

ketaki-deodhar: If Removed, show red cross icon (invalid) instead of displaying above message. I just need to confirm with Ethan and remove it

image

src/views/Restoration.vue
```diff
@@ -207,6 +225,26 @@ export default class Restoration extends Vue {
...
+
+      // find first applicant from fetched parties
+      const applicants = parties.find(
+        orgPerson => orgPerson.roles.some(role => role.roleType === RoleTypes.APPLICANT)
+      )
+
+      if (applicants === undefined) {
+        throw new Error(`Applicant not found for ${this.getBusinessId}`)
+      }
+
```
ketaki-deodhar, line 245: this will handle if parties and applicants are undefined. I tested it and error thrown will be like below: 

image

src/views/Restoration.vue
ketaki-deodhar: Please see comment on line: 245

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Got it. Sounds good. I'm leaving this with you.

Resolving this conversation.

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Great! You can resolve this conversation if you want.

src/views/Restoration.vue
severinbeauvais: Sorry, I got confused a bit.
- on line 238, you should change applicants -> applicant (since it's an object not an array)
- on line 291, you should change applicant type to OrgPersonIF (since it's an object not an array)

There rest looks OK 👍

ketaki-deodhar: done

severinbeauvais (approved):
👍

severinbeauvais (approved):
LGTM

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
jonathan-longe: @severinbeauvais and @ketaki-deodhar I hear you.  Let's leave this as is for now.  When we have some time for refactoring, I would suggest we create components for each filing type rather than adding `v-if="filingTypeXXX`.  Extending `PeopleAndRoles` common component to handle every filing type makes the common component more complicated and more difficult to read.  Worse, when we modify a common component to handle a new filing we risk breaking existing functionality that may not get tested.

severinbeauvais: Refactoring this will be an interesting exercise.

In Create UI, there are separate PeopleAndRoles components (for different filing types) but a common mixin as they share a lot of code. That was pretty easy to work on, so maybe that's a pattern we could follow here (future).

PR #453, opened 2023-03-15
15092 - add more restoration endpoints

jonathan-longe:
*Issue #:* /bcgov/entity#15092

Add the following endpoints:  

- /limitedRestorationToFull
- /limitedRestorationExtension
- /fullRestoration
- /limitedRestoration

By separating restorations to different endpoints, we can keep the code cleaner and easier to read.  

### Caution, this change will break existing functionality until business-filings-ui makes a similar change.

### DIFF
--- a/src/enums/routeNames.ts
+++ b/src/enums/routeNames.ts
@@ -4,7 +4,10 @@ export enum RouteNames {
   CHANGE = 'change',
   CONVERSION = 'conversion',
   CORRECTION = 'correction',
-  RESTORATION = 'restoration',
+  RESTORATION_FULL = 'fullRestoration',
+  RESTORATION_LIMITED = 'limitedRestoration',
+  RESTORATION_EXTENSION = 'limitedRestorationExtension',
+  RESTORATION_CONVERSION = 'limitedRestorationToFull',
   SIGN_IN = 'signin',
   SIGN_OUT = 'signout',
   SPECIAL_RESOLUTION = 'special-resolution'
--- a/src/enums/routeNames.ts
+++ b/src/enums/routeNames.ts
@@ -4,8 +4,6 @@ export enum RouteNames {
   CHANGE = 'change',
   CONVERSION = 'conversion',
   CORRECTION = 'correction',
-  RESTORATION_FULL = 'fullRestoration',
-  RESTORATION_LIMITED = 'limitedRestoration',
   RESTORATION_EXTENSION = 'limitedRestorationExtension',
   RESTORATION_CONVERSION = 'limitedRestorationToFull',
   SIGN_IN = 'signin',
--- a/src/mixins/common-mixin.ts
+++ b/src/mixins/common-mixin.ts
@@ -145,10 +145,8 @@ export default class CommonMixin extends Vue {
       case RouteNames.CHANGE: return 'Business Information'
       case RouteNames.CONVERSION: return 'Record Conversion'
       case RouteNames.CORRECTION: return 'Register Correction'
-      case RouteNames.RESTORATION: {
-        if (this.isLimitedExtendRestorationFiling) return 'Limited Restoration Extension'
-        if (this.isLimitedConversionRestorationFiling) return 'Conversion to Full Restoration'
-      }
+      case RouteNames.RESTORATION_EXTENSION: return 'Limited Restoration Extension'
+      case RouteNames.RESTORATION_CONVERSION: return 'Conversion to Full Restoration'
     }
     return 'Unknown Filing' // should never happen
   }
--- a/src/router/routes.ts
+++ b/src/router/routes.ts
@@ -44,8 +44,38 @@ export const routes = [
     }
   },
   {
-    path: '/restoration',
-    name: RouteNames.RESTORATION,
+    path: '/limitedRestorationToFull',
+    name: RouteNames.RESTORATION_CONVERSION,
+    component: Restoration,
+    meta: {
+      requiresAuth: true,
+      isStaffOnly: true,
+      filingType: FilingTypes.RESTORATION
+    }
+  },
+  {
+    path: '/limitedRestorationExtension',
+    name: RouteNames.RESTORATION_EXTENSION,
+    component: Restoration,
+    meta: {
+      requiresAuth: true,
+      isStaffOnly: true,
+      filingType: FilingTypes.RESTORATION
+    }
+  },
+  {
+    path: '/fullRestoration',
+    name: RouteNames.RESTORATION_FULL,
+    component: Restoration,
+    meta: {
+      requiresAuth: true,
+      isStaffOnly: true,
+      filingType: FilingTypes.RESTORATION
+    }
+  },
+  {
+    path: '/limitedRestoration',
+    name: RouteNames.RESTORATION_LIMITED,
     component: Restoration,
     meta: {
       requiresAuth: true,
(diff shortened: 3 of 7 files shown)


src/mixins/common-mixin.ts
```diff
@@ -145,10 +145,8 @@ export default class CommonMixin extends Vue {
       case RouteNames.CHANGE: return 'Business Information'
       case RouteNames.CONVERSION: return 'Record Conversion'
       case RouteNames.CORRECTION: return 'Register Correction'
-      case RouteNames.RESTORATION: {
-        if (this.isLimitedExtendRestorationFiling) return 'Limited Restoration Extension'
-        if (this.isLimitedConversionRestorationFiling) return 'Conversion to Full Restoration'
-      }
+      case RouteNames.RESTORATION_EXTENSION: return 'Limited Restoration Extension'
+      case RouteNames.RESTORATION_CONVERSION: return 'Conversion to Full Restoration'
```
jonathan-longe, line 149: I'm not sure we need this method.  As far as I can see, it's only used to set the title in the view.  I'd suggest we'd be better to add the text directly to the view.

src/router/routes.ts
```diff
@@ -44,8 +44,38 @@ export const routes = [
...
-    name: RouteNames.RESTORATION,
+    path: '/limitedRestorationToFull',
+    name: RouteNames.RESTORATION_CONVERSION,
+    component: Restoration,
+    meta: {
+      requiresAuth: true,
+      isStaffOnly: true,
+      filingType: FilingTypes.RESTORATION
+    }
+  },
```
jonathan-longe, line 55: For backwards compatibility, all the new endpoints point to the same component.  Once this PR is approved, we can point the endpoints to a dedicated view.

src/enums/routeNames.ts
```diff
@@ -4,7 +4,10 @@ export enum RouteNames {
   CHANGE = 'change',
   CONVERSION = 'conversion',
   CORRECTION = 'correction',
-  RESTORATION = 'restoration',
+  RESTORATION_FULL = 'fullRestoration',
+  RESTORATION_LIMITED = 'limitedRestoration',
+  RESTORATION_EXTENSION = 'limitedRestorationExtension',
+  RESTORATION_CONVERSION = 'limitedRestorationToFull',
```
JazzarKarim, line 10: Just something super small, do you think it would be better to name this `'limitedRestorationConversion'`? This way we would be consistent with the naming (sorry for being nitpicky).

severinbeauvais: I noticed this too but it would require a small change to the scheme and API.

@argush3 @vysakh-menon @kzdev420 What do you think? Is that something you small we should do, something Kevin or Karim could do?

severinbeauvais: Oh wait -- I was thinking this was the filing subtype field (not the route). I have no issue with whether you want to keep or change this.

src/enums/routeNames.ts
```diff
@@ -4,7 +4,10 @@ export enum RouteNames {
   CHANGE = 'change',
   CONVERSION = 'conversion',
   CORRECTION = 'correction',
-  RESTORATION = 'restoration',
+  RESTORATION_FULL = 'fullRestoration',
+  RESTORATION_LIMITED = 'limitedRestoration',
```
severinbeauvais, line 8: This app doesn't handle these 2 filings, so you don't need these routes.

src/mixins/common-mixin.ts
severinbeauvais: It's here as a common pattern (see methods above and also in other apps), but in the vein of having all filing-specific items in a single place (view), I'm OK with moving this (now or later).

src/enums/routeNames.ts
ketaki-deodhar: I was thinking the same. Extension and conversion are sub-types. Wondering if this changes anything in terms of initial load of the draft.

argush3: Old world describes the conversion filing as "Restoration Application (Convert Limited to Full)".  We can update it and I do think it's something Karim or Kevin can work on.

As for whether we should do it, I don't think it's required unless people really feel it's confusing.

src/router/routes.ts
```diff
@@ -44,8 +44,38 @@ export const routes = [
...
+    name: RouteNames.RESTORATION_EXTENSION,
+    component: Restoration,
+    meta: {
+      requiresAuth: true,
+      isStaffOnly: true,
+      filingType: FilingTypes.RESTORATION
+    }
+  },
+  {
+    path: '/fullRestoration',
```
severinbeauvais, line 67: What's the plan here, since Edit UI doesn't process full or limited resto filings?

src/router/routes.ts
```diff
@@ -44,8 +44,38 @@ export const routes = [
     }
   },
   {
-    path: '/restoration',
-    name: RouteNames.RESTORATION,
+    path: '/limitedRestorationToFull',
```
severinbeauvais, line 47: Is there a possible shorter path (route) name?

severinbeauvais (commented):
Looks fine but why are we doing this? How will it keep the code cleaner and easier to read?

src/enums/routeNames.ts
jonathan-longe: Good call.  I'll remove them

src/router/routes.ts
jonathan-longe: That's my mistake.  I'll remove those routes.

src/router/routes.ts
jonathan-longe: Yes, we can definitely shorten them, but then it means adding more complicated logic to filings-ui.  At present, it's nice and simple:

```javascript
if (applicationName === ApplicationTypes.EDIT_UI) {
  url = `${this.getEditUrl}${this.getIdentifier}/` + restorationType + `?restoration-id=${id}`
}
```

src/enums/routeNames.ts
severinbeauvais: If we're going to change the subtypes then now is the second-cheapest time to do it. Any strong opinions on whether we should do it?

jonathan-longe:
> Looks fine but why are we doing this? How will it keep the code cleaner and easier to read?

It's becoming clear that limited restorations conversions will have different logic from limited restoration extensions.  For example, as Karim has mentioned in [455,](https://github.com/bcgov/business-edit-ui/pull/455) limited restoration extensions require a call to the stateFiling endpoint.  We could add multiple `v-if` but I think it would be better to use separate endpoints and separate the logic flow once at the route level.

severinbeauvais:
> > Looks fine but why are we doing this? How will it keep the code cleaner and easier to read?
> 
> It's becoming clear that limited restorations conversions will have different logic from limited restoration extensions. For example, as Karim has mentioned in [455,](https://github.com/bcgov/business-edit-ui/pull/455) limited restoration extensions require a call to the stateFiling endpoint. We could add multiple `v-if` but I think it would be better to use separate endpoints and separate the logic flow once at the route level.

I like that answer 👍

src/mixins/common-mixin.ts
severinbeauvais: The pattern is having helper functions in a mixin to convert a type/route/etc to a string, for example.

severinbeauvais (approved):
OK from me to proceed once conversations are resolved and unit tests are fixed.

src/router/routes.ts
severinbeauvais: The subtype-renaming conversation above did not identify a strong need to change anything, so THIS change is also OK (since the route name and path is the subtype name).

PR #455, opened 2023-03-15
refactor store methods to use conventional format

jonathan-longe:
*Issue #:* /bcgov/entity#15092

Refactoring the store methods to use conventional format.  

### Actions
Instead of exporting individual actions and returning a void like this:
```javascript
export const setEntityType: ActionIF = ({ commit }, entityType: CorpTypeCd): void => {
  commit('mutateEntityType', entityType)
}
```

They're now wrapped in an `export default { ... }` and return promises as designed like this:
```javascript
export default {
  setEntityType ({ commit }, entityType: CorpTypeCd) {
    commit('mutateEntityType', entityType)
  },
 // many more
}
```

### Mutations and Getters

Similar for mutations and getters.  Instead of exporting individual methods like this:
```javascript
export const mutateBusinessId = (state: StateIF, businessId: string) => {
  state.stateModel.tombstone.businessId = businessId
}
```

We wrap them all in an `export default { ... }` like this:
```javascript
export default {
  mutateBusinessId (state: StateIF, businessId: string) {
    state.stateModel.tombstone.businessId = businessId
  },
  // many more
}
```


JazzarKarim (approved):
As we discussed yesterday, LGTM! Let's wait for other reviewers and see what they think. Thank you Jonathan 😃 This will help me add the action of setting the previously filed restoration in the store.

severinbeauvais:
> This will help me add the action of setting the previously filed restoration in the store.

I haven't looked through this PR yet, but I know @ketaki-deodhar is storing the state filing in a restoration extension/conversion filing. Is that the same thing you need to do, Karim?

JazzarKarim:
> > This will help me add the action of setting the previously filed restoration in the store.
> 
> I haven't looked through this PR yet, but I know @ketaki-deodhar is storing the state filing in a restoration extension/conversion filing. Is that the same thing you need to do, Karim?

Yes. I have an action that sets the state filing in the store.
```javascript
export const setLimitedFullRestoration: ActionIF = (context): void => {
  LegalServices.fetchFiling(context.getters.getStateFilingUrl)
    .then((response) => {
      context.commit('mutateLimitedFullRestoration', response.restoration)
    })
}
```
I need this to get all the previously filled in inputs like the expiry date and approval type for example.

severinbeauvais:
> _Issue #:_ /[bcgov/entity#15092](https://github.com/bcgov/entity/issues/15092)
> 
> They're now wrapped in an `export default { ... }` and return promises as designed like this:
> 
> ```js
> export default {
>   setEntityType ({ commit }, entityType: CorpTypeCd) {
>     commit('mutateEntityType', entityType)
>   },
>  // many more
> }

Well, technically, this doesn't return anything, but I do prefer this function style better.

Would it be possible to `export default class Actions` so that we don't need the commas between the methods?

src/store/actions/actions-model.ts
severinbeauvais, line 1: Yes, I am happy to see the actions/getters/etc moved from sub-folders to the store root 👍

src/store/index.ts
```diff
@@ -4,22 +4,20 @@ import Vuex from 'vuex'
...
 
-  const store = new Vuex.Store<any>({
+  return new Vuex.Store<any>({
     state: { ...States },
-    getters: { ...Getters },
-    mutations: { ...Mutations },
-    actions: { ...Actions }
+    getters,
+    mutations,
+    actions
```
severinbeauvais, line 21: I'll grant you that this is cleaner :)

severinbeauvais (approved):

jonathan-longe:
> Would it be possible to `export default class Actions` so that we don't need the commas between the methods?

Hasn't the Class API had been deprecated in Vue3?  My preference would be to use whatever approach makes our upgrade path the easiest.

severinbeauvais:
> > Would it be possible to `export default class Actions` so that we don't need the commas between the methods?
> 
> Hasn't the Class API had been deprecated in Vue3? My preference would be to use whatever approach makes our upgrade path the easiest.

Umm, I don't know. But I would think "class API" refers to Vue components, not generic classes.

This is not a deal-breaker for me.

PR #460, opened 2023-03-23
Extend limited restoration time limit

JazzarKarim:
*Issue #:* /bcgov/entity#14974

*Description of changes:*
- Added the extension time limit section
- Added extension section in review and certify page
- To summarize and show all the new components/work, here's a small demo:

https://user-images.githubusercontent.com/122301442/227372318-d75062c2-a2a4-4b44-9b33-66cfec0dc119.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "@babel/compat-data": "^7.19.1",
     "@bcrs-shared-components/action-chip": "1.1.5",
-    "@bcrs-shared-components/approval-type": "1.0.15",
+    "@bcrs-shared-components/approval-type": "1.0.16",
     "@bcrs-shared-components/breadcrumb": "2.1.5",
     "@bcrs-shared-components/business-lookup": "1.1.7",
     "@bcrs-shared-components/certify": "2.1.5",
@@ -26,7 +26,7 @@
     "@bcrs-shared-components/court-order-poa": "2.1.3",
     "@bcrs-shared-components/date-picker": "1.2.5",
     "@bcrs-shared-components/detail-comment": "1.1.1",
-    "@bcrs-shared-components/enums": "1.0.27",
+    "@bcrs-shared-components/enums": "1.0.36",
     "@bcrs-shared-components/fee-summary": "1.2.5",
     "@bcrs-shared-components/folio-number": "1.1.1",
     "@bcrs-shared-components/help-business-number": "1.1.1",
--- a/src/components/Restoration/ExtendTimeLimit.vue
+++ b/src/components/Restoration/ExtendTimeLimit.vue
@@ -101,7 +101,7 @@ export default class ExtendTimeLimit extends Vue {
   }
 
   /** The remaining number of months left for the previously filed limited restoration. */
-  get previousNumberOfMonths (): string {
+  get previousNumberOfMonths (): number {
     return this.subtractDates(this.getCurrentDate, this.getStateFilingRestoration?.expiry)
   }
 }
--- a/src/components/Restoration/ExtendTimeLimit.vue
+++ b/src/components/Restoration/ExtendTimeLimit.vue
@@ -5,11 +5,11 @@
     <section
       id="extension-time-section"
       class="section-container"
-      :class="{ 'invalid-section': invalidSectionExpiry }"
+      :class="{ 'invalid-section': !getExpiryValid }"
     >
       <v-row no-gutters>
         <v-col cols="12" sm="3">
-          <strong :class="{ 'error-text': invalidSectionExpiry }">Extension Time</strong>
+          <strong :class="{ 'error-text': !getExpiryValid }">Extension Time</strong>
         </v-col>
         <v-col cols="12" sm="9">
           <!-- Limited Restoration Radio Panel -->
@@ -28,14 +28,14 @@
     <!-- Approval Type Section (if applicable) -->
     <section id="approval-type-section"
       class="section-container"
-      :class="{ 'invalid-section': invalidSectionApprovalType }"
+      :class="{ 'invalid-section': !getApprovalTypeValid }"
     >
       <template v-if="approvalType == ApprovalTypes.VIA_COURT_ORDER">
         <ApprovalType
           :courtOrderNumber="courtOrderNumberText"
           :isCourtOrderOnly="true"
           :isCourtOrderRadio="false"
-          :invalidSection="invalidSectionApprovalType"
+          :invalidSection="!getApprovalTypeValid"
           @courtNumberChange="setRestorationCourtOrder({ fileNumber: $event })"
           @valid="setApprovalTypeValid($event)"
         />
@@ -87,7 +87,7 @@ export default class ExtendTimeLimit extends Vue {
 
   /** The limited restoration state filing's approval type. */
   get approvalType (): ApprovalTypes {
-    return this.getStateFilingRestoration.approvalType
+    return this.getStateFilingRestoration?.approvalType
   }
 
   /** The court order draft file number. */
@@ -100,19 +100,9 @@ export default class ExtendTimeLimit extends Vue {
     return DateUtilities.subtractMonthsToDate(this.previousNumberOfMonths, this.getRestoration.expiry)
   }
 
-  /** This extension time's validity state (when prompted by app). */
-  get invalidSectionExpiry (): boolean {
-    return !this.getExpiryValid
-  }
-
-  /** This approval type's validity state (when prompted by app). */
-  get invalidSectionApprovalType (): boolean {
-    return !this.getApprovalTypeValid
-  }
-
-  /** The previously selected number of months of limited restoration. */
+  /** The remaining number of months left for the previously filed limited restoration. */
   get previousNumberOfMonths (): string {
-    return this.subtractDates(this.getCurrentDate, this.getStateFilingRestoration.expiry)
+    return this.subtractDates(this.getCurrentDate, this.getStateFilingRestoration?.expiry)
   }
 }
 </script>
--- a/src/components/Restoration/ExtendTimeLimit.vue
+++ b/src/components/Restoration/ExtendTimeLimit.vue
@@ -18,8 +18,8 @@
             :expiryDate="expiry"
             :key="expiry"
             :maxNumberOfMonths=36
-            @expiry="setRestorationExpiry(addMonthsToDate(previousNumberOfMonths, $event))"
-            @valid="setExpiryValid($event)"
+            @expiry="mutateRestorationExpiry(addMonthsToDate(previousNumberOfMonths, $event))"
+            @valid="mutateExpiryValid($event)"
           />
         </v-col>
       </v-row>
@@ -36,8 +36,8 @@
           :isCourtOrderOnly="true"
           :isCourtOrderRadio="false"
           :invalidSection="!getApprovalTypeValid"
-          @courtNumberChange="setRestorationCourtOrder({ fileNumber: $event })"
-          @valid="setApprovalTypeValid($event)"
+          @courtNumberChange="mutateRestorationCourtOrder({ fileNumber: $event })"
+          @valid="mutateApprovalTypeValid($event)"
         />
       </template>
     </section>
@@ -46,7 +46,7 @@
 
 <script lang="ts">
 import Vue from 'vue'
-import { Action, Getter } from 'vuex-class'
+import { Action, Getter, Mutation } from 'vuex-class'
 import { ApprovalType } from '@bcrs-shared-components/approval-type'
 import { ApprovalTypes, RestorationTypes } from '@/enums'
 import Actions from '@/components/common/Actions.vue'
@@ -75,11 +75,12 @@ export default class ExtendTimeLimit extends Vue {
   @Getter getRestoration!: RestorationStateIF
   @Getter getStateFilingRestoration!: StateFilingRestorationIF
 
-  @Action setApprovalTypeValid!: ActionBindingIF
-  @Action setExpiryValid!: ActionBindingIF
-  @Action setRestorationCourtOrder!: ActionBindingIF
-  @Action setRestorationExpiry!: ActionBindingIF
   @Action setValidComponent!: ActionBindingIF
+
+  @Mutation mutateApprovalTypeValid!: ActionBindingIF
+  @Mutation mutateExpiryValid!: ActionBindingIF
+  @Mutation mutateRestorationCourtOrder!: ActionBindingIF
+  @Mutation mutateRestorationExpiry!: ActionBindingIF
 
   // Enum for template
   readonly ApprovalTypes = ApprovalTypes
--- a/src/components/Restoration/ExtendTimeLimit.vue
+++ b/src/components/Restoration/ExtendTimeLimit.vue
@@ -1,0 +1,131 @@
+<template>
+  <div id="extend-time-limit" class="pt-2">
+
+    <!-- Extension Time Section -->
+    <section
+      id="extension-time-section"
+      class="section-container"
+      :class="{ 'invalid-section': invalidSectionExpiry }"
+    >
+      <v-row no-gutters>
+        <v-col cols="12" sm="3">
+          <strong :class="{ 'error-text': invalidSectionExpiry }">Extension Time</strong>
+        </v-col>
+        <v-col cols="12" sm="9">
+          <!-- Limited Restoration Radio Panel -->
+          <LimitedRestorationPanel
+            :currentDate="getCurrentDate"
+            :expiryDate="expiry"
+            :key="expiry"
+            :maxNumberOfMonths=36
+            @expiry="setRestorationExpiry(addMonthsToDate(previousNumberOfMonths, $event))"
+            @valid="setExpiryValid($event)"
+          />
+        </v-col>
+      </v-row>
+    </section>
+
+    <!-- Approval Type Section (if applicable) -->
+    <section id="approval-type-section"
+      class="section-container"
+      :class="{ 'invalid-section': invalidSectionApprovalType }"
+    >
+      <template v-if="approvalType == ApprovalTypes.VIA_COURT_ORDER">
+        <ApprovalType
+          :courtOrderNumber="courtOrderNumberText"
+          :isCourtOrderOnly="true"
+          :isCourtOrderRadio="false"
+          :invalidSection="invalidSectionApprovalType"
+          @courtNumberChange="setRestorationCourtOrder({ fileNumber: $event })"
+          @valid="setApprovalTypeValid($event)"
+        />
+      </template>
+    </section>
+  </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Action, Getter } from 'vuex-class'
+import { ApprovalType } from '@bcrs-shared-components/approval-type'
+import { ApprovalTypes, RestorationTypes } from '@/enums'
+import Actions from '@/components/common/Actions.vue'
+import { Component, Watch } from 'vue-property-decorator'
+import { DateMixin, CommonMixin } from '@/mixins'
+import DateUtilities from '@/services/date-utilities'
+import { LegalServices } from '@/services/'
+import { LimitedRestorationPanel } from '@bcrs-shared-components/limited-restoration-panel'
+import { RestorationStateIF, StateFilingRestorationIF } from '@/interfaces'
+
+@Component({
+  mixins: [
+    CommonMixin,
+    DateMixin
+  ],
+  components: {
+    Actions,
+    ApprovalType,
+    LimitedRestorationPanel
+  }
+})
+export default class ExtendTimeLimit extends Vue {
+  @Getter getApprovalTypeValid!: boolean
+  @Getter getCurrentDate!: string
+  @Getter getExpiryValid!: boolean
+  @Getter getRestoration!: RestorationStateIF
+  @Getter getStateFilingRestoration!: StateFilingRestorationIF
+
+  @Action setApprovalTypeValid!: ActionBindingIF
+  @Action setExpiryValid!: ActionBindingIF
+  @Action setRestorationCourtOrder!: ActionBindingIF
+  @Action setRestorationExpiry!: ActionBindingIF
+  @Action setValidComponent!: ActionBindingIF
+
+  // Enum for template
+  readonly ApprovalTypes = ApprovalTypes
+  readonly RestorationTypes = RestorationTypes
+
+  /** The limited restoration state filing's approval type. */
+  get approvalType (): ApprovalTypes {
+    return this.getStateFilingRestoration.approvalType
+  }
+
+  /** The court order draft file number. */
+  get courtOrderNumberText (): string {
+    return this.getRestoration.courtOrder?.fileNumber || ''
+  }
+
+  /** The expiry draft date for extension. */
+  get expiry (): string {
+    return DateUtilities.subtractMonthsToDate(this.previousNumberOfMonths, this.getRestoration.expiry)
+  }
+
+  /** This extension time's validity state (when prompted by app). */
+  get invalidSectionExpiry (): boolean {
+    return !this.getExpiryValid
+  }
+
+  /** This approval type's validity state (when prompted by app). */
+  get invalidSectionApprovalType (): boolean {
+    return !this.getApprovalTypeValid
+  }
+
+  /** The previously selected number of months of limited restoration. */
+  get previousNumberOfMonths (): string {
+    return this.subtractDates(this.getCurrentDate, this.getStateFilingRestoration.expiry)
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+:deep() {
+  // Fix font of radio buttons.
+  .radio-button {
+    label {
+      font-weight: normal;
+      color: $gray7;
+    }
+  }
+}
+</style>
--- a/src/components/Restoration/ExtendTimeLimit.vue
+++ b/src/components/Restoration/ExtendTimeLimit.vue
@@ -1,5 +1,5 @@
 <template>
-  <div id="extend-time-limit" class="pt-2">
+  <v-card flat id="extend-time-limit" class="pt-2 mr-8">
 
     <!-- Extension Time Section -->
     <section
@@ -41,7 +41,7 @@
         />
       </template>
     </section>
-  </div>
+  </v-card>
 </template>
 
 <script lang="ts">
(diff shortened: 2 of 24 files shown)


src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -0,0 +1,131 @@
+<template>
+  <v-card flat id="extend-time-limit" class="pt-2 mr-8">
+
+    <!-- Extension Time Section -->
+    <section
+      id="extension-time-section"
+      class="section-container"
+      :class="{ 'invalid-section': invalidSectionExpiry }"
```
JazzarKarim, line 8: Show red border when invalid. Ditto line 31.

src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -0,0 +1,131 @@
...
+          <strong :class="{ 'error-text': invalidSectionExpiry }">Extension Time</strong>
+        </v-col>
+        <v-col cols="12" sm="9">
+          <!-- Limited Restoration Radio Panel -->
+          <LimitedRestorationPanel
+            :currentDate="getCurrentDate"
+            :expiryDate="expiry"
+            :key="expiry"
+            :maxNumberOfMonths=36
+            @expiry="setRestorationExpiry(addMonthsToDate(previousNumberOfMonths, $event))"
```
JazzarKarim, line 21: New expiry is going to be:
The expiry of the previously filed limited restoration + the new extension period.

src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -0,0 +1,131 @@
...
+        </v-col>
+      </v-row>
+    </section>
+
+    <!-- Approval Type Section (if applicable) -->
+    <section id="approval-type-section"
+      class="section-container"
+      :class="{ 'invalid-section': invalidSectionApprovalType }"
+    >
+      <template v-if="approvalType == ApprovalTypes.VIA_COURT_ORDER">
```
JazzarKarim, line 33: Only show if the previously filed limited restoration was approved by court order.

src/components/ViewWrapper.vue
```diff
@@ -323,10 +324,12 @@ export default class App extends Vue {
 
     let filingComplete: any
     try {
-      let filing: AlterationFilingIF | ChgRegistrationFilingIF | ConversionFilingIF | SpecialResolutionFilingIF
+      let filing: AlterationFilingIF | ChgRegistrationFilingIF | ConversionFilingIF |
+      RestorationFilingIF | SpecialResolutionFilingIF
       if (this.isAlterationFiling) filing = this.buildAlterationFiling(isDraft)
       if (this.isFirmChangeFiling) filing = this.buildChangeRegFiling(isDraft)
       if (this.isFirmConversionFiling) filing = this.buildConversionFiling(isDraft)
+      if (this.isRestorationFiling) filing = this.buildRestorationFiling(isDraft)
```
JazzarKarim, line 332: Save and resume.

src/mixins/date-mixin.ts
```diff
@@ -142,4 +142,20 @@ export default class DateMixin extends Vue {
...
+   */
+  addMonthsToDate (month: number, date: string): string {
+    return DateUtilities.addMonthsToDate(month, date)
+  }
+
+  /**
+   * DEPRECATED Decrease one date from another and return number of months as the difference.
+   * @example (2023-02-03, 2024-08-03) -> 18
+   */
+  subtractDates (dateFrom: string, dateTo: string): number {
```
JazzarKarim, line 158: The limited restoration shared component panel uses these methods. So, I had to add them here.

package.json
```diff
@@ -26,7 +26,7 @@
     "@bcrs-shared-components/court-order-poa": "2.1.3",
     "@bcrs-shared-components/date-picker": "1.2.5",
     "@bcrs-shared-components/detail-comment": "1.1.1",
-    "@bcrs-shared-components/enums": "1.0.27",
+    "@bcrs-shared-components/enums": "1.0.36",
```
severinbeauvais, line 29: OK, I see the other shared imports are WAY out of date. I'm OK to leave it for now, since it will require some looking at code to see what changed and then testing the changes. Please create a new ticket (Tech Debt) to do this.

src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -0,0 +1,131 @@
...
+      <v-row no-gutters>
+        <v-col cols="12" sm="3">
+          <strong :class="{ 'error-text': invalidSectionExpiry }">Extension Time</strong>
+        </v-col>
+        <v-col cols="12" sm="9">
+          <!-- Limited Restoration Radio Panel -->
+          <LimitedRestorationPanel
+            :currentDate="getCurrentDate"
+            :expiryDate="expiry"
+            :key="expiry"
```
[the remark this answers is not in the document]
severinbeauvais: Is this because the component doesn't update itself when the `expiryDate` prop changes? Normally we'd handle that with a watcher in the component, but that's a bit messy. This, however, feels like a cheat :)

I don't know how long Jonathan is away for but I'd like his opinion on this.

My opinion is this OK until we have a better way.

src/components/Restoration/ExtendTimeLimit.vue
severinbeauvais: 3 years now? Does Create UI have to be updated for the initial limited resto application? (The shared component defaults it to 24.)

severinbeauvais: > New expiry is going to be: The expiry of the previously filed limited restoration + the new extension period.

Did you confirm this? It doesn't sound right to me.

src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /** This extension time's validity state (when prompted by app). */
+  get invalidSectionExpiry (): boolean {
+    return !this.getExpiryValid
+  }
+
+  /** This approval type's validity state (when prompted by app). */
+  get invalidSectionApprovalType (): boolean {
+    return !this.getApprovalTypeValid
```
severinbeauvais, line 110: Why not just use `!getApprovalTypeValid` in the template?

Ditto line 105 above.

src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /** The previously selected number of months of limited restoration. */
+  get previousNumberOfMonths (): string {
+    return this.subtractDates(this.getCurrentDate, this.getStateFilingRestoration.expiry)
```
severinbeauvais, line 115: This will give you a different result as currentDate changes to the next month, so it's not what the comment says it is.

It _might_ be the remaining months to the original expiry date.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -41,12 +41,18 @@
...
 
 export default {
+  data () {
+    return {
+      ApprovalTypes: ApprovalTypes
+    }
+  },
   computed: {
     ...mapGetters(['isLimitedConversionRestorationFiling', 'isLimitedExtendRestorationFiling',
-      'hasFileNumber', 'getFileNumber', 'getFormattedExpiryText'])
```
severinbeauvais, line 49: Looks like `hasFileNumber` is no longer used either. Please remove.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -26,13 +26,13 @@
...
         </v-row>
-        <v-row no-gutters class="mt-3" v-if="hasFileNumber">
+        <v-row no-gutters class="mt-3" v-if="getStateFilingRestoration.approvalType === ApprovalTypes.VIA_COURT_ORDER">
           <v-col cols="3">
             <label><strong>Approval Type</strong></label>
           </v-col>
           <v-col cols="8" class="mt-n1">
             <div class="font-weight-bold">Approved by Court Order</div>
-            <div>Court Order Number: {{ getFileNumber }}</div>
+            <div v-if="getRestoration.courtOrder">Court Order Number: {{ getRestoration.courtOrder.fileNumber }}</div>
```
severinbeauvais, line 35: I'd be tempted to use a couple of local computed methods (getters) to replace the code here and on line 29, just to keep the template super clean.

`isApprovedByCourtOrder` and `courtOrder`, if there aren't already getters for them.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -41,12 +41,18 @@
 </template>
 
 <script lang="ts">
+import { ApprovalTypes } from '@bcrs-shared-components/enums'
 import { mapGetters } from 'vuex'
 
 export default {
+  data () {
+    return {
+      ApprovalTypes: ApprovalTypes
```
severinbeauvais, line 50: Nice. I bet you had to search how to do this 👍

src/enums/index.ts
```diff
@@ -17,5 +17,6 @@ export * from './routeNames'
 
 // external enum
 export { CorpTypeCd } from '@bcrs-shared-components/corp-type-module/'
-export { ActionTypes, EntityStates, FilingCodes, FilingNames, FilingTypes, RestorationTypes }
+export { ActionTypes, ApprovalTypes, EntityStates, FilingCodes, FilingNames, FilingTypes, RelationshipTypes,
```
severinbeauvais, line 20: Looks like RelationshipTypes isn't used in here.

src/mixins/date-mixin.ts
severinbeauvais: That's fine. We'll sort this out in a refactor ticket, maybe 15660.

src/mixins/filing-template-mixin.ts
```diff
@@ -776,9 +778,9 @@ export default class FilingTemplateMixin extends DateMixin {
    * @param filing the restoration filing
    * @param entitySnapshot the latest entity snapshot
    */
-  parseRestorationFiling (filing: RestorationFilingIF, entitySnapshot: EntitySnapshotIF): void {
+  parseRestorationFiling (filing: RestorationFilingIF): void {
     // store Entity Snapshot
-    this.setEntitySnapshot(entitySnapshot)
+    const entitySnapshot = this.getEntitySnapshot
```
severinbeauvais, line 783: This doesn't do what the comment says it does. What is the purpose of this code? Is it obsolete now due to some other changes?

src/mixins/filing-template-mixin.ts
```diff
@@ -791,8 +793,17 @@ export default class FilingTemplateMixin extends DateMixin {
...
+    if (filing.restoration.courtOrder) {
+      this.setRestorationCourtOrder(filing.restoration.courtOrder)
+    }
     this.setRestorationType(filing.restoration.type)
-    this.setRestorationExpiry(filing.restoration.expiry || null)
+    if (filing.restoration.expiry) {
+      this.setRestorationExpiry(filing.restoration.expiry)
+    } else {
+      // Reset radio button to 2 years
+      this.setRestorationExpiry(DateUtilities.addMonthsToDate(36, this.getStateFilingRestoration.expiry))
```
severinbeauvais, line 805: Please fix value (36) or comment (2 years).

src/services/date-utilities.ts
```diff
@@ -254,4 +254,60 @@ export default class DateUtilities {
...
+    if (!dateTo) {
+      dateTo = this.addMonthsToDate(24, dateFrom)
+    }
+    const expiryDate = this.yyyyMmDdToDate(dateTo)
+    const currDate = this.yyyyMmDdToDate(dateFrom)
+    const monthDiff = expiryDate.getMonth() - currDate.getMonth()
+    const yearDiff = (12 * (expiryDate.getFullYear() - currDate.getFullYear()))
+    let difference = monthDiff + yearDiff
+    // For example if Jan 31st and March 1, difference is 1 not 2
+    if ((expiryDate.getDate() + 25) < currDate.getDate()) {
```
severinbeauvais, line 308: Why add 25?

src/store/getters.ts
```diff
@@ -1085,7 +1106,8 @@ export default {
 
   getFormattedExpiryText: (state, getters) => (today = new Date()): string => {
     if (getters.getExpiryDateString) {
-      return DateUtilities.monthDiffToToday(getters.getExpiryDateString, today) + ' months, expires on ' +
+      return DateUtilities.subtractDates(getters.getStateFilingRestoration.expiry,
+        getters.getExpiryDateString) + ' months, expires on ' +
```
severinbeauvais, line 1110: Please break this up into variables and text.

src/store/actions.ts
```diff
@@ -299,5 +300,28 @@ export default {
...
+
+  setRestorationApprovalType ({ commit }, type: ApprovalTypes) {
+    commit('mutateRestorationApprovalType', type)
+  },
+
+  setStateFilingRestoration (context): void {
+    LegalServices.fetchFiling(context.getters.getStateFilingUrl)
+      .then((response) => {
+        context.commit('mutateStateFilingRestoration', response.restoration)
+      })
```
severinbeauvais, line 313: This should a promise that is resolved on success and rejected on failure.

Example: https://github.com/bcgov/business-filings-ui/blob/f56ca3a2358bf3da406973e1fe74fd2225e20db8/src/store/Business/actions.ts#L8

Also, you should await the call to this action since it's async.

Also, the return type should be `Promise<any>`.

src/components/Restoration/ExtendTimeLimit.vue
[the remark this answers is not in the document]
JazzarKarim: Yes, exactly! This way whenever the prop data changes (getters, setters etc), the component will update.

CC: @jonathan-longe

src/components/Restoration/ExtendTimeLimit.vue
JazzarKarim: > 3 years now? Does Create UI have to be updated for the initial limited resto application? (The shared component defaults it to 24.)

According to UXPin:
uxpin extension
For Create UI: It's 24 months.
For Edit UI: it's 36 months.

JazzarKarim: > > New expiry is going to be: The expiry of the previously filed limited restoration + the new extension period.
> 
> Did you confirm this? It doesn't sound right to me.

I spoke with Ethan about it. He told me it's business logic so I should ask Mihai.
I asked Mihai and he confirmed that this is the way they want to have it!

package.json
JazzarKarim: Deal, will do.

src/components/Restoration/ExtendTimeLimit.vue
severinbeauvais: OK. Thanks for checking. I'm OK with this.

src/components/Restoration/YourCompanySummary.vue
JazzarKarim: Indeed I had to Sev! 😸

src/mixins/filing-template-mixin.ts
JazzarKarim: I'll be fixing the comment! We're now getting the entity snapshot here not storing it. It's being stored now in `limitedRestorationExtension.vue` 
For exact line:
https://github.com/JazzarKarim/business-edit-ui/blob/fc30d04a839eb4cfda728a136c1df4ca0d4f6cb3/src/views/LimitedRestorationExtension.vue#L240

src/components/Restoration/YourCompanySummary.vue
severinbeauvais: By the way, when you see `x: x` in an object, you can simplify it to just `x`.

src/services/date-utilities.ts
JazzarKarim: This is like a cut-off value. Since some months have more days than others, the difference between them might be incorrect if I don't add this 25 value.

For example, take a look at this scenario:
Current date: Jan 31st
Expiry Date: March 1st.
If I do `((expiryDate.getDate()) < currDate.getDate())`, it's going to be 1 < 31, we return 1 which is correct.
With the 25:
26 < 31, we return 1 which is correct.
In this case, it doesn't matter. 

Now take a look at this scenario:
Current date: Jan 16th
Expiry Date: March 15th.
If I do `((expiryDate.getDate()) < currDate.getDate())`, it's going to be 15 < 16, we return 1 which is incorrect!
Now with the 25:
40 < 16, we return 2 since we didn't go and decrement the difference. We return in this case the correct value!

This was done in Create UI when I was doing the limited restoration creation. I've tested this extensively. Let me know if that clears things up!

src/components/Restoration/YourCompanySummary.vue
JazzarKarim: Ohh, Cool! Didn't know that. Thanks Sev!

src/services/date-utilities.ts
severinbeauvais: Please add a comment explaining why you use 25.

src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  /** The expiry draft date for extension. */
+  get expiry (): string {
+    return DateUtilities.subtractMonthsToDate(this.previousNumberOfMonths, this.getRestoration.expiry)
+  }
+
+  /** The remaining number of months left for the previously filed limited restoration. */
+  get previousNumberOfMonths (): string {
+    return this.subtractDates(this.getCurrentDate, this.getStateFilingRestoration?.expiry)
```
severinbeauvais, line 105: `subtractDates` returns a number, so this getter should return a number (line 104), or you should convert the number to a string.

JS is really lax with this, and apparently TS typings didn't catch this either (depends on whether it knows the signature of subtractDates). Anyways, best to be correct even if this works :)

src/store/state/state-model.ts
```diff
@@ -144,5 +146,26 @@ export const stateModel: StateModelIF = {
...
+    businessNameValid: false,
+    courtOrder: {
+      fileNumber: null
+    },
+    expiry: null,
+    noticeDate: null,
+    relationships: [],
+    restorationTypeValid: false,
+    type: null
+  }
```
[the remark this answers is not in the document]
severinbeauvais: I'm OK with the way you have it now.

src/store/actions.ts
severinbeauvais: Looks good 👍

severinbeauvais: I see that `setStateFilingRestoration` is awaited in LimitedRestorationExtension.vue. Is the same code in the restoration conversion view?

severinbeauvais:
Have another look at the unit tests.

Do they run for you locally? `npm run test:unit`

src/store/actions.ts
JazzarKarim: No it's not there. I didn't want to touch that file since it might be someone else's ticket. I can add it however. Should I do that?

severinbeauvais: Yeah, we're working on both filings at once, so if you add something to one then you should update the other one too (if applicable, but I think it is here).

JazzarKarim: It is applicable. Okay, noted. Will do that Sev.

JazzarKarim:
> Have another look at the unit tests.
> 
> Do they run for you locally? `npm run test:unit`
> 
> PS I usually redirect the results of the above to a desktop file that I can more easily navigate/search, ie. `npm run test:unit > /mnt/c/users/severin/Desktop/tests.log 2>&1`

14 Tests failed. Most of them are in the `LimitedRestorationExtension.spec.ts` and `LimitedRestorationToFull.spec.ts` files. I can take a look and try to fix them. Shall I do that in this ticket?
Also, Noted. That's very helpful. Thanks Sev.

severinbeauvais:
> 14 Tests failed. Most of them are in the `LimitedRestorationExtension.spec.ts` and `LimitedRestorationToFull.spec.ts` files. I can take a look and try to fix them. Shall I do that in this ticket? Also, Noted. That's very helpful. Thanks Sev.

As a general rule, if they were working before your changes then you should fix them. Hopefully it's something simple, like missing init.

JazzarKarim:
> > 14 Tests failed. Most of them are in the `LimitedRestorationExtension.spec.ts` and `LimitedRestorationToFull.spec.ts` files. I can take a look and try to fix them. Shall I do that in this ticket? Also, Noted. That's very helpful. Thanks Sev.
> 
> As a general rule, if they were working before your changes then you should fix them. Hopefully it's something simple, like missing init.

I think they were already not working even before my changes. I'll have a look regardless. I think like you said, it's a simple fix.

src/store/actions.ts
severinbeauvais: Looks good.

severinbeauvais (commented):
LGTM but please get another review or 2.

src/store/actions.ts
```diff
@@ -299,5 +300,37 @@ export default {
 
   setRestorationExpiry ({ commit }, expiry: string) {
     commit('mutateRestorationExpiry', expiry)
+  },
+
+  setRestorationApprovalType ({ commit }, type: ApprovalTypes) {
+    commit('mutateRestorationApprovalType', type)
+  },
```
jonathan-longe, line 307: Would it be possible to remove this action method and call the mutation directly?  I know you're following the pattern set previously, but there's no reason to call an action when calling mutation will suffice.   Actions are typically for async operations and anything that includes business logic.  Mutations should be used when you're simply changing state -- instantaneous operations.

JazzarKarim: So you mean in my code instead of calling `setRestorationApprovalType`, I would import and call `mutateRestorationApprovalType`? I can then remove all these new setters?

src/store/actions.ts
```diff
@@ -299,5 +300,37 @@ export default {
...
+  setRestorationCourtOrder ({ commit }, courtOrder: CourtOrderIF): void {
+    commit('mutateRestorationCourtOrder', courtOrder)
+  },
+
+  setExpiryValid ({ commit }, valid: boolean): void {
+    commit('mutateExpiryValid', valid)
+  },
+
+  setApprovalTypeValid ({ commit }, valid: boolean): void {
+    commit('mutateApprovalTypeValid', valid)
```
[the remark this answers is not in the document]
severinbeauvais: It doesn't look like [vue-property-decorator](https://www.npmjs.com/package/vue-property-decorator) has a `@Mutation` decorator. What's a clean way to declare an imported mutation?

src/store/mutations.ts
```diff
@@ -288,6 +288,25 @@ export default {
...
+    state.stateModel.restoration.courtOrder = courtOrder
+  },
+
+  mutateApprovalTypeValid (state: StateIF, valid: boolean) {
+    state.stateModel.validationFlags.flagsCompanyInfo.isValidApprovalType = valid
+  },
+
+  mutateExpiryValid (state: StateIF, valid: boolean) {
+    state.stateModel.validationFlags.flagsCompanyInfo.isValidExtensionTime = valid
+  }
```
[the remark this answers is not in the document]
severinbeauvais: From what I've read, that's one of the things Pinia does better, fwiw.

Typings are important to me so I don't want to lose them. How about writing an update action that takes type (`<T>`) ?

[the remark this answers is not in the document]
JazzarKarim: Was going to ask the same question as Sev. For example, if we're trying to update the courtOrder which is a type of CourtOrderIF, how would we do that?

src/components/Restoration/ExtendTimeLimit.vue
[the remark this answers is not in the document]
jonathan-longe: This appears to be an issue for all instances of the `<LimitedRestorationPanel>`.  As such, I'd suggest we should create a ticket to resolve it rather than handling it in the parent.  Please push back if you disagree.

src/store/actions.ts
jonathan-longe: Yes, correct

src/store/mutations.ts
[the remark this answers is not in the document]
jonathan-longe: > How about writing an update action that takes type (<T>) ?

Take a type as a parameter?  What do we use to verify that the type is correct?  From what I understand, we can't verify types at runtime -- right?  If I'm thinking about this correctly, a type parameter would need to verified at runtime.

> if we're trying to update the courtOrder which is a type of CourtOrderIF, how would we do that?

I see two options, either we create a `updateCourtOrder()` that checks the type, or we use a generic `updateAttribute()` that doesn't.  Either way, I think it's better than creating a getter and setter for each store value.

src/store/actions.ts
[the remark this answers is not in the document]
jonathan-longe: I hear you @severinbeauvais . I think the best solution is to start using the Composition API which not only has better support for typing but also helps in our transition to Vue3.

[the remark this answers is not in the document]
severinbeauvais: Yes, we could start using the Composition API for new things. It is supported in Vue 2.7.10, which we are using.

If we can solve the mutation-calling issue (2 comments in this PR) in a day or less, let's do it. If not, ignore it for this component and we'll tackle it in another component, or globally in the future.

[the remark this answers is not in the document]
JazzarKarim: Shall I try to fix that now in this PR? I don't think I'll be able to complete that by the end of the day however since this is uncharted territory for me.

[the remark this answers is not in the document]
severinbeauvais: It's looking like "ignore it for now". @jonathan-longe OK with this?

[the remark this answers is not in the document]
jonathan-longe: @JazzarKarim  I think we can use `@Mutation` like this:

```javascript
import { Action, Getter, Mutation } from 'vuex-class'

export class SomeComponent extends Vue {
@Mutation('foo') mutationFoo

// blah, blah, blah
}
```
See:  https://github.com/ktsn/vuex-class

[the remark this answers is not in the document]
JazzarKarim: Alright then, on it!

src/store/actions.ts
severinbeauvais: Looks great! I'm OK with this conversation (but it's not mine to resolve).

src/store/actions.ts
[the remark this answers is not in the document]
severinbeauvais: Looks great! I'm OK with this conversation (but it's not mine to resolve).

src/store/mutations.ts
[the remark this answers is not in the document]
severinbeauvais: Alas, mutation decorators, like action decorators, don't have typing of the value being set. Oh well 🤷‍♂️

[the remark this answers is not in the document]
severinbeauvais: Or can they? Does this work?
```
@Mutation mutateExpiryValid!: boolean
```

jonathan-longe (approved):
Good work @JazzarKarim thanks for making the requested changes

package.json
severinbeauvais: What's the new ticket's number?

severinbeauvais (approved):

JazzarKarim:
> Good work @JazzarKarim thanks for making the requested changes

Thank you for your awesome feedback 😺

package.json
JazzarKarim: https://github.com/bcgov/entity/issues/15764

PR #461, opened 2023-03-24
15499 Firebase deployment and convert configmap and keycloak to .env

pwei1018:
*Issue #:* [15499](https://github.com/bcgov/entity/issues/15499)

*Description of changes:*
1. Enable firebase deployment;
2. Convert configuration.json to .env;
3. Convert keycloak.json to .env;
4. Remove PaymentSystemAlert and replace with banner-text;


.env
```diff
@@ -1 +0,0 @@
-VUE_APP_PATH = businesses/edit
```
severinbeauvais, line 1: Should this file be deleted?

severinbeauvais: Never mind! GH showed me the old contents so I thought the file still existed.

pwei1018: Yes, it will generate during deployment. Or, developer create it own version. Don't need to put into the repo.

src/utils/FetchConfig.ts
```diff
@@ -19,103 +19,67 @@ export async function FetchConfig (): Promise<any> {
...
-  }
-
-  const response = await axios.get(url, { headers }).catch(() => {
-    return Promise.reject(new Error('Could not fetch configuration.json'))
-  })
-
-  const authWebUrl: string = response.data['AUTH_WEB_URL']
+  const authWebUrl: string = process.env.VUE_APP_AUTH_WEB_URL
   sessionStorage.setItem('AUTH_WEB_URL', authWebUrl)
-  console.info('Set Auth Web URL to: ' + authWebUrl)
```
severinbeauvais, line 38: Thanks for deleting the unneeded console statements! 👍

tests/unit/FetchConfig.spec.ts
```diff
@@ -1,48 +1,41 @@
...
-          SITEMINDER_LOGOUT_URL: 'siteminder logout url',
-          STATUS_API_URL: 'status api url',
-          STATUS_API_VERSION: '/status api version'
-        }
-      }))
   })
 
   afterEach(() => {
-    sinon.restore()
   })
```
severinbeauvais, line 39: Since this method is empty, please delete it.

severinbeauvais (commented):
Very nice cleanup!

This looks substantially like the Edit UI PR for this feature, so if that one is correct then I think this one is correct also.

severinbeauvais (approved):

tests/unit/FetchConfig.spec.ts
severinbeauvais, line 39: Since this method is empty, please delete it.

PR #464, opened 2023-03-29
14974 Updated extend time limit section to respond to UXA comments

JazzarKarim:
*Issue #:* /bcgov/entity#14974

*Description of changes:*
- As per Ethan's comments, this PR fixes the style and spacing for the Approval type and the limited restoration radio panel.
extensions ux

- Ethan's comments: https://github.com/bcgov/entity/issues/14974#issuecomment-1489055310

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

### DIFF
--- a/src/components/Restoration/ExtendTimeLimit.vue
+++ b/src/components/Restoration/ExtendTimeLimit.vue
@@ -8,10 +8,10 @@
       :class="{ 'invalid-section': !getExpiryValid }"
     >
       <v-row no-gutters>
-        <v-col cols="12" sm="3">
+        <v-col cols="12" sm="2" class="pt-2">
           <strong :class="{ 'error-text': !getExpiryValid }">Extension Time</strong>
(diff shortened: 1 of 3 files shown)


src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -29,8 +29,9 @@
     <section id="approval-type-section"
       class="section-container"
       :class="{ 'invalid-section': !getApprovalTypeValid }"
+      v-if="approvalType == ApprovalTypes.VIA_COURT_ORDER"
```
JazzarKarim, line 32: Section was taking space although approval type was approved via registrar. Having the v-if here instead of in the template would get rid of that.

ketaki-deodhar (approved):

PR #465, opened 2023-03-31
Fixed the content path issue.

pwei1018:
*Issue #:* [15499](https://github.com/bcgov/entity/issues/15499)

*Description of changes:*
The current VUE_APP_PATH of GCP CD setting cause a null value exception after deploy to firebase. It must be changed to "/" value to avoid exception.

Also, open a ticket to investigate BASE_URL issue. https://github.com/bcgov/entity/issues/15817

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
-# Base Path Openshift: businesses/edit Firebase:
-VUE_APP_PATH=businesses/edit
+# Base Path Openshift: /businesses/edit Firebase: /
+VUE_APP_PATH="/businesses/edit"
 
 #vaults Shared
 VUE_APP_ADDRESS_COMPLETE_KEY=
--- a/devops/vaults.env
+++ b/devops/vaults.env
@@ -1,5 +1,5 @@
-# Base Path Openshift: businesses/edit Firebase:
-VUE_APP_PATH=
+# Base Path Openshift: /businesses/edit Firebase: /
+VUE_APP_PATH="/"
 
 #vaults Shared
 VUE_APP_ADDRESS_COMPLETE_KEY="op://canadapost/$APP_ENV/address-key/ADDRESS_COMPLETE_KEY"
--- a/src/utils/FetchConfig.ts
+++ b/src/utils/FetchConfig.ts
@@ -1,5 +1,3 @@
-/* eslint-disable no-console */
-
 import { AxiosInstance as axios } from '@/utils/'
 
 /**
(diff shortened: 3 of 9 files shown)


devops/vaults.env
```diff
@@ -1,5 +1,5 @@
-# Base Path Openshift: businesses/edit Firebase:
-VUE_APP_PATH=
+# Base Path Openshift: /businesses/edit Firebase: /
+VUE_APP_PATH="/"
```
severinbeauvais, line 2: Did you run this app locally to test this?

src/utils/FetchConfig.ts
```diff
@@ -96,11 +94,9 @@ export async function FetchConfig (): Promise<any> {
   // eg, "/basePath/BCxxx/"
   const vueRouterBase = processEnvBaseUrl + id + '/'
   sessionStorage.setItem('VUE_ROUTER_BASE', vueRouterBase)
-  console.info('Set Vue Router Base to: ' + vueRouterBase)
```
severinbeauvais, line 99: Thanks for this cleanup!

pwei1018:
> LGTM. Did you run this locally to verify that a filing (eg, correction) loaded up, and could be saved and resumed, fine?

I tried incorporation filing and seems fine. No sure other filings, do you have account that can let try it?

devops/vaults.env
pwei1018: Yes, test local and in firebase.

severinbeauvais:
> > LGTM. Did you run this locally to verify that a filing (eg, correction) loaded up, and could be saved and resumed, fine?
> 
> I tried incorporation filing and seems fine. No sure other filings, do you have account that can let try it?

It should not matter what account you used. If an IA filing worked then the paths obviously work. This is fine.

devops/vaults.env
severinbeauvais: Excellent - thanks!

.env.example
```diff
@@ -1,5 +1,5 @@
-# Base Path Openshift: businesses/edit Firebase:
-VUE_APP_PATH=businesses/edit
+# Base Path Openshift: /businesses/edit Firebase: /
+VUE_APP_PATH="/businesses/edit"
```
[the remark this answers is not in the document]
severinbeauvais: Actually I'm asking about the route for each UI -- what are they? I noticed differences in the VUE_APP_PATH -- why can't they all be "/business" when running locally?

PR #466, opened 2023-03-31
15080-Edit UI - Extension and Conversion - Summary page and validation

ketaki-deodhar:
*Issue #:* /bcgov/entity#15080

*Description of changes:*
This is (WIP). I will have 2 PRs for this. Second PR will include unit test fixes
- [x] verify/update Applicant Information block
- [x] verify/update Your Company block
   - see right side of uxpin for display if something has been changed
- [x] verify/update Document Delivery section
- [x] verify/update Certify section
- [x] verify/update Staff Payment section
- [x] verify/update validation of components
- [x] verify/update save and resume later
- [x] clean up unused fields (resources, interfaces, views)

Fixed bugs:
- [x] Applicant was not populated correctly (this was working earlier, broke because of some other code changes)
- [x] Could not start a Restoration to Full application. Fixed by setting entitySnapshot in state

### DIFF
--- a/src/components/Restoration/RestorationSummary.vue
+++ b/src/components/Restoration/RestorationSummary.vue
@@ -58,50 +58,14 @@ export default class RestorationSummary extends Mixins(
 
     return `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.`
   }
-
-  /** Calculates the sum of restoration fees. */
-  get restorationFees (): string {
-    const validFees = this.getCurrentFees.filter(f => f.filingFees !== null && f.futureEffectiveFees !== null)
-    if (validFees.length === 0) {
-      return ''
-    }
-    /** Calculates the sum of filing fees. */
-    const filingFeesSum = validFees.map(f => f.filingFees).reduce((a, b) => a + b, 0)
-    /** Calculates the sum of future effective fees. */
-    const futureEffectiveFeesSum = validFees.map(f => f.futureEffectiveFees).reduce((a, b) => a + b, 0)
-    return `($${(filingFeesSum + futureEffectiveFeesSum).toFixed(2)} Fee)`
-  }
-
-  protected onDeleteClicked (): void {
-    this.$root.$emit('delete-all')
-  }
 }
 </script>
 
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
 
-.summary-header {
-  display: flex;
-  background-color: $BCgovBlue5O;
-  padding: 1.25rem;
-}
-
-.summary-title {
-  padding-left: 0.5rem;
-}
-
 .company-name {
   font-size: 1.5rem;
-}
-
-.actions {
-  position: absolute;
-  right: 0;
-
-  #btn-delete-restoration {
-    margin-top: -6px;
-  }
 }
 
 .inner-col-1 {
--- a/src/components/Restoration/RestorationSummary.vue
+++ b/src/components/Restoration/RestorationSummary.vue
@@ -1,30 +1,5 @@
 <template>
   <v-card flat id="restoration-summary">
-    <!-- Section Header -->
-    <div class="summary-header px-4 mb-2 rounded-t">
-      <v-row no-gutters>
-        <v-col cols="9">
-          <img class="header-icon" src="@/assets/images/currency-usd-circle.svg">
-          <label class="summary-title">Restoration Notice Changes {{restorationFees}}</label>
-        </v-col>
-
-        <!-- Actions -->
-        <v-col cols="3">
-          <div class="actions mr-4">
-            <v-btn
-              text color="primary"
-              id="btn-delete-restoration"
-              :disabled="isBusySaving"
-              @click="onDeleteClicked()"
-            >
-              <v-icon small>mdi-delete</v-icon>
-              <span>Delete</span>
-            </v-btn>
-          </div>
-        </v-col>
-      </v-row>
-    </div>
-
     <!-- Business Name -->
     <template v-if="hasBusinessNameChanged">
       <v-divider class="mx-4" />
--- a/src/components/Restoration/YourCompanySummary.vue
+++ b/src/components/Restoration/YourCompanySummary.vue
@@ -4,12 +4,12 @@
     <div class="summary-header px-4 mb-2 rounded-t">
       <v-row no-gutters>
         <v-col cols="9">
-          <img class="header-icon" src="@/assets/images/currency-usd-circle.svg">
+          <v-icon class="header-icon">mdi-domain</v-icon>
           <label class="summary-title pl-2">Your Company</label>
         </v-col>
       </v-row>
     </div>
-    <!-- Business Name -->
+
     <template>
       <v-divider class="mx-4" />
       <div class="section-container business-name-summary">
@@ -36,6 +36,9 @@
           </v-col>
         </v-row>
       </div>
+      <div class="section-container">
+        <OfficeAddresses :isSummaryView="true" />
+      </div>
     </template>
   </v-card>
 </template>
@@ -43,6 +46,7 @@
 <script lang="ts">
 import { ApprovalTypes } from '@bcrs-shared-components/enums'
 import { mapGetters } from 'vuex'
+import { OfficeAddresses } from '@/components/common/'
 
 export default {
   data () {
@@ -62,6 +66,9 @@ export default {
     getCourtOrderFileNumber () {
       return this.getRestoration.courtOrder.fileNumber
     }
+  },
+  components: {
+    OfficeAddresses
   }
 }
 </script>
--- a/src/components/Restoration/YourCompanySummary.vue
+++ b/src/components/Restoration/YourCompanySummary.vue
@@ -98,7 +98,6 @@ export default {
       return this.getRestoration.courtOrder
     },
     getCourtOrderFileNumber () {
-      console.log('Court Order', this.getRestoration.courtOrder.fileNumber)
       return this.getRestoration.courtOrder.fileNumber
     },
     getCompanyName (): string {
--- a/src/components/Restoration/YourCompanySummary.vue
+++ b/src/components/Restoration/YourCompanySummary.vue
@@ -12,7 +12,31 @@
 
     <template>
       <v-divider class="mx-4" />
-      <div class="section-container business-name-summary">
+
+      <!-- Business Name -->
+      <template v-if="hasBusinessNameChanged">
+        <div class="section-container business-name-summary">
+          <v-row no-gutters>
+            <v-col cols="3">
+              <label><strong>Company Name</strong></label>
+            </v-col>
+
+            <v-col cols="8" class="mt-n1">
+              <div class="company-name font-weight-bold text-uppercase">{{ getCompanyName }}</div>
+              <div class="company-name mt-2">{{ getNameRequest.nrNumber }}</div>
+            </v-col>
+          </v-row>
+        </div>
+      </template>
+
+      <!-- Name Translation -->
+      <template v-if="haveNameTranslationsChanged">
+        <div class="section-container name-translation-summary">
+          <NameTranslation :isSummaryMode="true" />
+        </div>
+      </template>
+
+      <div class="section-container">
         <v-row no-gutters>
           <v-col cols="3">
             <label><strong>Restoration Type</strong></label>
@@ -46,7 +70,8 @@
 <script lang="ts">
 import { ApprovalTypes } from '@bcrs-shared-components/enums'
 import { mapGetters } from 'vuex'
-import { OfficeAddresses } from '@/components/common/'
+import { OfficeAddresses, NameTranslation } from '@/components/common/'
+import { CommonMixin, DateMixin, FilingTemplateMixin } from '@/mixins/'
 
 export default {
   data () {
@@ -55,8 +80,17 @@ export default {
     }
   },
   computed: {
-    ...mapGetters(['isLimitedConversionRestorationFiling', 'isLimitedExtendRestorationFiling',
-      'getRestoration', 'getFormattedExpiryText', 'getStateFilingRestoration']),
+    ...mapGetters([
+      'getBusinessNumber',
+      'getFormattedExpiryText',
+      'getNameRequestLegalName',
+      'getNameRequest',
+      'getRestoration',
+      'getStateFilingRestoration',
+      'hasBusinessNameChanged',
+      'haveNameTranslationsChanged',
+      'isLimitedConversionRestorationFiling',
+      'isLimitedExtendRestorationFiling']),
     getStateFilingApprovalType () {
       return this.getStateFilingRestoration?.approvalType
     },
@@ -64,11 +98,17 @@ export default {
       return this.getRestoration.courtOrder
     },
     getCourtOrderFileNumber () {
+      console.log('Court Order', this.getRestoration.courtOrder.fileNumber)
       return this.getRestoration.courtOrder.fileNumber
+    },
+    getCompanyName (): string {
+      if (this.getNameRequestLegalName) return this.getNameRequestLegalName
+      return `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.`
     }
   },
   components: {
-    OfficeAddresses
+    OfficeAddresses,
+    NameTranslation
   }
 }
 </script>
--- a/src/components/Restoration/YourCompanySummary.vue
+++ b/src/components/Restoration/YourCompanySummary.vue
@@ -47,7 +47,7 @@
           </v-col>
           <v-col cols="8" class="mt-n1" v-if="isLimitedConversionRestorationFiling">
             <div class="font-weight-bold">Conversion to Full Restoration</div>
-            <div>[TODO - Applicant's relationship: Director, Shareholder]</div>
+            <div>Applicant's relationship: {{ getRelationships }}</div>
           </v-col>
         </v-row>
         <v-row no-gutters class="mt-3" v-if="getStateFilingApprovalType === ApprovalTypes.VIA_COURT_ORDER">
@@ -104,6 +104,9 @@ export default {
     getCompanyName (): string {
       if (this.getNameRequestLegalName) return this.getNameRequestLegalName
       return `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.`
+    },
+    getRelationships () {
+      return this.getRestoration.relationships.join(', ') || '[Unknown]'
     }
   },
   components: {
--- a/src/components/common/CertifySection.vue
+++ b/src/components/common/CertifySection.vue
@@ -71,7 +71,7 @@ export default class CertifySection extends Mixins(DateMixin) {
 
   /** Get the entity type in readable format */
   get readableEntityType (): string {
-    return this.getResource?.certifyText ? this.getResource?.certifyText : GetCorpFullDescription(this.getEntityType)
+    return this.getResource?.certifyText ? this.getResource.certifyText : GetCorpFullDescription(this.getEntityType)
   }
 
   /** Get the certify resource message */
--- a/src/components/common/CertifySection.vue
+++ b/src/components/common/CertifySection.vue
@@ -71,7 +71,7 @@ export default class CertifySection extends Mixins(DateMixin) {
 
   /** Get the entity type in readable format */
   get readableEntityType (): string {
-    return this.getResource?.certifyText ? this.getResource.certifyText : GetCorpFullDescription(this.getEntityType)
+    return (this.getResource?.certifyText || GetCorpFullDescription(this.getEntityType))
   }
 
   /** Get the certify resource message */
--- a/src/components/common/CertifySection.vue
+++ b/src/components/common/CertifySection.vue
@@ -71,7 +71,7 @@ export default class CertifySection extends Mixins(DateMixin) {
 
   /** Get the entity type in readable format */
   get readableEntityType (): string {
-    return GetCorpFullDescription(this.getEntityType)
+    return this.getResource?.certifyText ? this.getResource?.certifyText : GetCorpFullDescription(this.getEntityType)
   }
 
   /** Get the certify resource message */
--- a/src/components/common/DocumentsDelivery.vue
+++ b/src/components/common/DocumentsDelivery.vue
@@ -18,7 +18,7 @@
           </v-col>
         </v-row>
 
-        <v-row v-if="isRoleStaff" no-gutters class="mt-6">
+        <v-row v-if="isRoleStaff && userEmailOptional" no-gutters class="mt-6">
           <v-col cols="3" class="px-0">
             <label :class="{ 'error-text': documentDeliveryInvalid }"><strong>Completing Party</strong></label>
           </v-col>
@@ -79,6 +79,8 @@ export default class DocumentsDelivery extends Mixins(CommonMixin) {
 
   /** Whether to perform validation. */
   @Prop({ default: false }) readonly validate!: boolean
+
+  @Prop({ default: true }) readonly userEmailOptional!: boolean
 
   // Local properties
   private optionalEmail = ''
--- a/src/components/common/DocumentsDelivery.vue
+++ b/src/components/common/DocumentsDelivery.vue
@@ -20,7 +20,7 @@
 
         <v-row v-if="isRoleStaff && userEmailOptional" no-gutters class="mt-6">
           <v-col cols="3" class="px-0">
-            <label :class="{ 'error-text': documentDeliveryInvalid }"><strong>Completing Party</strong></label>
+            <label :class="{ 'error-text': documentDeliveryInvalid }"><strong>{{ emailLabel }}</strong></label>
           </v-col>
           <v-col cols="9" class="px-0">
             <v-text-field
@@ -39,10 +39,10 @@
 
         <v-row v-else no-gutters class="mt-6">
           <v-col cols="3" class="px-0">
-            <label><strong>Completing Party</strong></label>
+            <label><strong>{{ emailLabel }}</strong></label>
           </v-col>
           <v-col cols="9" class="px-0">
-            <span class="info-text">{{ getUserEmail }}</span>
+            <span class="info-text">{{ userEmail }}</span>
           </v-col>
         </v-row>
       </v-card>
@@ -55,7 +55,7 @@ import { Component, Mixins, Emit, Watch, Prop } from 'vue-property-decorator'
 import { Action, Getter } from 'vuex-class'
 import { CommonMixin } from '@/mixins/'
 import { FilingNames } from '@/enums/'
-import { ActionBindingIF, FlagsReviewCertifyIF } from '@/interfaces/'
+import { ActionBindingIF, FlagsReviewCertifyIF, ResourceIF } from '@/interfaces/'
 import { ContactPointIF } from '@bcrs-shared-components/interfaces/'
 
 // FUTURE: update this component so it doesn't set changes flag initially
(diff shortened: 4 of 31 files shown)


src/components/Restoration/RestorationSummary.vue
```diff
@@ -1,30 +1,5 @@
...
-            >
-              <v-icon small>mdi-delete</v-icon>
-              <span>Delete</span>
-            </v-btn>
-          </div>
-        </v-col>
-      </v-row>
-    </div>
-
     <!-- Business Name -->
```
ketaki-deodhar, line 3: Removed as do not need this for Restoration

src/components/common/DocumentsDelivery.vue
```diff
@@ -109,6 +113,14 @@ export default class DocumentsDelivery extends Mixins(CommonMixin) {
     return (this.validate && !this.getFlagsReviewCertify.isValidDocumentOptionalEmail)
   }
 
+  get userEmail (): string {
+    return this.applicantEmail ? this.applicantEmail : this.getUserEmail
+  }
+
+  get emailLabel (): string {
+    return this.getResource.userEmailLabel ? this.getResource.userEmailLabel : 'Completing Party'
+  }
```
ketaki-deodhar, line 122: compute email label. It is different for other filings and Restorations

src/components/common/DocumentsDelivery.vue
```diff
@@ -109,6 +113,14 @@ export default class DocumentsDelivery extends Mixins(CommonMixin) {
     return (this.validate && !this.getFlagsReviewCertify.isValidDocumentOptionalEmail)
   }
 
+  get userEmail (): string {
+    return this.applicantEmail ? this.applicantEmail : this.getUserEmail
+  }
```
ketaki-deodhar, line 118: compute email value. It is different for other filings(user email) and Restorations (applicant email)

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -4,15 +4,39 @@
...
+
+      <!-- Name Translation -->
+      <template v-if="haveNameTranslationsChanged">
+        <div class="section-container name-translation-summary">
+          <NameTranslation :isSummaryMode="true" />
+        </div>
+      </template>
+
+      <div class="section-container">
         <v-row no-gutters>
```
ketaki-deodhar, line 40: Added Business name and Name translation as they are shown in one block as per UXpin

src/components/common/PeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -412,7 +412,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, OrgPersonMix
   @Prop({ default: true }) readonly showDeliveryAddressColumn!: boolean
   @Prop({ default: true }) readonly showRolesColumn!: boolean
   @Prop({ default: false }) readonly showEmailColumn!: boolean
-  @Prop({ default: true }) readonly showEmailUnderName!: string
+  @Prop({ default: true }) readonly showEmailUnderName!: boolean
```
ketaki-deodhar, line 416: Not part of this but fixed to boolean as it was wrong type

tests/unit/LimitedRestorationExtension.spec.ts
```diff
@@ -282,7 +282,7 @@ describe('Restoration component - edit page', () => {
   })
 })
 
-describe('Restoration component - summary page (with filing changes)', () => {
+xdescribe('Restoration component - summary page (with filing changes)', () => {
   const { assign } = window.location
```
ketaki-deodhar, line 286: Will fix this in next PR

tests/unit/LimitedRestorationToFull.spec.ts
```diff
@@ -282,7 +282,7 @@ describe('LimitedRestorationToFull component - edit page', () => {
   })
 })
 
-describe('LimitedRestorationToFull component - summary page (with filing changes)', () => {
+xdescribe('LimitedRestorationToFull component - summary page (with filing changes)', () => {
   const { assign } = window.location
```
ketaki-deodhar, line 286: Will fix this in next PR

src/views/LimitedRestorationExtension.vue
```diff
@@ -58,10 +61,14 @@
             </div>
           </v-card>
 
+          <YourCompanySummary class="mt-10" />
+
           <DocumentsDelivery
             class="mt-10"
             sectionNumber="1."
             :validate="getAppValidate"
+            :userEmailOptional="userEmailOptional"
```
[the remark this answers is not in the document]
ketaki-deodhar: Here is the screenshot of Summary page:

image

src/views/LimitedRestorationToFull.vue
```diff
@@ -42,21 +46,36 @@
...
-          <ListPeopleAndRoles class="mt-10" />
-
           <DocumentsDelivery
             class="mt-10"
             sectionNumber="1."
             :validate="getAppValidate"
+            :userEmailOptional="userEmailOptional"
+            :applicantEmail="applicantEmail"
             @valid="setDocumentOptionalEmailValidity($event)"
           />
```
ketaki-deodhar, line 80: Here is the screenshot of Sumamry page:

image

src/views/LimitedRestorationToFull.vue
```diff
@@ -263,11 +288,14 @@ export default class LimitedRestorationToFull extends Vue {
       // set applicant orgPerson
       entitySnapshot.orgPersons = this.parseApplicantOrgPerson(applicant)
 
-      // Set the previously filed limited restoration in the store.
-      await this.setStateFilingRestoration()
+      this.setEntitySnapshot(entitySnapshot)
```
ketaki-deodhar, line 291: this fixes the opening Conversion applications (currently broken in DEV)

src/components/Restoration/RestorationSummary.vue
severinbeauvais: Please also remove the obsolete button event handler and the unused CSS.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -36,13 +60,18 @@
           </v-col>
         </v-row>
       </div>
+      <div class="section-container">
```
severinbeauvais, line 63: Should this be conditional (ie, show only if it has changed)?

src/components/common/QuestionWrapper.vue
```diff
@@ -39,7 +39,7 @@ export default {
   counter-reset: header-counter;
 }
 
-h2::before {
+#question_container h2::before {
```
severinbeauvais, line 42: How many question containers are rendered on a page? We should not have duplicate IDs, so maybe a class should be used here.

src/interfaces/restoration-interfaces/restoration-interface.ts
```diff
@@ -16,6 +16,7 @@ export interface RestorationIF {
   nameTranslations?: NameTranslationIF[]
   offices: AddressesIF
   parties: OrgPersonIF[]
-  relationships?: string[]
+  noticeDate?: string // YYYY-MM-DD
+  relationships?: RelationshipTypes[]
```
severinbeauvais, line 20: While you're here, please remove the ? from the required properties (and add the missing prop?) << @jonathan-longe Is `type` your prop?

Ref: https://github.com/bcgov/business-schemas/blob/1f068c23df4427d2ada72dc16c869884b105d24c/src/registry_schemas/schemas/restoration.json#L13

src/views/LimitedRestorationToFull.vue
```diff
@@ -337,6 +365,52 @@ export default class LimitedRestorationToFull extends Vue {
...
+  }
+
+  get showCourtOrderRadio (): boolean {
+    let courtOrderRadio
+    if (this.getStateFilingRestoration?.approvalType === ApprovalTypes.VIA_COURT_ORDER) {
+      courtOrderRadio = false
+    } else {
+      courtOrderRadio = true
+    }
+    return courtOrderRadio
```
severinbeauvais, line 411: No need to change, but next time you could do
`const courtOrderRadio = (something !== something_else)`

src/components/common/DocumentsDelivery.vue
severinbeauvais: Looks good.

No need to change but here's a shortcut: `return (this.applicantEmail || this.getUserEmail)`

src/components/common/DocumentsDelivery.vue
severinbeauvais: Could also do `return (a || b)` here, fwiw.

src/views/LimitedRestorationToFull.vue
```diff
@@ -27,7 +26,12 @@
           <QuestionWrapper
             id="approval-type"
             title="Approval Type">
```
severinbeauvais, line 28: Move > to a new line.

src/views/LimitedRestorationToFull.vue
```diff
@@ -171,18 +191,23 @@ export default class LimitedRestorationToFull extends Vue {
...
   @Action setResource!: ActionBindingIF
   @Action setStateFilingRestoration!: ActionBindingIF
+  @Action setRestorationRelationships!: ActionBindingIF
 
   /** Whether App is ready. */
   @Prop({ default: false }) readonly appReady!: boolean
   @Prop({ default: 0 }) readonly restorationId!: number
 
+  // Enum for template
+  readonly ApprovalTypes = ApprovalTypes
```
severinbeauvais, line 209: I think this isn't used in the template, so you don't need this.

src/components/Restoration/YourCompanySummary.vue
ketaki-deodhar: I don't think so. The current implementation of component is as below (screenshots)
If it not changed:
image

If changed, show changed label:
image

src/views/LimitedRestorationExtension.vue
```diff
@@ -271,6 +262,19 @@ export default class LimitedRestorationExtension extends Vue {
...
+      if (!restorationFiling.restoration.expiry) {
+        // New limited restoration extension
+        // Set the previously filed limited restoration in the store.
+        await this.setStateFilingRestoration()
+        // parse draft restoration filing into store
+        this.parseRestorationFiling(restorationFiling)
+      } else {
+        this.parseRestorationFiling(restorationFiling)
+        await this.setStateFilingRestoration()
+      }
```
severinbeauvais, line 276: I don't understand this code (Karim?)... why set state filing first then parse it, vs parsing it then setting state filing?

src/views/LimitedRestorationExtension.vue
```diff
@@ -271,6 +262,19 @@ export default class LimitedRestorationExtension extends Vue {
...
+
       if (!this.restorationResource) {
         throw new Error(`Invalid restoration resource entity type = ${this.getEntityType}`)
       }
```
severinbeauvais, line 280: Duplicate code on lines 282-283 and 285-286.

src/views/LimitedRestorationExtension.vue
ketaki-deodhar: Yes. Implemented by @JazzarKarim

src/views/LimitedRestorationExtension.vue
severinbeauvais, line 280: Is the code on line 293-294 still needed, since you deleted the notice?

Check what the other filings do... Do they all set the fee prices from filing data, or do some skip that? I'm wondering if it needs to be there for an external reason or not for this filing.

severinbeauvais: Ditto for LimitedRestorationToFull.vue lines 312-313.

src/views/LimitedRestorationExtension.vue
```diff
@@ -305,6 +309,10 @@ export default class LimitedRestorationExtension extends Vue {
     return this.getResource.changeData?.orgPersonInfo.orgPersonLabel
   }
 
+  get userEmailOptional (): boolean {
+    return this.getResource.userEmailOptional
+  }
+
   // build applicant orgPerson and assign id (uuid)
```
severinbeauvais, line 316: If you use a JSDoc style comment then your IDE tools should show you the comments when you hover on the function.

severinbeauvais: Ditto for LimitedRestorationToFull.vue line 316.

src/components/common/DocumentsDelivery.vue
ketaki-deodhar: I agree! I will rename the prop.

src/views/LimitedRestorationExtension.vue
JazzarKarim: There was a problem I couldn't solve unless I did this.

The problem was when I set the state filing then parse it, the template was rendering before I was able to set approval type court order file number. So when I try to continue a draft, the text field shows empty since I was getting null. When I parse then set, this fixed the problem. 

The other issue was when I create a new limited restoration extension, we have no expiry date. When I parse then set; in the `parseRestorationFiling` method in `filing-template-mixin.ts`, there was a problem. 
On line 806: https://github.com/bcgov/business-edit-ui/blob/09e0f265f5260a0379a610caf59798d62b7eef13/src/mixins/filing-template-mixin.ts#L806
`this.getStateFilingRestoration?.expiry` was returning null. 

I tried literally everything to fix this issue and the if statement was the only way I was able to work around it. The issue mainly was things being called before actually being set in the store etc.

src/components/Restoration/YourCompanySummary.vue
severinbeauvais: OK, that looks good. Thanks for confirming :)

src/views/LimitedRestorationExtension.vue
severinbeauvais: OK, may be a reactivity (or mount) issue that this works around.

1. Please create a Tech Debt ticket to fix this in the future.
2. Please add a comment to this code block quickly explaining the issue (and reference the new ticket number).

JazzarKarim: Sounds good. I'll be creating a new ticket now!

severinbeauvais: There is a dependency between `parseRestorationFiling()` and `setStateFilingRestoration` -- the parsing uses "approvalType" and "expiry" from the state filing.

Therefore these won't be stored if you don't set the state filing first. @JazzarKarim What tests can we run to verify what's broken by this?

severinbeauvais: Is there a scenario where `this.mutateRestorationApprovalType(this.getStateFilingRestoration?.approvalType)` (from Filing Template Mixin) could in fact store an empty (undefined) value?

JazzarKarim: No not any that I'm aware of. I've tried every single possible scenario before merging my PR back then.

severinbeauvais: OK, sounds mitigated for now.

Please post number of new ticket (to fix the root cause) here.

src/views/LimitedRestorationExtension.vue
ketaki-deodhar: @severinbeauvais ,other filing set this from Filing data. I will test a bit more after removing this.

image

src/views/LimitedRestorationExtension.vue
JazzarKarim: New ticket number: https://github.com/bcgov/entity/issues/15862

src/views/LimitedRestorationExtension.vue
severinbeauvais: Yes, please test a bit. (I see corrections don't have this.)

Anyway, it probably doesn't hurt to leave this.

src/views/LimitedRestorationExtension.vue
severinbeauvais: Ketaki, I'll wait for your code comment before resolving this conversation.

src/components/common/QuestionWrapper.vue
severinbeauvais: Since the ID is declared in LimitedRestorationExtension.vue, and it is the parent of all the children question wrappers, I'm thinking this file is where the auto-numbering CSS belongs.

@jonathan-longe , what do you think?

ketaki-deodhar: Modifying it was Jonathan's suggestions to make this work for restorations :)

src/views/LimitedRestorationExtension.vue
ketaki-deodhar: I tested after removing but the filling fails to load properly. I will leave it here for now. Might need a deep dive to remove this code.

src/views/LimitedRestorationExtension.vue
ketaki-deodhar: All good from my side. @JazzarKarim, I will add ticket number as reference as I am modifying the code in this ticket. Please let me know if you would like to add more explanation.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -23,7 +47,7 @@
           </v-col>
           <v-col cols="8" class="mt-n1" v-if="isLimitedConversionRestorationFiling">
             <div class="font-weight-bold">Conversion to Full Restoration</div>
-            <div>[TODO - Applicant's relationship: Director, Shareholder]</div>
+            <div>Applicant's relationship: {{ getRelationships }}</div>
```
jonathan-longe, line 50: Thanks for fixing that @ketaki-deodhar !

src/interfaces/restoration-interfaces/restoration-interface.ts
ketaki-deodhar: Somehow missed this. Fixing it now. @jonathan-longe, could you please let me know if `type` is your prop?

src/components/Restoration/RestorationSummary.vue
severinbeauvais: Thanks. LGTM.

src/components/common/DocumentsDelivery.vue
severinbeauvais: LGTM.

src/components/common/CertifySection.vue
```diff
@@ -71,7 +71,7 @@ export default class CertifySection extends Mixins(DateMixin) {
 
   /** Get the entity type in readable format */
   get readableEntityType (): string {
-    return GetCorpFullDescription(this.getEntityType)
+    return this.getResource?.certifyText ? this.getResource?.certifyText : GetCorpFullDescription(this.getEntityType)
```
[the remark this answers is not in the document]
severinbeauvais: Thanks. LGTM.

(PS If you touch this again, you could change to "return (a || b)", but you don't have to.)

src/views/LimitedRestorationExtension.vue
```diff
@@ -58,10 +56,14 @@
...
+          <YourCompanySummary class="mt-10" />
+
           <DocumentsDelivery
             class="mt-10"
             sectionNumber="1."
             :validate="getAppValidate"
+            :userEmailOptional="userEmailOptional"
+            :applicantEmail="applicantEmail"
             @valid="setDocumentOptionalEmailValidity($event)"
           />
```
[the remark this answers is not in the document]
severinbeauvais: Thanks. LGTM.

src/views/LimitedRestorationExtension.vue
severinbeauvais: Thanks. LGTM.

src/views/LimitedRestorationExtension.vue
severinbeauvais: Thanks. LGTM.

src/views/LimitedRestorationToFull.vue
severinbeauvais: Thanks. LGTM.

src/views/LimitedRestorationExtension.vue
severinbeauvais: Thanks. LGTM.

src/components/common/CertifySection.vue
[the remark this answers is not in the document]
ketaki-deodhar: Done

src/interfaces/restoration-interfaces/restoration-interface.ts
jonathan-longe: Nope, I don't recognize it.  It looks like `type` was added by @JazzarKarim

tests/unit/LimitedRestorationToFull.spec.ts
ketaki-deodhar: please ignore these recent updates. Need to figure out how to fix these tests,. Same for the Extension unit tests.

src/components/common/QuestionWrapper.vue
jonathan-longe: The `id="question_container"` has to exist in the parent in order for the `QuestionWrapper` to work.  I'd be okay changing it `id` to a `class` if there's going to be more than one set of prefix numbering displayed at a time.

src/interfaces/restoration-interfaces/restoration-interface.ts
JazzarKarim: Didn't add it also! It was already there when I worked on my PR. What's the issue exactly?

src/components/common/QuestionWrapper.vue
jonathan-longe: @severinbeauvais the point of adding the CSS to the 'QuestionWrapper` is so the logic and presentation is contained within the QuestionWrapper.  As a developer, you should be able to wrap the any common component and have it display the prefix numbering, as long as, you have a container labeled correctly using `id="questions_container"`.  If we move the CSS to parent, then we'll need to repeat the CSS everywhere it's needed which breaks the DRY principle.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -61,7 +99,18 @@ export default {
     },
     getCourtOrderFileNumber () {
       return this.getRestoration.courtOrder.fileNumber
+    },
+    getCompanyName (): string {
+      if (this.getNameRequestLegalName) return this.getNameRequestLegalName
+      return `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.`
+    },
```
jonathan-longe, line 106: I would suggest this logic might be better located in the store because the logic behind the numbered company name can be more complex than `${this.getBusinessNumber || '[Incorporation Number]'} B.C. Ltd.` and it's logic that I'm guessing we'll want to reuse.  If I remember correctly, these are some examples:
- 0123456 B.C. Community Contribution Company
- 0123456 B.C. Ltd.
- 0123456 BC Unlimited Liability Company

src/interfaces/restoration-interfaces/restoration-interface.ts
```diff
@@ -7,16 +7,17 @@ import { ContactPointIF } from '@bcrs-shared-components/interfaces/'
 // Ref: https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/restoration.json
 //
 export interface RestorationIF {
```
severinbeauvais, line 9: This interface still doesn't match the schema (required vs non-required properties).

src/components/common/QuestionWrapper.vue
severinbeauvais: Understood, but having that CSS in there means a couple of icky things:
- the parent has to declare an ID on a wrapper
- the CSS has to be global

I'm not sure which is the best approach.

src/store/state/state-model.ts
```diff
@@ -152,6 +152,7 @@ export const stateModel: StateModelIF = {
     businessNameValid: false,
     expiry: null,
     expiryValid: true,
+    relationships: null,
```
severinbeauvais, line 155: Would this be better as an empty array?

src/components/common/QuestionWrapper.vue
severinbeauvais: > I'm not sure which is the best approach.

My SOP in this case is to not change anything, so I'm OK for leaving this for now.

src/interfaces/restoration-interfaces/restoration-interface.ts
severinbeauvais: I was asking about required vs optional props, but there's another conversation for this below so I'm going to close this conversation.

src/store/state/state-model.ts
severinbeauvais: ^^ It depends what you do with it before data is set in it.

ketaki-deodhar: I think empty array is a the way to go.

src/interfaces/restoration-interfaces/restoration-interface.ts
ketaki-deodhar: Required properties are :
"type",
 "approvalType",
 "offices",
"parties",
 "contactPoint"

src/store/state/state-model.ts
severinbeauvais: Well, maybe. It depends. An empty array isn't falsy so ...

src/interfaces/restoration-interfaces/restoration-interface.ts
severinbeauvais: Looks good now. Thanks.

severinbeauvais (approved):

src/components/Restoration/YourCompanySummary.vue
ketaki-deodhar: @jonathan-longe, I will keep a note of this. This affects almost all filings nd needs to be fixed. How about a techDebt ticket? Does this case qualify for techDebt?

jonathan-longe: Sure, if it's difficult to change, then creating a tech debt ticket would be most appropriate.

JazzarKarim (approved):
Looks Good! 👍

jonathan-longe (approved):

src/components/common/QuestionWrapper.vue
jonathan-longe: > Understood, but having that CSS in there means a couple of icky things:
> 
> * the parent has to declare an ID on a wrapper
> * the CSS has to be global

@severinbeauvais this is an important issue to address.  In my view, CSS _is_ designed to be applied globally.  Yes, it's true, CSS can be put into the `<style>` section of a Vue component and this is an okay approach if the style should only apply to the component, but if we restrict ourselves to only applying styles per component, we won't enjoy the full benefits of CSS.  

But, I hear your concern.  Rather than placing the CSS in the QuestionsWrapper component with "scoped" removed from `<style scoped>`, it would be cleaner if we moved the CSS to a file in `/src/assets/sytles/`?  That way we don't end up having globally applicable CSS buried inside a component.

severinbeauvais: I think there are benefits to having scoped CSS as well as global CSS.

In this case, yes, this CSS is "more global" than the component in question, so moving it to the base CSS file should be considered. Do we know for sure that `question_wrapper` is used on only 1 page at a time? If not then this might not work as pages may use the same number value...?

jonathan-longe: >  Do we know for sure that question_wrapper is used on only 1 page at a time? If not then this might not work as pages may use the same number value...?

Yes, from what I've seen, that's true.  If it turns out not to be true, then perhaps we could use a class instead of an id.

severinbeauvais: OK, if it works for now then let's leave it as as ID for now.

> Rather than placing the CSS in the QuestionsWrapper component with "scoped" removed from <style scoped>, it would be cleaner if we moved the CSS to a file in /src/assets/sytles/? That way we don't end up having globally applicable CSS buried inside a component.

@ketaki-deodhar In your next PR, could you please the CSS to `base.scss` or `layout.scss`?

Thanks. I'll resolve this conversation now.

PR #468, opened 2023-04-05
15880 Fixed bug - saving expiry date in limited restoration conversions

JazzarKarim:
*Issue #:* /bcgov/entity#15880

*Description of changes:*
- No longer saving expiry date when doing limited restoration conversion to full

no longer saving expiry

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -300,7 +300,6 @@ export default class FilingTemplateMixin extends DateMixin {
       },
       restoration: {
         approvalType: this.getStateFilingRestoration?.approvalType,
-        expiry: this.getRestoration.expiry,
         type: this.getRestoration.type,
         business: {
           identifier: this.getBusinessId,
@@ -311,6 +310,11 @@ export default class FilingTemplateMixin extends DateMixin {
         contactPoint: this.getContactPoint,
         relationships: this.getRestoration.relationships
       }
+    }
+
+    // Set expiry date property if it's not null
+    if (this.getRestoration.expiry) {
+      filing.restoration.expiry = this.getRestoration.expiry
     }
 
     // Apply NR / business name / business type change to filing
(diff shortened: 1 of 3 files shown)


src/mixins/filing-template-mixin.ts
```diff
@@ -831,9 +831,11 @@ export default class FilingTemplateMixin extends DateMixin {
     this.mutateRestorationType(filing.restoration.type)
     if (filing.restoration.expiry) {
       this.mutateRestorationExpiry(filing.restoration.expiry)
-    } else {
+    } else if (filing.restoration.type === RestorationTypes.LTD_EXTEND) {
       // Reset radio button to 2 years
       this.mutateRestorationExpiry(DateUtilities.addMonthsToDate(24, this.getStateFilingRestoration?.expiry))
+    } else {
+      this.mutateRestorationExpiry(undefined)
```
severinbeauvais, line 838: This is not a good idea because undefined is not reactive, so something watching this (or dependent on it in a getter) won't be triggered on a future change.

severinbeauvais: Or test it and tell me it works as-is :)

severinbeauvais: Do we even need this line?

JazzarKarim: At first I set it as null but when I save and then resume, I'm getting a field expiry: null. Setting it to undefined eliminates this field.

I've tested it and it works as is. This is since we only use any form of dates in extensions. Any getter that is dependent on the expiry would be there. However, I'm open to suggestions for a better way if applicable?

severinbeauvais: `expiry` is initialized to null in the store (see state-model.ts), so if you don't set it here when parsing a draft filing then it stays at that value -- which is good.

Where do  you see "expiry: null" ? Perhaps that component needs to handle this case. Expiry isn't meaningful if the restoration type is Conversion, right?

JazzarKarim: Yes. If restoration type is Conversion, it doesn't make sense to have expiry there. We also see that when creating a new Full Restoration in Create UI. 

This is an example of when we keep expiry as null in Conversions:
expiry null

severinbeauvais: 1. What do you get if you save a draft limited resto, then you resume the draft and save a draft full resto?
2. The null expiry above may be rejected by the API. In that case you'll have to remove it (as we do with some other properties).

JazzarKarim: 1. No expiry.
fullExpiry

2. See next commit.

ketaki-deodhar (approved):

jonathan-longe (approved):

PR #472, opened 2023-04-07
15927 - Pinia upgrade

seeker25:
https://github.com/bcgov/entity/issues/15927

Testing Criteria:
Full regression - anything that touches the store should be tested. 

Poem:
The old state manager served us well,
But now we seek a newer spell,
To handle data with greater ease,
And bring our code up to speed.

Pinia is the fresh new face,
With reactivity at its base,
It simplifies our code so fine,
And makes our app run like a vine.

Migrating may take some time,
But the benefits are surely prime,
For Pinia's speed and cleaner style,
Will make our app run mile by mile.

So let us bid farewell to old,
And embrace the new with arms so bold,
For Pinia's here to take the lead,
And bring our code up to speed.


package.json
```diff
@@ -90,5 +95,8 @@
     "vuetify-loader": "^1.9.2",
     "vuex-class": "^0.3.2",
     "vuex-module-decorators": "^1.2.0"
+  },
+  "overrides": {
+    "webpack": "5.78"
```
seeker25, line 100: This needed to be included, if it wasn't - it would run two versions of webpack (vue-cli-service would target the old one)

src/main.ts
```diff
@@ -103,7 +103,9 @@ async function start () {
       }
     }),
     router: getVueRouter(),
+    // We still need Vuex for sbc-common-components.
     store: getVuexStore(),
```
seeker25, line 107: Still need Vuex for sbc-common-components

tests/unit/Actions.spec.ts
```diff
@@ -256,37 +242,37 @@ describe.skip('Emits error event if NR validation fails in file and pay', () =>
...
+    store.stateModel.validationFlags = {
       flagsCompanyInfo: {
         isValidOrgPersons: true,
         isValidShareStructure: true
       },
       flagsReviewCertify: {
         isValidEffectiveDate: true
       }
-    }
+    } as any
```
seeker25, line 269: as any aren't ideal.. we should probably fix these. But they'll need to have the required fields.

tests/unit/ChangeBusinessType.spec.ts
```diff
@@ -46,25 +47,25 @@ describe('Change Business Type component', () => {
 
   it('Should have tooltip for Cooperative', () => {
     // init entity type
-    store.state.stateModel.tombstone.entityType = 'CP'
+    store.stateModel.tombstone.entityType = CorpTypeCd.COOP
```
seeker25, line 50: I think I remember Severin you say something about being cautious about using `CorpTypeCd` enums.. 

I did in a couple of spots put in some asserts to check the enum values.

.browserslistrc
```diff
@@ -1,3 +1,3 @@
 > 1%
 last 2 versions
-not ie <= 10
+not dead
```
seeker25, line 3: I changed this otherwise, Vue will compile a legacy package version and a modern package version. Let me know if this is ok.

src/store/store.ts
```diff
@@ -0,0 +1,1398 @@
...
+      const isOrganization = this.getOrgPeople[0]?.officer?.partyType === PartyTypes.ORGANIZATION
+      const isDba = isProprietor && isOrganization
+      return !this.isRoleStaff && isDba
+    },
+
+    getExpiryDateString (): string {
+      return this.stateModel.restoration?.expiry
+    },
+
+    getFormattedExpiryText: (state) => (today = new Date()): string => {
```
seeker25, line 1125: This was a tad annoying to deal with

thorwolpert:
the poem was good, you should leave it in!

tests/unit/Actions.spec.ts
severinbeauvais: I think we shouldn't use types in the unit tests.

tests/unit/ChangeBusinessType.spec.ts
severinbeauvais: I don't remember why I said that, but I think you shouldn't use enums in the unit tests.

.browserslistrc
severinbeauvais: Yeah, we don't support IE11 or earlier.

tests/unit/ChangeBusinessType.spec.ts
seeker25: I disagree with you, but I'll take them out if you'd like

seeker25:
There's still an issue with the webpack SCSS bundling, it's not including the hashes files in the index.html. I'm still trying to figure that out.

seeker25:
I think this is functional now. I'll grab the changes from https://github.com/bcgov/business-edit-ui/pull/470 and integrate them in after someone takes a quick peak at this

src/store/store.ts
jonathan-longe: @seeker25 apologies for causing this annoyance.  Help me understand what you found annoying so I don't cause problems in the future.

jonathan-longe:
> the poem was good, you should leave it in!

Until now, I had know idea that poet and developer were on the same chipset!

src/store/index.ts
```diff
@@ -1,22 +1,26 @@
...
 }
+
+/**
+ * Configures and returns Pinia Store.
+ */
+export function getPiniaStore () {
+  Vue.use(PiniaVuePlugin)
+
+  return createPinia()
+}
```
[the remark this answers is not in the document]
thorwolpert: I think the more common approach in Vuex is to separate them, but in Pinia they are more often together in a single module. I keep mine together and it follows the useXxxxStore pattern better which I think is promoted by the creator of Pinia

src/App.vue
```diff
@@ -449,12 +450,12 @@ export default class App extends Vue {
...
-        addressRegion: isStaff ? '' : this.getOrgInfo?.mailingAddress.region,
-        postalCode: isStaff ? '' : this.getOrgInfo?.mailingAddress.postalCode,
-        streetAddress: isStaff ? '' : this.getOrgInfo?.mailingAddress.street,
-        streetAddressAdditional: isStaff ? '' : this.getOrgInfo?.mailingAddress.streetAdditional
+        addressCity: isStaff ? '' : this.getOrgInfo?.mailingAddress?.city,
+        addressCountry: isStaff ? '' : this.getOrgInfo?.mailingAddress?.country,
+        addressRegion: isStaff ? '' : this.getOrgInfo?.mailingAddress?.region,
+        postalCode: isStaff ? '' : this.getOrgInfo?.mailingAddress?.postalCode,
+        streetAddress: isStaff ? '' : this.getOrgInfo?.mailingAddress?.street,
+        streetAddressAdditional: isStaff ? '' : this.getOrgInfo?.mailingAddress?.streetAdditional
```
severinbeauvais, line 458: Are you sure this extra conditional chaining is needed -- did you see an example of an org without a mailing address object?

src/components/Conversion/ConversionNOB.vue
```diff
@@ -133,6 +134,7 @@ export default class ConversionNOB extends Mixins(CommonMixin) {
 
   /** Called when user has clicked the Done button. */
   protected onDoneClicked (): void {
+    // eslint-disable-next-line no-undef
```
severinbeauvais, line 137: Must be a new eslint warning. Is there an easy way to fix it? Thanks for the disable for now.

src/components/Restoration/YourCompanySummary.vue
```diff
@@ -80,7 +80,7 @@ export default {
     }
   },
   computed: {
-    ...mapGetters([
+    ...mapState(useStore, [
```
severinbeauvais, line 83: Are we going from using getters to reading state directly?

src/App.vue
seeker25: was changed to pass the unit test, will try it without and see if it can be fixed

src/components/Restoration/YourCompanySummary.vue
seeker25: image

seeker25: image

src/components/common/EffectiveDateTime.vue
```diff
@@ -74,15 +74,19 @@
...
 import { DateMixin } from '@/mixins/'
 import { EffectiveDateTypes } from '@/enums/'
 import { EffectiveDateTimeIF, FormFieldType, FormIF } from '@/interfaces/'
 import { VuetifyRuleFunction } from '@/types'
 
-enum PeriodTypes {
+import { useStore } from '@/store/store'
+
+export enum PeriodTypes {
+  // eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
seeker25: Yup agreed - was weird that this error was showing up even though it's used in the component. My first pass I'd rather not change too much functionality as alot of files already have plenty of changes.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "4.0.7",
+  "version": "4.0.8",
```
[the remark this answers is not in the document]
seeker25: Changed to 4.2.0, I'm just working on pulling in the other PR now

seeker25:
"I wonder, though, how this work was approved for this sprint?"

Needed to be done - especially with the way we wanted to rewrite the components to be more reliant on the store.  If we wanted to build components to the status quo, probably could have waited. 

"we have several people already working in this project, so there will be lots of rebasing and merging" 

Yeah same with Vue 2.7 upgrade. These upgrades put you in a way better position for Vue 3, if you wanted to take the Vue-cli-service route. You're also using Typescript 4 now instead of 3. Webpack 5. These upgrades probably should have happened with the Vue 2.7 upgrade honestly. 

"we will now have a separate store architecture (or at least package) in this UI compared with the other ones"

The pattern has already been figured out - I've already created a partial PR for `SBC-AUTH` and will be upgrading to Pinia there as well. I don't think it will be very difficult to do `business-filings-ui` etc. and will do it myself if needed.

src/components/Alteration/Articles/CompanyProvisions.vue
```diff
@@ -121,21 +121,22 @@
...
 import { Component, Emit, Prop, Mixins, Watch } from 'vue-property-decorator'
-import { Getter } from 'vuex-class'
+import { Getter } from 'pinia-class'
 import { CommonMixin } from '@/mixins/'
+import { useStore } from '@/store/store'
 
 @Component({})
 export default class CompanyProvisions extends Mixins(CommonMixin) {
+  @Getter(useStore) getEditLabel!: string
+  @Getter(useStore) getEditedLabel!: string
```
severinbeauvais, line 131: Agreed -- "external imports" should be at the top of the file/class.

src/components/common/YourCompany/NameTranslations/ListNameTranslation.vue
```diff
@@ -138,14 +138,15 @@ import { Getter } from 'vuex-class'
 import { NameTranslationIF } from '@/interfaces/'
 import { ActionTypes } from '@/enums/'
 import { CommonMixin } from '@/mixins/'
+import { useStore } from '@/store/store'
 
 @Component({})
 export default class ListNameTranslation extends Mixins(CommonMixin) {
+  @Getter(useStore) isCorrectionFiling!: boolean
```
severinbeauvais, line 145: Also update line 137?

src/mixins/common-mixin.ts
```diff
@@ -1,7 +1,9 @@
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
+import { Getter } from 'pinia-class'
 import { RouteNames } from '@/enums/'
 import { ConfirmDialogType } from '@/interfaces/'
+import { useStore } from '@/store/store'
```
severinbeauvais, line 6: Looks like we no longer use a Getter in this class, so this and line 3 can be deleted. Please and thanks!

src/mixins/filing-template-mixin.ts
```diff
@@ -14,90 +14,90 @@ import { ActionTypes, CoopTypes, CorrectionErrorTypes, EffectOfOrders, FilingTyp
...
+  @Action(useStore) setProvisionsRemoved!: ActionBindingIF
+  @Action(useStore) setNewResolutionDates!: ActionBindingIF
+  @Action(useStore) setFileNumber!: ActionBindingIF
+  @Action(useStore) setHasPlanOfArrangement!: ActionBindingIF
+  @Action(useStore) setSpecialResolution!: ActionBindingIF
+  @Action(useStore) setCorrectionStartDate!: ActionBindingIF
+  @Action(useStore) setRestorationApprovalType!: ActionBindingIF
+  @Action(useStore) setRestorationCourtOrder!: ActionBindingIF
+  @Action(useStore) setRestorationExpiry!: ActionBindingIF
+  @Action(useStore) setRestorationType!: ActionBindingIF
```
severinbeauvais, line 100: 1. Any idea if there's a way to import the entire getters/actions instead?

2. I've seen where components that use this mixin inherit these imports, but I don't really like that. I think it's better for components to be explicit and precise about what getters and actions they use.

3. Maybe point 2 above means we shouldn't do point 1... thoughts?

seeker25: 1. I bet there probably is a way to import all of the getters / actions. 

2. You get typescript errors - it asks you to use declare to override 

3. You could change it all you want after this PR. I just wanted to maintain similar structure, that way people don't get too side tracked.

src/store/store.ts
```diff
@@ -0,0 +1,1448 @@
...
+        const formattedExpiryText = numberOfExtensionMonths + ' months, expires on ' + expiryDatePacific
+        return formattedExpiryText
+      }
+      return '[no expiry date]'
+    }
+  },
+  actions: {
+    // no context as first argument, use `this` instead
+    setEntityType (entityType: CorpTypeCd) {
+      this.stateModel.tombstone.entityType = entityType
```
severinbeauvais, line 1189: So actions are the new mutations?

src/store/store.ts
```diff
@@ -0,0 +1,1448 @@
...
+import { resourceModel, stateModel } from './state'
+import { LegalServices } from '@/services'
+
+export const useStore = defineStore('store', {
+  // convert to a function
+  state: (): StateIF => ({
+    stateModel: stateModel,
+    resourceModel: resourceModel
+  }),
+  getters: {
```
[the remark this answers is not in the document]
seeker25: I'd say split the stores into separate responsibilities would probably be the best way to solving that issue

src/store/store.ts
seeker25: Yes

src/views/LimitedRestorationExtension.vue
```diff
@@ -337,7 +338,7 @@ export default class LimitedRestorationExtension extends Vue {
 
   get applicantEmail (): string {
     const currentApplicant = this.getOrgPeople.filter(orgPerson => !this.wasRemoved(orgPerson))
-    return currentApplicant[0].officer.email
+    return currentApplicant[0]?.officer.email
```
severinbeauvais, line 341: 👍

src/views/LimitedRestorationToFull.vue
```diff
@@ -308,12 +308,12 @@ export default class LimitedRestorationToFull extends Vue {
...
-        lastName: applicant.officer.lastName,
-        middleName: applicant.officer.middleName,
-        organizationName: applicant.officer.organizationName,
-        partyType: applicant.officer.partyType,
+        email: applicant?.officer.email,
+        firstName: applicant?.officer.firstName,
+        lastName: applicant?.officer.lastName,
+        middleName: applicant?.officer.middleName,
+        organizationName: applicant?.officer.organizationName,
+        partyType: applicant?.officer.partyType,
```
severinbeauvais, line 316: Shouldn't need to do this since `applicant` was checked on line 261.

If the intent is to protect this method from bad data then applicant should be checked directly, maybe on line 306.

If we must use conditional chaining here then it should be done lines 308, 309 and 319 as well (but my vote is we don't need CC here).

tests/unit/Actions.spec.ts
```diff
@@ -55,37 +58,14 @@ const nrData = {
...
-  let setChanged: any
-  let setValidity: any
-  let setEditing: any
 
   beforeAll(async () => {
     // initialize store
-    store.state.stateModel.tombstone.filingType = 'correction'
-    store.state.stateModel.correctionInformation.type = 'CLIENT'
+    store.stateModel.tombstone.filingType = FilingTypes.CORRECTION
+    store.stateModel.correctionInformation.type = CorrectionErrorTypes.CLIENT
```
severinbeauvais, line 65: I think we shouldn't use types/IFs or enums in the unit tests, so that they are more fragile, and so that an error in a type or enum isn't cancelled out in the unit tests.

tests/unit/Actions.spec.ts
```diff
@@ -55,37 +58,14 @@ const nrData = {
...
-
-    setValidity = async (val: boolean) => {
-      // set certify state flag since it's initially false
-      await wrapper.vm.$store.commit('mutateCertifyStateValidity', val)
-    }
-
-    setEditing = async (val: boolean) => {
-      // set any editing flag
-      await wrapper.vm.$store.commit('mutateEditingCompanyName', val)
-    }
```
severinbeauvais, line 88: Yes, this was probably wrong since we shouldn't have to await mutations (I think).

However, some actions are async and need to be awaited, fwiw.

seeker25: I think I just rolled this into the actions

tests/unit/Actions.spec.ts
```diff
@@ -102,7 +82,9 @@ describe('Action button states', () => {
   })
 
   it('renders buttons once Entity Type is known', async () => {
-    await setEntityType('BEN')
+    store.setEntityType(CorpTypeCd.BENEFIT_COMPANY)
+
+    await Vue.nextTick()
```
severinbeauvais, line 87: Did you add this because the test failed otherwise, or did you add it because we were previously awaiting the mutation?

Any idea if we always have to await actions now?

tests/unit/AlterationSummary.spec.ts
```diff
@@ -33,21 +37,21 @@ describe('Alteration Summary component', () => {
...
   beforeAll(() => {
     // init store
-    store.state.stateModel.currentJsDate = new Date('2020-03-01T16:30:00Z')
-    store.state.stateModel.tombstone.currentDate = '2021-03-01'
-    store.state.stateModel.entitySnapshot = entitySnapshot
-    store.state.stateModel.shareStructureStep.shareClasses = []
-    store.state.stateModel.entitySnapshot.shareStructure = { shareClasses: [] }
+    store.stateModel.currentJsDate = new Date('2020-03-01T16:30:00Z')
+    store.stateModel.tombstone.currentDate = '2021-03-01'
+    store.stateModel.entitySnapshot = entitySnapshot as any
```
severinbeauvais, line 42: Or could add `any` to entitySnapshot. Ditto for nameTranslationsListChanged .

tests/unit/App.spec.ts
```diff
@@ -345,52 +349,53 @@ describe.skip('Numbered company setup', () => {
...
-    expect(store.state.stateModel.tombstone.filingId).toBe(54321)
+    expect(store.stateModel.nameRequest.nrNumber).toEqual('')
+    expect(store.stateModel.tombstone.filingId).toBe(54321)
 
     // Validate no NR Details
-    expect(store.state.stateModel.nameRequest.details.approvedName).toBeUndefined()
-    expect(store.state.stateModel.nameRequest.details.status).toBeUndefined()
-    expect(store.state.stateModel.nameRequest.details.consentFlag).toBeUndefined()
-    expect(store.state.stateModel.nameRequest.details.expirationDate).toBeUndefined()
+    expect((store.stateModel.nameRequest.details as any).approvedName).toBeUndefined()
```
severinbeauvais, line 381: Why is this needed -- is `details` typed now? (That would be awesome.)

tests/unit/Change.spec.ts
```diff
@@ -211,62 +213,63 @@ describe('Change component', () => {
...
+    store.stateModel.tombstone.keycloakRoles = []
+    store.stateModel.summaryMode = false
     wrapper.vm.showFee = true
   })
 
   it('resource subtitle states individual sole proprietors can change legal name', () => {
-    store.state.stateModel.tombstone.entityType = 'SP'
-    store.state.stateModel.peopleAndRoles.orgPeople = [
+    store.stateModel.tombstone.entityType = CorpTypeCd.SOLE_PROP
+    expect(CorpTypeCd.SOLE_PROP).toBe('SP')
```
severinbeauvais, line 252: I think we probably don't need to test enums...?

tests/unit/Change.spec.ts
```diff
@@ -286,40 +289,41 @@ describe('Change component', () => {
...
 
     // Validate Office Addresses
-    expect(state.officeAddresses.deliveryAddress.streetAddress)
-      .toBe('reg delivery_address - address line one')
-    expect(state.officeAddresses.mailingAddress.streetAddress)
-      .toBe('rec mailing_address - address line two')
+    // expect(state.officeAddresses.deliveryAddress.streetAddress)
+    //   .toBe('reg delivery_address - address line one')
+    // expect(state.officeAddresses.mailingAddress.streetAddress)
+    //   .toBe('rec mailing_address - address line two')
```
severinbeauvais, line 295: Needs to be fixed.

tests/unit/ChangeBusinessType.spec.ts
severinbeauvais: Curious what you think about this.

tests/unit/Change.spec.ts
seeker25: Sounds good, wasn't sure if you wanted to put it in there incase someone changed the `@\enums\` reference and thus cause an issue with all of the enums changing

tests/unit/CompanyProvisions.spec.ts
```diff
@@ -31,7 +35,6 @@ describe('company provisions', () => {
     const wrapper = shallowMount(CompanyProvisions,
       {
         vuetify,
-        store,
```
severinbeauvais, line 34: Since we no longer need to inject store into the component instance, does that mean the store is "external"?

tests/unit/Change.spec.ts
seeker25: Future unit test. The data structure is wrong

tests/unit/CompanyProvisions.spec.ts
seeker25: Yes, we don't use the vuex store at all - sbc-header uses it though and imports it in sbc-common-components

tests/unit/Change.spec.ts
seeker25: image

src/components/common/YourCompany/CompanyName/CorrectNameOptions.vue
```diff
@@ -59,10 +59,9 @@ import Vue from 'vue'
 import { Component, Emit, Prop } from 'vue-property-decorator'
 import { CorrectNameOptionIF } from '@/interfaces/'
 import { NameChangeOptions } from '@/enums/'
-
-// for some reason, CorrectXXX cannot be imported from ./
-// (it breaks a bunch of unit tests)
-import { CorrectCompanyName, CorrectNameToNumber, CorrectNameRequest } from '@/components/common/'
+import CorrectCompanyName from './CorrectCompanyName.vue'
+import CorrectNameRequest from './CorrectNameRequest.vue'
+import CorrectNameToNumber from './CorrectNameToNumber.vue'
```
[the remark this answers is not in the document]
seeker25: It did, those components wouldn't load until I changed these imports - gave me a bit of a heads up.

tests/unit/name-request-mixin.spec.ts
```diff
@@ -2,29 +2,32 @@ import Vue from 'vue'
...
   beforeEach(async () => {
-    store.state.stateModel.tombstone.entityType = 'BEN'
-    store.state.resourceModel = {
+    store.stateModel.tombstone.entityType = CorpTypeCd.BENEFIT_COMPANY
+    store.resourceModel = {
       entityReference: '',
       displayName: '',
       addressLabel: '',
-      filingData: '',
+      filingData: null,
```
severinbeauvais, line 23: 👍

tests/unit/YourCompany.spec.ts
```diff
@@ -86,7 +90,8 @@ describe('YourCompany in a BEN correction', () => {
 })
 
 describe('YourCompany in a SP alteration', () => {
-  const store = getVuexStore()
+  setActivePinia(createPinia())
+  const store = useStore()
```
severinbeauvais, line 94: Any idea if each describe block just uses 1 global store (in this file), or are they all separate instances?

tests/unit/ChangeBusinessType.spec.ts
seeker25: I think they should be left in, otherwise the store type definition should be changed. 

Here's a prime example of where ENUM's went wrong recently with AUTH-WEB (not in the unit tests.. but same idea):

https://github.com/bcgov/sbc-auth/pull/2303/files

seeker25: This problem could have been identified at compile time, if we used enums instead of strings there. Or would have been a non issue because the enum value would have been correct versus the string that was entered.

src/store/store.ts
seeker25: Tricky syntax - wasn't too bad though

src/store/store.ts
seeker25: Might be a good idea, not sure if the class way can identify through typescript if the action returns a promise or not

tests/unit/LimitedRestorationExtension.spec.ts
```diff
@@ -261,16 +262,17 @@ describe('Restoration component - edit page', () => {
...
 
     // Validate business identifier
     expect(state.tombstone.businessId).toBe('BC1234567')
 
     // Validate Business
     expect(state.businessInformation.legalType).toBe('BEN')
     expect(state.businessInformation.legalName).toBe('1234567 B.C. LTD.')
-    expect(state.businessInformation.stateFiling.restoration.type).toBe('limitedRestoration')
-    expect(state.businessInformation.stateFiling.restoration.parties[0].role).toBe('applicant')
+    // Travis Semple - stateFiling is defined as a string here, not an object. should be fixed.
```
severinbeauvais, line 273: Dropping a comment here so I can check this is fixed by final commit :)

src/views/LimitedRestorationExtension.vue
seeker25: I've left this one, if you want it reverted let me know

src/views/LimitedRestorationToFull.vue
seeker25: Reverted

tests/unit/App.spec.ts
seeker25: Reverted, not sure why it wouldn't pick up details before

tests/unit/ErrorContact.spec.ts
```diff
@@ -2,6 +2,8 @@ import Vue from 'vue'
 import Vuetify from 'vuetify'
 import { mount } from '@vue/test-utils'
 import ErrorContact from '@/components/common/ErrorContact.vue'
+import { createPinia, setActivePinia } from 'pinia'
+import { useStore } from '@/store/store'
```
severinbeauvais, line 6: Looks like these aren't needed -- please remove.

tests/unit/ResolutionDates.spec.ts
```diff
@@ -65,15 +69,15 @@ const shareClasses = [
     'hasRightsOrRestrictions': true,
     'series': []
   }
-]
+] as unknown as ShareClassIF[]
```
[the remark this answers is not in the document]
seeker25: If i remove it: 

image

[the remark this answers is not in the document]
seeker25: I could cast it `as any` for now, but technically someone should fix the test data's data structure

tests/unit/LimitedRestorationExtension.spec.ts
seeker25: I think this is out of the scope of implementing Pinia - this is an existing bug

tests/unit/CorrectNameOptions.spec.ts
```diff
@@ -6,23 +6,25 @@ import CorrectNameOptions from '@/components/common/YourCompany/CompanyName/Corr
...
 import CorrectNameRequest from '@/components/common/YourCompany/CompanyName/CorrectNameRequest.vue'
 import CorrectNameToNumber from '@/components/common/YourCompany/CompanyName/CorrectNameToNumber.vue'
+import { createPinia, setActivePinia } from 'pinia'
+import { useStore } from '@/store/store'
 
 Vue.use(Vuetify)
 
 const vuetify = new Vuetify({})
+setActivePinia(createPinia())
+const store = useStore()
```
severinbeauvais, line 16: I believe this component doesn't use the store, so this can be deleted. Please and thanks!

src/views/LimitedRestorationToFull.vue
```diff
@@ -358,7 +358,7 @@ export default class LimitedRestorationToFull extends Vue {
 
   get applicantEmail (): string {
     const currentApplicant = this.getOrgPeople.filter(orgPerson => !this.wasRemoved(orgPerson))
-    return currentApplicant[0].officer.email
+    return currentApplicant[0]?.officer?.email
```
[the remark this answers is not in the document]
seeker25: Fixed

tests/unit/Change.spec.ts
seeker25: Removed enum checks

tests/unit/App.spec.ts
```diff
@@ -603,15 +609,23 @@ describe('App component - other', () => {
...
+        data: {
+          mailingAddress: {
+            city: 'Victoria',
+            country: 'CA',
+            postalCode: 'V8W 3E6',
+            region: 'BC',
+            street: '2-940 Blanshard St',
+            streetAdditional: ''
+          }
+        }
```
severinbeauvais, line 621: Hah, so this was an example of an org with missing mailing address (which I asked about earlier).

Therefore, having the data above is needed, but I think the conditional chaining isn't needed (or desired).

I guess my approach is I'd rather see an error than a hidden fallback value.

tests/unit/Actions.spec.ts
severinbeauvais: Is the new linter (or Typescript) complaining about undeclared properties?

src/components/Conversion/ConversionNOB.vue
severinbeauvais: The *.d.ts files may need to be reviewed. I previously (with [Vetur](https://vuejs.github.io/vetur/)) but no longer (with [Volar](https://marketplace.visualstudio.com/items?itemName=Vue.volar)) had full context help with properties and methods. It would be nice to get that back.

src/components/common/EffectiveDateTime.vue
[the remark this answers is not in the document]
severinbeauvais: I see you created a new external enum so this is resolved. Thanks.

severinbeauvais:
> "I wonder, though, how this work was approved for this sprint?"
> 
> Needed to be done - especially with the way we wanted to rewrite the components to be more reliant on the store. If we wanted to build components to the status quo, probably could have waited. I was told we want to integrate components more deeply into the store, but at the same time we shouldn't be trying to reinventing the wheel and writing a bunch of code that could possibly be replaced already by Pinia.

Fair enough. It wouldn't have been my decision to do this now but I'll go with it. (Better now than any time in the recent weeks 😛 )

> "we have several people already working in this project, so there will be lots of rebasing and merging"
> 
> Yeah same with Vue 2.7 upgrade. These upgrades put you in a way better position for Vue 3, if you wanted to take the Vue-cli-service route. You're also using Typescript 4 now instead of 3. Webpack 4 -> Webpack 5. These upgrades probably should have happened with the Vue 2.7 upgrade honestly.

I tried some of those upgrades in the fall and found a lot of conflicting dependencies. Thanks for resolving it this time around.

> I also noticed some of the unit test debugging wasn't working correctly, and this has recently been resolved. You can actually step into components now and view their source.

Nice.

> "we will now have a separate store architecture (or at least package) in this UI compared with the other ones"
> 
> The pattern has already been figured out - I've already created a partial PR for `SBC-AUTH` and will be upgrading to Pinia there as well. I don't think it will be very difficult to do `business-filings-ui` etc. and will do it myself if needed.

Yes, thanks for the pattern.

Please create tickets for the other UIs and we can plan them accordingly. I don't want you to do it all but I would like to know ahead of time when is is done.

seeker25:
> > "I wonder, though, how this work was approved for this sprint?"
> > Needed to be done - especially with the way we wanted to rewrite the components to be more reliant on the store. If we wanted to build components to the status quo, probably could have waited. I was told we want to integrate components more deeply into the store, but at the same time we shouldn't be trying to reinventing the wheel and writing a bunch of code that could possibly be replaced already by Pinia.
> 
> Fair enough. It wouldn't have been my decision to do this now but I'll go with it. (Better now than any time in the recent weeks 😛 )
> 
> > "we have several people already working in this project, so there will be lots of rebasing and merging"
> > Yeah same with Vue 2.7 upgrade. These upgrades put you in a way better position for Vue 3, if you wanted to take the Vue-cli-service route. You're also using Typescript 4 now instead of 3. Webpack 4 -> Webpack 5. These upgrades probably should have happened with the Vue 2.7 upgrade honestly.
> 
> I tried some of those upgrades in the fall and found a lot of conflicting dependencies. Thanks for resolving it this time around.
> 
> > I also noticed some of the unit test debugging wasn't working correctly, and this has recently been resolved. You can actually step into components now and view their source.
> 
> Nice.
> 
> > "we will now have a separate store architecture (or at least package) in this UI compared with the other ones"
> > The pattern has already been figured out - I've already created a partial PR for `SBC-AUTH` and will be upgrading to Pinia there as well. I don't think it will be very difficult to do `business-filings-ui` etc. and will do it myself if needed.
> 
> Yes, thanks for the pattern.
> 
> Please create tickets for the other UIs and we can plan them accordingly. I don't want you to do it all but I would like to know ahead of time when is is done.

I did too, I've tried probably at least 5 times to get the libraries upgraded. I'll create the tickets now and update this PR as well to reference the ticket.

src/store/store.ts
severinbeauvais, line 59: Given that getters and actions are separate objects anyway, I'm leaning towards separate files/imports for them. I see no advantage of having everything in this one file.

src/App.vue
severinbeauvais: Reverted, thanks.

src/components/common/YourCompany/CompanyName/CorrectNameOptions.vue
```diff
@@ -127,7 +126,7 @@ export default class CorrectNameOptions extends Vue {
   }
 
   get isOneOption (): boolean {
-    return (this.correctionNameChoices.length === 1)
+    return (this.correctionNameChoices?.length === 1)
```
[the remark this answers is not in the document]
severinbeauvais: Reverted, thanks.

src/components/common/YourCompany/YourCompany.vue
```diff
@@ -494,12 +495,12 @@ export default class YourCompany extends Mixins(
   get nameChangeOptions (): Array<NameChangeOptions> {
     // if this is a numbered company, remove correct-name and name-to-number options
     if (this.isNumberedCompany) {
-      return this.getResource.changeData.nameChangeOptions.filter(option => (
+      return this.getResource.changeData?.nameChangeOptions?.filter(option => (
```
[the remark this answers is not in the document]
severinbeauvais: Reverted, thanks.

src/mixins/common-mixin.ts
severinbeauvais: Fixed, thanks.

src/mixins/filing-template-mixin.ts
severinbeauvais: OK, don't worry about it. Thanks for looking into it.

seeker25:
https://github.com/bcgov/entity/issues/15931
https://github.com/bcgov/entity/issues/15930
https://github.com/bcgov/entity/issues/15929
https://github.com/bcgov/entity/issues/15928
https://github.com/bcgov/entity/issues/15927

src/views/LimitedRestorationExtension.vue
severinbeauvais: I like this one :)

tests/unit/Actions.spec.ts
seeker25: Yup

src/store/store.ts
seeker25: We can move them after this PR - I'd rather keep them together and have smaller stores - but it is possible:

https://github.com/vuejs/pinia/issues/802#issuecomment-1018780409

tests/unit/LimitedRestorationExtension.spec.ts
severinbeauvais: This may be confusion between the state filing object and the state filing URL (string).

Yes, this is out of scope. Do you mind creating a bug/techdebt ticket? Thx,

tests/unit/ResolutionDates.spec.ts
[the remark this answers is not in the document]
severinbeauvais: Yeah, we need to figure out our approach: totally untyped or fully typed unit tests. Thoughts?

tests/unit/Change.spec.ts
severinbeauvais: Awesome, thanks!

severinbeauvais: The take-away here is that typing in the unit tests is nice to have?

tests/unit/ResolutionDates.spec.ts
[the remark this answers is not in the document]
seeker25: full typed

tests/unit/Actions.spec.ts
severinbeauvais: In this example, it is returning a promise, since it makes a network call and doesn't resolve until that's done.

So, do we need to await some actions?

And should we type actions differently when we import them? cc: @jonathan-longe

tests/unit/App.spec.ts
seeker25: Build can't handle `interface | object` for some reason

tests/unit/ResolutionDates.spec.ts
```diff
@@ -307,7 +307,7 @@ describe('Resolution Dates component - edit mode', () => {
   it('displays the Correct button for correction filings', () => {
     store.stateModel.tombstone.entityType = CorpTypeCd.BENEFIT_COMPANY
     store.stateModel.tombstone.filingType = FilingTypes.CORRECTION
-    store.stateModel.shareStructureStep.shareClasses = shareClasses
+    store.stateModel.shareStructureStep.shareClasses = shareClasses as any
```
seeker25, line 310: If I didn't have these in, would complain about missing properties

tests/unit/App.spec.ts
severinbeauvais: I changed my mind. What happens if you change the declarations to `| any` instead? (in src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts)

src/components/Conversion/ConversionNOB.vue
seeker25: Might need some changes to get Volar working similar to AUTH-WEB 

https://github.com/bcgov/sbc-auth/blob/main/auth-web/volar.config.js

https://github.com/bcgov/sbc-auth/blob/f497bdee0ecb899645b1113f318f0bb25c4f9340/auth-web/package.json#L61

tests/unit/Change.spec.ts
severinbeauvais: If we were going to do this, then we should check all enums. But I'm not going to worry about this for now.

tests/unit/ChangeBusinessType.spec.ts
severinbeauvais: Previously, the unit tests would let you get away with untyped objects, etc, so I didn't want to mix types and non-types (in order to make if fragile).

I agree that full typing makes it more fragile, so let's go that route. Change what you're willing to change (thanks in advance) but I realize this is OOS for this PR.

tests/unit/Actions.spec.ts
severinbeauvais: OK, that's new and nice 👍

severinbeauvais:
You have a commit named "Fix for strict build." ... how did you strict build this? It would be nice to have this enabled from now on. (Is it a linter option?)

seeker25:
The CI runs a build-check that is "strict"

src/components/Conversion/ConversionNOB.vue
severinbeauvais: Thanks, I'll try this ^^

tests/unit/CompanyProvisions.spec.ts
seeker25: We don't need the Vuex store, because none of them use them anymore. They all use the Pinia store which is loaded in via `setActivePinia(createPinia())`

tests/unit/App.spec.ts
severinbeauvais: What happens if you remove `| object` from the 3 interfaces that use it? I tried locally and don't see any errors.

src/store/store.ts
severinbeauvais: Thanks. I see you got the last word though 😆

tests/unit/YourCompany.spec.ts
seeker25: Might be shared, if you want a fresh store:
https://pinia.vuejs.org/cookbook/testing.html

image

tests/unit/App.spec.ts
seeker25: Seems to work now

src/store/store.ts
seeker25: https://pinia.vuejs.org/cookbook/options-api.html

seeker25: Doesn't look as explicit as `@Action` though

tests/unit/Actions.spec.ts
seeker25: https://pinia.vuejs.org/core-concepts/actions.html

tests/unit/ResolutionDates.spec.ts
severinbeauvais: Beauty.

src/components/Restoration/YourCompanySummary.vue
seeker25: Not sure about mapWritableState yet..

src/components/common/YourCompany/NameTranslations/ListNameTranslation.vue
seeker25: ok sounds good

tests/unit/YourCompany.spec.ts
severinbeauvais: Good to know, thanks.

src/store/store.ts
severinbeauvais: 1. It remains to be seen whether `mapStores` passes along the typings.

2. I hope we never use `mapWritableState` and always use actions instead. I'm thinking the actual store state should always use accessors as a best practise.

jonathan-longe (approved):

tests/unit/Actions.spec.ts
severinbeauvais: The Pinia page doesn't have an example of using their async action. I assume it needs to be awaited.

src/components/Restoration/YourCompanySummary.vue
severinbeauvais: Well, you replaced all `@Mutation`, which is as close to writing directly to the state as anything, with actions that write directly to the state. I think this is fine -- we still go through store methods to change its state.

What I don't want to see is this (pseudo-code):
```
@State myVariable: string

this.myVariable = "hi there"
```

PR #473, opened 2023-04-10
15882 wire up approval type for restorations

jonathan-longe:
*Issue #:* /bcgov/entity#15882

Bug fixes related to Conversion to Full Restoration.  Specifically:

#### Relationship
- [x] Upgrade relationship shared component to fix left margin
- [x] Relationship component not saving to draft
- [x] Relationship component is not validating (use must select at least one checkbox)

#### Approval type
- [x] Approval type component should show on the "Review and Cerifty" page (step 2)
- [x] Approval type values should be saved and restored using the "Save and Resume Later" button
- [x] Approval type is not validated prior to showing the "Review and Cerifty" page (step 2)
- [x] Approval type should display its validation status when user attempts to navigate to step 2


src/mixins/filing-template-mixin.ts
```diff
@@ -824,7 +824,7 @@ export default class FilingTemplateMixin extends DateMixin {
     })
 
     // restore Restoration data
-    this.mutateRestorationApprovalType(this.getStateFilingRestoration?.approvalType)
+    this.mutateRestorationApprovalType(filing.restoration.approvalType)
```
JazzarKarim, line 827: I don't think this'll work! This is since when we create a new filing, it only has a type. We were getting the approval type from state filing on load to know whether we should show the approval type section or not. Also, the approval type is always going to be the same as the state filing in this case. 

For reference, this is what we get when creating a brand new limited to full filing:
for reference

jonathan-longe: Hey @JazzarKarim thanks for spotting this.  I'll make a change so this works.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "4.1.1",
+  "version": "4.1.2",
```
severinbeauvais, line 3: Check with Travis whether you or he should go first. Neither is going to Test today or Prod tomorrow so it's entirely up to you guys.

src/components/Restoration/ExtendTimeLimit.vue
```diff
@@ -92,7 +93,6 @@ export default class ExtendTimeLimit extends Vue {
     return this.getStateFilingRestoration?.approvalType
   }
 
-  /** The court order draft file number. */
   get courtOrderNumberText (): string {
```
severinbeauvais, line 96: If this method is now obsolete, might as well delete it.

src/interfaces/store-interfaces/state-interfaces/flags-company-info-interface.ts
```diff
@@ -21,4 +21,5 @@ export interface FlagsCompanyInfoIF {
   isValidCreateSpecialResolution: boolean
   isValidApprovalType: boolean
   isValidExtensionTime: boolean
+  isValidRelationship?: boolean
```
severinbeauvais, line 24: The other ones are required and True by default. Why is this one different?

src/store/getters.ts
```diff
@@ -1081,6 +1082,18 @@ export default {
     return state.stateModel.stateFilingRestoration
   },
 
+  getRelationships (state: StateIF): RelationshipTypes[] {
```
severinbeauvais, line 1085: I think this getter is too granular, since relationships only apply to restoration filings.

AFAIK, a computed method (local getter) is also cached and reactive, so having the getter here doesn't buy us anything.

Thoughts?

src/store/getters.ts
```diff
@@ -1081,6 +1082,18 @@ export default {
...
+    return state.stateModel.restoration.relationships
+  },
+
+  getRelationshipValid (state: StateIF): boolean {
+    return state.stateModel.validationFlags.flagsCompanyInfo.isValidRelationship
+  },
+
+  getIsRestorationTypeCourtOrder (state: StateIF): boolean {
+    return state.stateModel.restoration.courtOrder?.fileNumber !== null
```
severinbeauvais, line 1094: Same comment as above.

Also, is it possible the file number is undefined or empty string? (What is the value if you enter court order and then change the radio button to unset it?

src/store/getters.ts
```diff
@@ -1166,5 +1179,10 @@ export default {
     }
 
     return 'Edits Saved' // should never happen
+  },
+
+  /** The court order draft file number. */
+  getCourtOrderNumberText (_state, getters): string {
+    return getters.getRestoration.courtOrder?.fileNumber || ''
```
severinbeauvais, line 1186: Why return empty string? Don't we always check if "falsy"? For example, `getCourtOrderNumberTest || 'Unknown'` ?

src/store/state/state-model.ts
```diff
@@ -152,6 +153,9 @@ export const stateModel: StateModelIF = {
     approvalType: null,
     approvalTypeValid: true,
     businessNameValid: false,
+    courtOrder: {
+      fileNumber: null
+    },
```
severinbeauvais, line 158: Please verify that there is conflict between this court order and the one in the newAlteration store object.

src/views/LimitedRestorationToFull.vue
```diff
@@ -180,11 +192,19 @@ export default class LimitedRestorationToFull extends Vue {
...
+
+  @Mutation mutateRestorationCourtOrder!: ActionBindingIF
 
   /** Whether App is ready. */
   @Prop({ default: false }) readonly appReady!: boolean
   @Prop({ default: 0 }) readonly restorationId!: number
 
+  /* isDataLoaded is a flag that is to "true" after the component's data loaded
+  it's used a work-around because the shared components aren't reactive to
+  changes after the shared components are mounted. */
```
severinbeauvais, line 205: If you use JSDoc syntax here then hovering over the property gives you its comment :)

severinbeauvais: Is there a ticket to fix this later?

jonathan-longe:
Will respond to comments after rebasing

src/views/LimitedRestorationToFull.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <ViewWrapper>
+  <ViewWrapper v-if="isDataLoaded">
```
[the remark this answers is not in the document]
jonathan-longe: Yes, I added the code comment where `isDataLoaded` is declared.

src/store/getters.ts
jonathan-longe: If we're going to follow best practices and not repeat ourselves then any method that is repeated in more than one place needs to be moved to location that's common.  In this particular case `getRelationships()` is required in both the YourCompanySummary and the LimitedRestorationtoFull so it makes sense to move it to the store.  We could, if you prefer, move it somewhere else that's common (like a mixin).  In general, I think it business logic should not be located in components.  It makes it difficult to reuse and hence more likely to be copied and duplicated elsewhere.

src/store/getters.ts
jonathan-longe: This method and identical logic was duplicated in two places previously.  I'm simply copying it to the store without changing the logic.

src/views/LimitedRestorationToFull.vue
jonathan-longe: Good point, I'll change to use the JSDoc syntax.

src/store/getters.ts
severinbeauvais: My first thought was that I'd like to reduce our use of mixins and save them for non-store data.

My second thought is that it might be nice to have "something" specific to restorations. Surely not a store module. Maybe a mixin? I dunno.

Given that this getter is used in a couple of places, I'm OK with the current implementation.

src/store/getters.ts
severinbeauvais: Oh, groan, OK.

src/views/LimitedRestorationToFull.vue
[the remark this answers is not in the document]
jonathan-longe: Okay to resolve?

package.json
jonathan-longe: Okay to resolve?

src/components/Restoration/ExtendTimeLimit.vue
jonathan-longe: Removed

src/views/LimitedRestorationToFull.vue
jonathan-longe: Okay to resolve?

package.json
```diff
@@ -1,5 +1,4 @@
 {
-  "name": "business-edit-ui",
```
severinbeauvais, line 2: Wait, what?

package.json
severinbeauvais: Almost. Please update the app version.

src/views/LimitedRestorationToFull.vue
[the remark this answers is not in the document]
severinbeauvais: I see the comment so I'm satisfied. Ketaki, please resolve this conversation if OK with you.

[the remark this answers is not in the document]
ketaki-deodhar: I am good. Resolving this.

src/store/getters.ts
jonathan-longe: This getter method has been moved to `store.ts` as part of the Pinia upgrade.  Like `getRelationships()` (see comment above) I think it's important that our business logic not reside in components and not be duplicated.

src/views/LimitedRestorationToFull.vue
jonathan-longe: Weird, the comment has been changed to a JSDoc comment, but Github does not show the change.

package.json
jonathan-longe: Wow, good catch, I didn't remove that line consciously.

package.json
jonathan-longe: Done

src/mixins/filing-template-mixin.ts
jonathan-longe: Hey @severinbeauvais I think we made this change together.  I'd like to discuss the issue Karim has raised above to make sure I understand the implications.

package.json
jonathan-longe: Done

src/interfaces/store-interfaces/state-interfaces/flags-company-info-interface.ts
jonathan-longe: Ah, well spotted, I'll fix it.

src/store/state/state-model.ts
jonathan-longe: Verified, no conflict.  We're using different getters and setters for the filings.

src/mixins/filing-template-mixin.ts
JazzarKarim: Unless I'm mistaken, the way it was should be fine which is setting approval type the same as that of state filing. Please correct me if I'm wrong.
CC: @severinbeauvais

jonathan-longe: Fixed by wrapping the action in a conditional.  Now, if the `approvalType` attribute isn't contained in the restoration object, we won't set it.  Thanks @JazzarKarim for spotting that!

JazzarKarim: Awesome! Thanks Jonathan!

I just have one small question. Since we're using the approval type to determine whether we're going to show the approval type section in `ExtendTimeLimit.vue`, will your change have implications on that?

I'm thinking it might since when we create a new limited restoration extension now, the restoration object won't know if it was previously approved by court order or registrar. Correct me if I'm wrong!

JazzarKarim: Nevermind, I just noticed this: https://github.com/bcgov/business-edit-ui/blob/65f9e45b82871a2141f6486500061a3e8e535dd8/src/components/Restoration/ExtendTimeLimit.vue#L93
It should be good! Thanks Jonathan!

src/store/getters.ts
severinbeauvais: My second point here is that the comparison is against null only, but if the courtOrder object doesn't exist then the first part of the comparison will be undefined (which is not equal to null).

I'd be tempted to use `!` to check this instead.

src/views/LimitedRestorationToFull.vue
severinbeauvais: Probably the file didn't get committed.

src/mixins/filing-template-mixin.ts
```diff
@@ -829,6 +833,10 @@ export default class FilingTemplateMixin extends DateMixin {
 
     // restore Restoration data
     this.setRestorationApprovalType(this.getStateFilingRestoration?.approvalType)
+    if (filing.restoration.approvalType) {
+      this.mutateRestorationApprovalType(filing.restoration.approvalType)
```
severinbeauvais, line 837: Lines 835 and 837 look like duplicates (sort of).

Now the question is, which approval type needs to be restored here? Probably the second one if it exists, else the first one.

src/views/LimitedRestorationToFull.vue
jonathan-longe: > Probably the file didn't get committed.

On the PR's "conversation" tab the comment shows as not changed, but on the "files" tab, the changes are shown.  I promise I didn't forget to commit the file!  🤞

src/store/getters.ts
jonathan-longe: I like that!  Fixed

src/mixins/filing-template-mixin.ts
jonathan-longe: Good call.  I've changed the logic as per your suggestion

JazzarKarim (approved):
Looks good! 👍

ketaki-deodhar (approved):

src/mixins/filing-template-mixin.ts
severinbeauvais: Thanks.

Is there any possiblity that line 838 is setting `undefined` into the store?

src/views/LimitedRestorationToFull.vue
severinbeauvais: 🤷‍♂️

tests/unit/ExtendTimeLimit.spec.ts
```diff
@@ -1,4 +1,4 @@
-import Vue from 'vue'
+import Vue, { nextTick } from 'vue'
```
severinbeauvais, line 1: Looks like `nextTick` isn't being used.

tests/unit/FetchErrorDialog.spec.ts
```diff
@@ -26,12 +25,12 @@ describe.skip('Fetch Error Dialog', () => {
...
     expect(wrapper.attributes('contentclass')).toBe('fetch-error-dialog')
     expect(wrapper.isVisible()).toBe(true)
-    expect(wrapper.find('#dialog-title').text()).toBe('Unable to Resume Application')
+    expect(wrapper.find('#dialog-title').text()).toBe('Unable to Retrieve Filing')
     expect(wrapper.findAll('p').length).toBe(1)
-    expect(wrapper.findAll('p').at(0).text()).toContain('We were unable to resume your')
-    expect(wrapper.find(ErrorContact).exists()).toBe(false)
-    expect(wrapper.find('#dialog-exit-button').exists()).toBe(true)
-    expect(wrapper.find('#dialog-retry-button').exists()).toBe(true)
+    expect(wrapper.findAll('p').at(0).text()).toContain('We were unable to retrieve your')
```
severinbeauvais, line 30: Is this a change requested by the UX team?

tests/unit/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "unit",
+  "version": "1.0.0",
+  "dependencies": {
+  }
+}
```
severinbeauvais, line 6: What is this? Should we have it in the other projects?

tests/unit/LimitedRestorationToFull.spec.ts
```diff
@@ -244,6 +244,7 @@ describe('LimitedRestorationToFull component - edit page', () => {
 
     // enable filing and wait for all queries to complete
     await wrapper.setProps({ appReady: true })
+    wrapper.vm.$data.isDataLoaded = true
```
severinbeauvais, line 247: I'm surprised you need to wait for anything to finish after setting `appReady` -- since appReady is the trigger to (async) load a couple of filings.

src/store/store.ts
```diff
@@ -1195,7 +1195,7 @@ export const useStore = defineStore('store', {
     },
 
     getIsRestorationTypeCourtOrder (): boolean {
-      return this.stateModel.restoration.courtOrder?.fileNumber !== null
+      return !!this.stateModel.restoration.courtOrder?.fileNumber
```
severinbeauvais, line 1198: Should this be a single `!`?

src/store/store.ts
```diff
@@ -1180,11 +1180,24 @@ export const useStore = defineStore('store', {
...
+    getCourtOrderNumberText (): string {
+      return this.stateModel.restoration.courtOrder?.fileNumber || ''
+    },
+
+    getRelationships (): RelationshipTypes[] {
+      return this.stateModel.restoration.relationships
+    },
+
+    getIsRestorationTypeCourtOrder (): boolean {
+      return !!this.stateModel.restoration.courtOrder?.fileNumber
```
severinbeauvais, line 1198: Should this be a single `!`?

tests/unit/FetchErrorDialog.spec.ts
```diff
@@ -15,7 +14,7 @@ const store = useStore()
 // Prevent the warning "[Vuetify] Unable to locate target [data-app]"
 document.body.setAttribute('data-app', 'true')
 
-describe.skip('Fetch Error Dialog', () => {
+describe('Fetch Error Dialog', () => {
```
jonathan-longe, line 17: These tests were being skipped.  I've re-enabled the tests and then changed the error message to match current behavior.  If you prefer, I can skip the test again.

tests/unit/FetchErrorDialog.spec.ts
jonathan-longe: I'm changing the tests to match current behavior.  I think this is better than having the tests skipped.

src/store/store.ts
jonathan-longe: I believe a single `!` reverses a boolean value.  A double `!!` casts any type to a boolean.  Since `stateModel.restoration.courtOrder?.fileNumber` could return a string, I think a double exclamation mark is required.

tests/unit/ExtendTimeLimit.spec.ts
jonathan-longe: Good catch, I'll remove it.

tests/unit/FetchErrorDialog.spec.ts
severinbeauvais: IMHO, if the tests pass, leave them. If they fail, fix them or skip them again.

tests/unit/FetchErrorDialog.spec.ts
severinbeauvais: Agreed. LGTM

src/store/store.ts
severinbeauvais: OK, I'm the one who's tired. This is well and good. Resolving.

(PS Another way to do this is `Boolean(fileNumber)`.)

src/mixins/filing-template-mixin.ts
jonathan-longe: I haven't changed the logic, `this.setRestorationApprovalType(this.getStateFilingRestoration?.approvalType)` is exactly as written previously.

severinbeauvais: Yes, but I think it was a bug. Karim, thoughts?

JazzarKarim: `this.setRestorationApprovalType(this.getStateFilingRestoration?.approvalType)` should work just fine! I've tested it and it'll never return `undefined`.

tests/unit/package.json
jonathan-longe: No, good call, it's not necessary.  I was trying something new, but this file shouldn't have been committed.

src/mixins/filing-template-mixin.ts
jonathan-longe: Thanks @JazzarKarim and @severinbeauvais .  I'll leave this as is unless either of you object.

tests/unit/FetchErrorDialog.spec.ts
jonathan-longe: Thanks for the clarification.  I've enabled the tests and made sure they're passing.  In retrospect, I should have left the tests skipped, but now that they're passing I'd like to leave them as is.

src/mixins/filing-template-mixin.ts
severinbeauvais: OK to leave it as is. (That conditional chaining is bugging me though, lol.)

tests/unit/LimitedRestorationToFull.spec.ts
severinbeauvais: I guess this works, so nothing to change here.

severinbeauvais (approved):
GTG

severinbeauvais:
No wait, another test failure.

You probably need to do
```
store.stateModel.restoration = { courtOrder: {} }
```

jonathan-longe:
> No wait, another test failure.

Yup, I'm on it ... fix in the works

tests/unit/ExtendTimeLimit.spec.ts
```diff
@@ -91,10 +91,6 @@ describe('Time Limit Extension component', () => {
   })
 
   it('get correct court order file number.', () => {
-    const wrapper = wrapperFactory()
-    const extendTimeLimit = wrapper.vm as any // wrapper.vm type is Vue
-
-    expect(extendTimeLimit.courtOrderNumberText).toEqual('')
-    wrapper.destroy()
+    expect(store.getCourtOrderNumberText).toEqual('')
```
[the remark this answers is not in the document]
jonathan-longe: Hmm, I could be wrong, but my feeling is that since it's a store getter there's no need to mount any component.  Is there a benefit to mounting a component when testing store getters?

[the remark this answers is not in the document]
severinbeauvais: It's possible that mounting a component sets some store state.

PR #494, opened 2023-04-28
Fix dev tools - Update BreadCrumbResources.ts

seeker25:
Fix dev tools working intermittently - sometimes it would work when you first load the page, but not when you have the page already open and go into the Vue tools.

Fixes this error:

image

The Pinia store is being used before the Vue instance is initialized I believe

### DIFF
--- a/src/components/SpecialResolution/CreateSpecialResolution.vue
+++ b/src/components/SpecialResolution/CreateSpecialResolution.vue
@@ -62,7 +62,6 @@
           >
             <div class="d-flex flex-column flex-sm-row justify-center align-center">
               <img
-                slot-scope=""
                 src="@/assets/images/BCRegistries_CoopSpecialResolution-x2.png"
                 :alt="getSpecialResolutionResource.label"
                 class="preview-image"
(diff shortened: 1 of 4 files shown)


src/resources/BreadCrumbResources.ts
```diff
@@ -1,17 +1,14 @@
...
-const store = useStore()
-
 /** Returns URL param string with Account ID if present, else empty string. */
 function getParams (): string {
   const accountId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id as string
   return accountId ? `?accountid=${accountId}` : ''
 }
 
 export function getEntityDashboardBreadcrumb (): BreadcrumbIF {
+  const store = useStore()
```
seeker25, line 11: Fairly certain we can assume the store has been initialized, because App.vue calls this.

seeker25:
Not sure if you want the package + package-lock.json updated for this.. fairly small change.

src/components/SpecialResolution/CreateSpecialResolution.vue
```diff
@@ -62,7 +62,6 @@
           >
             <div class="d-flex flex-column flex-sm-row justify-center align-center">
               <img
-                slot-scope=""
```
seeker25, line 65: This was giving us a warning

severinbeauvais: This was auto added by eslint. Looks like it wasn't doing much anyway :)

severinbeauvais (approved):
LGTM.

seeker25:
Should be fine, I didn't see any runtime issues - just dev tools.

PR #507, opened 2023-06-14
13550 verify entity type for sp/gp change and conversion filing

bennypc:
*Issue #:* /bcgov/entity#13550

*Description of changes:*
Created a switch statement to check for the entity type within Change and Conversion UI, checks if entity is a benefit company


JazzarKarim (commented):
Can you post a screenshot of the console when you try to do a change for BEN for example?

src/views/Change.vue
```diff
@@ -202,6 +203,13 @@ export default class Change extends Mixins(CommonMixin, FeeMixin, FilingTemplate
       // fetch entity snapshot
       const entitySnapshot = await this.fetchEntitySnapshot()
 
+      switch (entitySnapshot.businessInfo.legalType) {
+        case CorpTypeCd.BENEFIT_COMPANY:
+        case CorpTypeCd.BC_COMPANY:
+        case CorpTypeCd.BC_CCC:
+        case CorpTypeCd.BC_ULC_COMPANY:
+          throw new Error(`Invalid entity type, must not be benefit company id`)
+      }
```
severinbeauvais, line 212: This looks like a good place to check this 👍 

However, this could fail if we add new entity types and forget to check this code. (Why would we? This is for a firm change filing.)

1. I think it would be better to throw the error if the current legal type is NOT a firm (SP or GP).
2. Can `entitySnapshot` be null? Can `entitySnapshow.businessInfo` be null? If so then you can use optional chaining (*) to prevent a de-reference error.

(*) We should use optional chaining sparingly, since it can mask errors that we might want to otherwise see or handle.

severinbeauvais (commented):
I think you should also add checks to the other filings. For example:
- Special Resolutions are for coops only
- Alterations are for corps (BC/BEN/CCC/ULC) only
- Limited Restoration Extensions are for corps only
- Limited Restoration to Full are for corps only
- Corp Corrections are for corps only
- Firm Corrections are for firms (SP/GP) only

There will soon be a Special Resolution Correction (or maybe Coop Correction) for coops only -- check with Jia Xu and Travis Semple in RocketChat to see if they add add the entity type check to their code.

severinbeauvais:
Re:
image

Codecov complains when the coverage "goes down". I put that in quotes because the coverage can "go down" even if you make no code changes, so something's not right there. I think we're fine as long as code coverage doesn't drop a lot or go below 80%.

bennypc:
> Can you post a screenshot of the console when you try to do a change for BEN for example?

Screenshot 2023-06-15 at 2 06 39 PM

severinbeauvais:
> Screenshot 2023-06-15 at 2 06 39 PM

Is there an error dialog as well?

Also you should get one of our UX people (Ethan, Yui or Janis) to approve the error message.

src/views/Change.vue
bennypc: i have implemented both of these changes in the new commit

src/views/Change.vue
```diff
@@ -203,13 +203,14 @@ export default class Change extends Mixins(CommonMixin, FeeMixin, FilingTemplate
...
-        case CorpTypeCd.BC_COMPANY:
-        case CorpTypeCd.BC_CCC:
-        case CorpTypeCd.BC_ULC_COMPANY:
-          throw new Error(`Invalid entity type, must not be benefit company id`)
+      switch (entitySnapshot?.businessInfo?.legalType) {
+        case CorpTypeCd.SOLE_PROP:
+        case CorpTypeCd.PARTNERSHIP:
+          break // acceptable types
+        default:
+          throw new Error(`Invalid entity type, must be a firm (Sole Prop or General Partnership)`)
```
severinbeauvais, line 211: Can you show a screenshot of this message being displayed please?

bennypc: image

src/views/Change.vue
severinbeauvais: PS Let reviewers resolve the conversations they started -- it's a "bookmark" for them to review their comments. Otherwise they have to re-review the entire code and try to remember what they commented on and see if the code is fixed.

src/views/Change.vue
severinbeauvais: LGTM

tests/unit/CorrectNameOptions.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+        propsData: {
+          ...propsData
+        },
+        vuetify
+      })
+    }
+  })
+
+  it('renders the appropriate choices for a numbered company', async () => {
+    correctionNameChoices = ['correct-new-nr']
```
severinbeauvais, line 28: Might as well use the enum here and below.

src/enums/nameChangeOptions.ts
```diff
@@ -0,0 +1,5 @@
+export enum NameChangeOptions {
+  CORRECT_NAME = 'correct-name',
+  CORRECT_NAME_TO_NUMBER = 'correct-name-to-number',
+  CORRECT_NEW_NR = 'correct-new-nr'
+}
```
severinbeauvais, line 5: Don't add this file -- it's already imported from the shared components repo. See the index file in this folder.

tests/unit/CorrectNameOptions.spec.ts
```diff
@@ -0,0 +1,59 @@
+import Vue from 'vue'
```
severinbeauvais, line 1: Why did you add this test suite? It's not related to the safety checks you implemented...?

bennypc: Wait I added it? I don't think I added any test suites

bennypc: Should I delete it then?

tests/unit/CorrectNameOptions.spec.ts
severinbeauvais: Obsolete comment.

tests/unit/CorrectNameOptions.spec.ts
severinbeauvais: According to GitHub, yes, it's a new file in your commit:

image

If you didn't mean to do this then please delete it in a new commit, and it will disappear from the "total" commits here.

bennypc:
> I think you should also add checks to the other filings. For example:
> 
> * Special Resolutions are for coops only
> * Alterations are for corps (BC/BEN/CCC/ULC) only
> * Limited Restoration Extensions are for corps only
> * Limited Restoration to Full are for corps only
> * Corp Corrections are for corps only
> * Firm Corrections are for firms (SP/GP) only
> 
> There will soon be a Special Resolution Correction (or maybe Coop Correction) for coops only -- check with Jia Xu and Travis Semple in RocketChat to see if they can add the entity type check to their code.

It seems like the other filings already have a way to check this (except for Alterations). For example in Limited Restoration Extension theres

`get restorationResource (): ResourceIF {
    switch (true) {
      case this.isBcCompany: return BcRestorationResource
      case this.isBenefitCompany: return BenRestorationResource
      case this.isBcCcc: return CccRestorationResource
      case this.isBcUlcCompany: return UlcRestorationResource
    }
    return null // should never happen
  }`

severinbeauvais (approved):
LGTM, but please wait another review or 2 before merging. Thx.

chenhongjing (approved):
LGTM 😺

JazzarKarim (approved):
LGTM 👍

PR #516, opened 2023-07-20
16572 Alteration - Edit UI change for business type alteration

Jxio:
*Issue #:* /bcgov/entity[16572](https://github.com/bcgov/entity/issues/16572)

*Description of changes:*
1. Enable "change" button for BC, BEN, ULC (BC already enabled)
2. Update type dropdown options
3. Suffix change
4. Done button pressed screen
UX Design: https://preview.uxpin.com/c21a28a1679da5e2b71044251e8296995e00d9c0#/pages/162390910/simulate/sitemap


src/components/common/BcRegComments.vue
```diff
@@ -0,0 +1,311 @@
...
+  @Prop({ default: false }) readonly confirmArticles!: boolean
+  @Prop({ default: false }) readonly isBenefit!: boolean
+  @Prop({ default: false }) readonly isUnlimitedLiability!: boolean
+  @Prop({ default: false }) readonly isCommunityContribution!: boolean
+  @Prop({ default: false }) readonly isBcLimited!: boolean
+  @Prop({ default: '' }) readonly selectedEntityType: CorpTypeCd
+
+  protected helpToggle = false
+
+  data () {
```
seeker25, line 255: Can't you just create a property localConfirmArticles like like 253?

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -324,21 +258,24 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
   @Getter(useStore) isConflictingLegalType!: boolean
+  @Getter(useStore) isNumberedCompany!: boolean
 
   @Action(useStore) setEntityType!: ActionBindingIF
+  @Action(useStore) setNameRequest!: ActionBindingIF
+  @Action(useStore) setNameChangedByType!: ActionBindingIF
 
-  protected selectedEntityType = null as CorpTypeCd
+  selectedEntityType = null as CorpTypeCd
   protected confirmArticles = false
```
seeker25, line 274: Don't need `protected`, please remove for these lines

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -397,6 +392,64 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   protected submitTypeChange () {
     this.setEntityType(this.selectedEntityType)
     this.isEditingType = false
+
+    if (this.shouldUpdateName()) {
+      const originalName = this.getEntitySnapshot?.businessInfo.legalName
```
seeker25, line 397: 397 -> 400 looks very similar to what we have in the store, might be a candidate for moving over to there.

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -397,6 +392,64 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
+    }
+  }
+
+  private shouldUpdateName (): boolean {
+    return this.isNumberedCompany && !this.hasBusinessNameChanged
+  }
+
+  private getUpdatedName (originalName: string): string {
+    if (this.isUnlimitedLiability || this.isCommunityContribution) {
+      return originalName.endsWith(' LTD.') ? originalName : originalName + ' LTD.'
```
seeker25, line 419: Are there more patterns to catch than just LTD?

src/components/common/YourCompany/EntityName.vue
```diff
@@ -44,7 +44,7 @@
           <!-- Business Type Info -->
           <template
             v-if="!hasNewNr && hasBusinessNameChanged && (isAlterationFiling || isFirmChangeFiling ||
```
seeker25, line 46: This is already too long, I'd recommend converting to a computed

src/components/common/YourCompany/EntityName.vue
```diff
@@ -111,8 +133,8 @@
           <div class="actions mr-4">
             <!-- FUTURE: only show buttons for named company -->
             <v-btn
-              v-if="hasCompanyNameChanged || (hasBusinessNameChanged && (isAlterationFiling ||
-                isFirmChangeFiling || isSpecialResolutionFiling))"
+              v-if="(hasCompanyNameChanged || (hasBusinessNameChanged && (isAlterationFiling ||
```
seeker25, line 136: Already too long, should be converted into computed

src/components/common/YourCompany/EntityName.vue
```diff
@@ -137,8 +159,8 @@
               <span>{{ getEditLabel }}</span>
             </v-btn>
             <span
-              v-if="hasCompanyNameChanged || (hasBusinessNameChanged &&
-                (isAlterationFiling || isFirmChangeFiling || isSpecialResolutionFiling))"
+              v-if="(hasCompanyNameChanged || (hasBusinessNameChanged &&
```
seeker25, line 162: Already too long, should be converted into computed

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -376,19 +356,34 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
-    return (this.selectedEntityType === CorpTypeCd.BENEFIT_COMPANY)
-  }
-
   /** Type change helper information */
   get typeChangeInfo (): string {
     return this.getResource.changeData?.typeChangeInfo
   }
 
+  /** Entity type options based on the company type */
+  get entityTypeOptions (): Array<{ value: string, SHORT_DESC: string, text: string }> {
```
seeker25, line 365: Create a concrete type for a return?

src/enums/entityTypeInfo.ts
```diff
@@ -0,0 +1,25 @@
+import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module'
+
+// Define a map of entity types to titles and info strings
+export const entityTypeInfo = {
```
seeker25, line 4: Not sure if it's worth moving these to the already existing resource files?

src/components/common/BcRegComments.vue
```diff
@@ -0,0 +1,311 @@
+<template>
+  <div>
+    <template v-if="isBenefit">
```
seeker25, line 3: I think this is fine not being in the resource file, it's fairly dynamic and requires quite a bit of html

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -376,19 +356,34 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
+    if (this.isBcCompany) {
```
seeker25, line 366: This is already done in the alteration view, to choose the correct resource. If you moved the business types to the resources, you wouldn't need to do this.

src/components/common/BcRegComments.vue
```diff
@@ -0,0 +1,311 @@
...
+  onLocalConfirmArticlesChanged (newValue: boolean) {
+    this.$emit('update:confirmArticles', newValue)
+  }
+
+  get confirmLabel (): string {
+    return entityTypeInfo[this.selectedEntityType]?.info || ''
+  }
+
+  get updatedArticleTitle (): string {
+    return entityTypeInfo[this.selectedEntityType]?.title || ''
```
seeker25, line 278: Grab from resource file

src/components/common/YourCompany/ChangeBusinessType.vue
seeker25: EG. `this.getResource.entityTypeOptions`

tests/unit/BcRegComments.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      isCommunityContribution: true,
+      selectedEntityType: 'CC'
+    }, 'Community Contribution Company Articles'],
+    [{
+      confirmArticles: false,
+      isBcLimited: true,
+      selectedEntityType: 'BC'
+    }, 'Limited Company Articles']
+  ]
+  test.each(typeCommentTests)(
```
seeker25, line 57: Cool - didn't know you could do this

src/components/common/BcRegComments.vue
Jxio: ok, will do, thanks!

src/components/common/YourCompany/ChangeBusinessType.vue
Jxio: just LTD, I confirmed with @ethantspitt

src/enums/entityTypeInfo.ts
Jxio: good catch! I'll update.

src/components/common/YourCompany/ChangeBusinessType.vue
Jxio: It's originally protected there before, I can just remove it, but can't use readonly cause property's value needs to be changed during lifecycle.

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -397,6 +392,64 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
+          nrNumber: this.getEntitySnapshot?.businessInfo?.nrNumber
+        }
+
+        this.setNameRequest(nameRequest)
+        this.setNameChangedByType(true)
+      }
+    }
+  }
+
+  private shouldUpdateName (): boolean {
```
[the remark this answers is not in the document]
Jxio: it's good to learn, thanks!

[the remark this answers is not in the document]
seeker25: I think it makes sense when you're using Typescript still, outside of the Vue context.. like building services etc. 

But not inside the Vue components themselves - I don't think we do much inheritance, we mostly do composition

pwei1018:
Temporary Url for review: https://business-edit-dev--pr-516-37q7zf1c.web.app

src/resources/Alteration/BC.ts
```diff
@@ -16,7 +16,32 @@ export const BcAlterationResource: ResourceIF = {
     correctNameOptions: [
       CorrectNameOptions.CORRECT_NEW_NR,
       CorrectNameOptions.CORRECT_NAME_TO_NUMBER
-    ]
+    ],
+    entityTypeOptions: [
```
seeker25, line 20: Awesome

seeker25:
As discussed, we need to move the mixin to a util hopefully.

pwei1018:
Temporary Url for review: https://business-edit-dev--pr-516-37q7zf1c.web.app

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "4.4.9",
+  "version": "4.5.0",
```
severinbeauvais, line 3: Yes, this is a significant new feature, so I agree with increasing the minor release number.

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -4,7 +4,10 @@
       <!-- Row Title -->
       <v-col cols="3">
         <label :class="{'error-text': invalidSection}"><strong>Business Type</strong></label>
-        <v-col md="1">
+        <v-col
+          md="1"
+          class="pa-0"
+        >
```
severinbeauvais, line 10: Check this on a narrow viewport (screen). Generally, columns should stack without weird whitespace.

(Do you have screenshots of normal vs narrow?)

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -242,7 +174,7 @@
             <span>Undo</span>
           </v-btn>
           <v-btn
-            v-else-if="isBcCompany"
+            v-else-if="isBcCompany || isBcUlcCompany || isBenefitCompany"
```
severinbeauvais, line 177: What about CCC?

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -397,6 +326,64 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
+  get isUnlimitedLiability (): boolean {
+    return (this.selectedEntityType === CorpTypeCd.BC_ULC_COMPANY)
+  }
+
+  /** Check if current entity selection is a Community Contribution Company */
+  get isCommunityContribution (): boolean {
+    return (this.selectedEntityType === CorpTypeCd.BC_CCC)
+  }
+
+  /** Check if current entity selection is a Benefit Company */
```
severinbeauvais, line 376: Comment.

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -409,47 +396,17 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
 
-.v-icon {
+:deep(.v-icon) {
```
severinbeauvais, line 399: Add comment indicating what you're doing (or why). This will help the next developer to look at this, and may help the upcoming upgrade from Vuetify 2 to 3.

src/components/common/YourCompany/EntityName.vue
```diff
@@ -266,13 +285,27 @@ export default class EntityName extends Mixins(NameRequestMixin) {
...
+            (this.hasBusinessNameChanged &&
+             (this.isAlterationFiling || this.isFirmChangeFiling || this.isSpecialResolutionFiling))) &&
+            !this.isNameChangedByType
+  }
+
+  get shouldShowTypeDetail () {
+    return !this.hasNewNr && this.hasBusinessNameChanged &&
+           (this.isAlterationFiling || this.isFirmChangeFiling || this.isFirmConversionFiling) &&
+           !this.isNameChangedByType
+  }
```
severinbeauvais, line 307: I like these 2 getters instead of having the code in the template.

Could you add a comment to each to describe when then are True?

(Also add return types.)

src/components/common/YourCompany/ChangeBusinessType.vue
Jxio: (changeFrom) CCC not included on UX.

src/resources/Alteration/BC.ts
severinbeauvais: Agreed! Nicely done.

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -356,39 +292,32 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
-  }
-
   /** Type change helper information */
   get typeChangeInfo (): string {
     return this.getResource.changeData?.typeChangeInfo
   }
 
+  /** Entity type options based on the company type */
+  get entityTypeOptions (): EntityTypeOption[] {
+    return this.getResource.changeData?.entityTypeOptions
```
severinbeauvais, line 308: If changeData is falsy then this returns an undefined.

Should this return an empty array if changeData is falsy? Or can the code that uses this handle an undefined value?

src/interfaces/alteration-interfaces/alteration-interface.ts
```diff
@@ -33,3 +33,9 @@ export interface AlterationIF extends CoopAlterationIF {
   contactPoint: ContactPointIF
   courtOrder?: CourtOrderIF
 }
+
+export interface EntityTypeOption {
+  value: string;
+  SHORT_DESC: string;
+  text: string;
+}
```
severinbeauvais, line 41: As this isn't an "alteration interface", I recommend putting it in its own file, maybe.

Also, why "SHORT_DESC"? Could this follow our usual camel case convention?

src/store/store.ts
```diff
@@ -1386,6 +1391,9 @@ export const useStore = defineStore('store', {
       this.stateModel.nameRequest = nameRequest
       if (!this.stateModel.tombstone.ignoreChanges) this.stateModel.tombstone.haveUnsavedChanges = true
     },
+    setNameChangedByType (changedByType: boolean) {
+      if (changedByType) this.stateModel.tombstone.nameChangedByType = changedByType
```
severinbeauvais, line 1395: Why the "if"? Is there no way to set this to False?

src/utils/resource-utils.ts
severinbeauvais, line 1: Could this fit in with the Resource Model that already exists? See store/state/ and store/store.ts.

src/components/common/BcRegEntityDetails.vue
```diff
@@ -0,0 +1,310 @@
...
+          containing a benefit provision.
+        </p>
+      </div>
+
+      <section class="text-body-3">
+        <!-- Help Section Toggle -->
+        <HelpSectionToggle
+          :isHelpVisible="helpToggle"
+          @toggle="helpToggle = !helpToggle"
+        />
```
severinbeauvais, line 20: This component ^^ doesn't do much. I think it would be a better design to have this component handle the showing and hiding of the content (ie, fully encapsulate the show/hide logic).

You could do this by passing a template with the content to display (ie, the help section below).

See also HelpSection.vue.

src/components/common/BcRegEntityDetails.vue
```diff
@@ -0,0 +1,310 @@
+<template>
+  <div>
+    <template v-if="isBenefit">
```
severinbeauvais, line 3: I'm sort of OK with having the HTML for all entity types in this one file, but did you consider having separate sub-components for each type?

You could use dynamic components here.
Example: https://github.com/bcgov/business-filings-ui/blob/fafe0c3af7da84fa91f3677f785cfe5c7f0adb8b/src/components/Dashboard/FilingHistoryList.vue#L46

You may have already tried that and your current implementation may be cleaner. If so, say so :)

src/components/common/YourCompany/ChangeBusinessType.vue
seeker25: Probably could write 

`this.getResource.changeData?.entityTypeOptions || []`

src/utils/resource-utils.ts
seeker25: It was rewritten as a mixin, I suggested putting into utils

src/components/common/BcRegComments.vue
```diff
@@ -0,0 +1,311 @@
...
+})
+export default class BcRegComments extends Vue {
+  @Prop({ default: false }) readonly confirmArticles!: boolean
+  @Prop({ default: false }) readonly isBenefit!: boolean
+  @Prop({ default: false }) readonly isUnlimitedLiability!: boolean
+  @Prop({ default: false }) readonly isCommunityContribution!: boolean
+  @Prop({ default: false }) readonly isBcLimited!: boolean
+  @Prop({ default: '' }) readonly selectedEntityType: CorpTypeCd
+
+  protected helpToggle = false
```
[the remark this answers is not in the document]
severinbeauvais: Theoretically, protected would allow the template to see this property but it wouldn't be visible outside this class/component.

However that doesn't work, so I'm OK with removing the accessor when the template needs to see this property.

src/components/common/BcRegEntityDetails.vue
seeker25: Would consider it, but this is mostly a 1 off?

src/components/common/YourCompany/EntityName.vue
severinbeauvais: When you change this to a computed (get) then you also have the opportunity to add a comment to describe the particular business logic (which I recommend doing, if easy).

src/components/common/YourCompany/ChangeBusinessType.vue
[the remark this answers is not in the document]
severinbeauvais: Sometimes, some methods are called externally (eg, for validation/reset).

Still, the accessors don't work correctly in TS so it's not a big deal.

severinbeauvais (commented):
Looks very good 👍 

I'll review your changes then approve if OK.

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: Please confirm with UX designer.

src/utils/resource-utils.ts
severinbeauvais: I guess it is more of a utility (with formal parameters) instead of a state/store query...

OK, resolving.

src/components/common/YourCompany/ChangeBusinessType.vue
seeker25: We've confirmed with UX designer and BA and John Lane

Here's the matrix currently:

seeker25: https://preview.uxpin.com/c21a28a1679da5e2b71044251e8296995e00d9c0#/pages/162402442/simulate/sitemap

seeker25: Straight from Dana:

image

src/components/common/BcRegEntityDetails.vue
seeker25: HelpSection.vue is fairly specific, it probably should have been retooled to support slots

src/components/common/YourCompany/ChangeBusinessType.vue
Jxio: this is what it looks like before:
Pasted Graphic
after I added class="pa-0”: (normal)
Pasted Graphic 1
(narrow):
Pasted Graphic 2

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: Perfect, thanks.

src/components/common/BcRegEntityDetails.vue
severinbeauvais: Probably. I'm suggesting to keep this one but with slots.

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: That's good. Thanks.

src/interfaces/alteration-interfaces/alteration-interface.ts
Jxio: I moved to its own file, and changed to camel case. "SHORT_DESC" is from existing type alteration code.

src/store/store.ts
Jxio: yeah you're right, updated

src/components/common/BcRegEntityDetails.vue
Jxio: I put this section back instead of a component, maybe better not changing HelpSection.vue.

pwei1018:
Temporary Url for review: https://business-edit-dev--pr-516-37q7zf1c.web.app

src/components/common/BcRegEntityDetails.vue
severinbeauvais: I liked the HelpSectionToggle component and recommend you revert it.

My comment above is that you should pass a slot to the component (which is currently your section block), and have all the show/hide logic in the component. Currently that logic is both in BcRegEntityDetails and HelpSectionToggle. You can free BcRegEntityDetails from knowing anything about whether the help is shown/hidden, which is a nice abstraction/ encapsulation/ separation of concern. For example:

In BcRegEntityDetails.vue:
```
<HelpSectionToggle>
  <template #content> <!-- you may be able to use section instead of template here -->
    <!-- your help content goes here -->
  </template>
```

In HelpSectionToggle.vue:
```
<template>
  <div class="help-section-toggle">
    <div
      class="info-text help-toggle"
      @click="isHelpVisible= !isHelpVisible"
    >
      <!-- put your v-icon and show/hide span here -->

      <slot name="content" v-if="isHelpVisible" />
</div>
</template>
```

src/components/Alteration/BcRegEntityDetails.vue
```diff
@@ -0,0 +1,342 @@
...
+}
+.help-toggle {
+  color: $app-blue;
+
+  :hover {
+    cursor: pointer;
+  }
+}
+.provision-help {
+  // padding: 40px;
```
severinbeauvais, line 339: Remove if not used.

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: Well, yes, but I can see that from the code. The real question is what/why are you doing this? For example, "make icon taller so that it matches the text beside it" (or whatever).

src/interfaces/alteration-interfaces/alteration-interface.ts
severinbeauvais: Great, thanks!

src/components/common/BcRegEntityDetails.vue
Jxio: I'm thinking move it to ChangeBusinessType.vue:
```
        <BcRegEntityDetails
          :helpToggle="helpToggle"
          ...
        >
          <template #toggle>
            <div
              class="info-text help-toggle pt-2"
              @click="helpToggle = !helpToggle"
            >
              <v-icon
                class="pr-2 mt-n1"
                color="primary"
              >
                mdi-help-circle-outline
              </v-icon>
              <span v-if="!isHelpVisible">Learn More</span>
              <span v-else>Hide Learn More</span>
            </div>
          </template>
        </BcRegEntityDetails>
```
at BcRegEntityDetails.vue:
```
<!-- Help Section Toggle -->
<slot name="toggle" />
```
feel HelpSectionToggle can be removed cause it's too simple. what do you think?

severinbeauvais: Would you have multiple copies of the above?

I'm looking for simple/clean/easy-to-review code and use of sub-components as needed to accomplish that.

There are different ways to lay this out. Can I leave this with you?

Jxio: no multiple copies, thanks for that! I updated all code under  on BcRegEntityDetails with slots, I think better to keep as its own template cause it's specific to alteration, not for common components.

src/components/common/YourCompany/ChangeBusinessType.vue
Jxio: I removed my comment also removed my change of `:deep`, cause it doesn't need anymore after using slots for "help toggle" section. the `.v-icon` is not added by me, I think it's better to keep as original.

pwei1018:
Temporary Url for review: https://business-edit-dev--pr-516-37q7zf1c.web.app

src/components/common/BcRegEntityDetails.vue
severinbeauvais: You've done things differently than I would have, and that's OK.

My initial concern -- and it is still a concern -- is that the visibility variables are in 2 components (parent and child), and I was trying to get away from that. (In fact, you have both helpToggle and isHelpVisible in ChangeBusinessType.vue.)

I think it would be more simple to just have a help component, which handles its own show/hide, that you pass a slot of help data to.

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: ok

severinbeauvais: PS If you do need to set this CSS, and it's in a child component, I suggest moving this CSS to the child component instead of using deep() here.

The idea here is "try to keep related things together."

Jxio: ok, thanks!

src/components/common/BcRegEntityDetails.vue
Jxio: I've replaced isHelpVisible with helpToggle. Both of these variables were already present in the existing alteration code, and I was simply utilizing them. I didn't realize initially that they were serving the same purpose. Therefore, I've now removed isHelpVisible.

Now, only ChangeBusinessType.vue is handling the show/hide logic, while BcRegEntityDetails.vue is merely retrieving the display status.

severinbeauvais: Did you try this? It seems simpler to me, and only has visibility/display status in 1 component.

> In BcRegEntityDetails.vue:
> 
> ```
> <HelpSectionToggle>
>   <template #content> 
>     
>   </template>
> </HelpSectionToggle>
> ```
> 
> In HelpSectionToggle.vue:
> 
> ```
> <template>
>   <div class="help-section-toggle">
>     <div @click="isHelpVisible= !isHelpVisible">
>       
>   </div>
> 
>   <slot name="content" v-if="isHelpVisible" />
> </template>
> ```

Jxio: I added back HelpSectionToggle.vue with slots. and only let it control visibility.

severinbeauvais: Sorry for the extra work, but it looks great! And it's clean, compact, reusable, etc. Thanks for fixing it.

Jxio: thanks for point out, it's good opportunity to learn vue slots.

severinbeauvais (approved):
Looks good.

pwei1018:
Temporary Url for review: https://business-edit-dev--pr-516-37q7zf1c.web.app

severinbeauvais:
In the future, please wait for a couple of approvals before merging. Thanks!

Jxio:
> In the future, please wait for a couple of approvals before merging. Thanks!

ok will do it next time.

PR #520, opened 2023-07-25
17178 - Alteration - Feature flag - Change Business Types

seeker25:
*Issue #:* /bcgov/entity#17178

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "4.5.1",
+  "version": "4.5.4",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -184,7 +184,7 @@
             <span>Undo</span>
           </v-btn>
           <v-btn
-            v-else-if="isBcCompany || isBcUlcCompany || isBenefitCompany"
+            v-else-if="enableEditButton"
             id="btn-correct-business-type"
             text
             color="primary"
@@ -247,7 +247,7 @@ import { BcRegContacts } from '@/components/common/'
 import { CommonMixin } from '@/mixins/'
 import { CorpTypeCd, GetCorpFullDescription } from '@bcrs-shared-components/corp-type-module/'
 import { ActionBindingIF, EntitySnapshotIF, EntityTypeOption, ResourceIF } from '@/interfaces/'
-import { ResourceUtilities } from '@/utils'
+import { GetFeatureFlag, ResourceUtilities } from '@/utils'
 import { useStore } from '@/store/store'
 
 @Component({
@@ -286,10 +286,12 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   confirmArticles = false
   isEditingType = false
   dropdown: boolean = null
+  supportedEntityTypes = null
 
   /** Called when component is mounted. */
   mounted (): void {
     this.initializeEntityType()
+    this.supportedEntityTypes = GetFeatureFlag('supported-alteration-change-business-types') || []
   }
 
   /** Define the entity type locally once the value has been populated in the store. */
@@ -317,7 +319,18 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
 
   /** Entity type options based on the company type */
   get entityTypeOptions (): EntityTypeOption[] {
-    return this.getResource.changeData?.entityTypeOptions || []
+    const entityTypeOptions = this.getResource.changeData?.entityTypeOptions || []
+    return entityTypeOptions.filter((option: EntityTypeOption) => {
+      return this.supportedEntityTypes?.includes(option.value)
+    })
+  }
+
+  get enableEditButton (): boolean {
+    // Exclude CCC - Originally: isBcCompany || isBcUlcCompany || isBenefitCompany
+    if (this.getEntitySnapshot?.businessInfo?.legalType === CorpTypeCd.BC_CCC) {
+      return false
+    }
+    return this.supportedEntityTypes?.includes(this.getEntitySnapshot?.businessInfo?.legalType)
   }
 
   get minimumThreeDirectorError (): boolean {
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -286,12 +286,12 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   confirmArticles = false
   isEditingType = false
   dropdown: boolean = null
-  supportedEntityTypes = null
+  supportedEntityTypes: Array<string> = []
 
   /** Called when component is mounted. */
   mounted (): void {
     this.initializeEntityType()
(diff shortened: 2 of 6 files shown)


seeker25:
image

Jxio (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "4.5.1",
+  "version": "4.5.4",
```
severinbeauvais, line 3: Why not 4.5.2?

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -286,10 +286,12 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   confirmArticles = false
   isEditingType = false
   dropdown: boolean = null
+  supportedEntityTypes = null
 
   /** Called when component is mounted. */
   mounted (): void {
     this.initializeEntityType()
+    this.supportedEntityTypes = GetFeatureFlag('supported-alteration-change-business-types') || []
```
severinbeauvais, line 294: The feature flags can update in real time, so you might want to fetch them in `entityTypeOptions()` instead.

If this component is only mounted when needed (ie, short lifespan) then this probably doesn't matter, but if this component is always mounted then you should check the FF right before using it.

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -286,10 +286,12 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   confirmArticles = false
   isEditingType = false
   dropdown: boolean = null
+  supportedEntityTypes = null
```
severinbeauvais, line 289: If you initialize this as `[] as Array<string>` then you don't need conditional chaining on line 324 and you'll have typing :)

src/utils/feature-flag-utils.ts
```diff
@@ -14,7 +14,8 @@ const defaultFlagSet: LDFlagSet = {
   'conversion-ui-enabled': false,
   'restoration-ui-enabled': false,
   'sentry-enable': false, // by default, no sentry logs
-  'supported-correction-entities': []
+  'supported-correction-entities': [],
+  'supported-alteration-change-business-types': ['BEN', 'BC', 'CC', 'ULC']
```
severinbeauvais, line 18: Don't merge this. This value must not cause the app to wrong incorrectly if LaunchDarkly can't be reached.

package.json
seeker25: Alot of other PR's 4.5.2, 4.5.3 etc

src/components/common/YourCompany/ChangeBusinessType.vue
seeker25: I didn't want to be slamming it too much, rather do it on mount - this was intentional

package.json
severinbeauvais: You're going to have to rebase/merge anyway...

src/utils/feature-flag-utils.ts
seeker25: I'll set it back to 'BEN', 'BC', that's the default today if that's ok?

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: OK, was just checking that you thought about it.

BTW, getting the feature flag here doesn't cause a network message (which is why it isn't async). The FFs are automatically synced in the background via a stream, I think.

package.json
seeker25: Yup? What's the problem?

src/utils/feature-flag-utils.ts
severinbeauvais: I think this should be `[]`.

You can set the dev LD key in your .env file to get the actual flags (which are OK in dev, right).

seeker25: Yup, the problem was unit tests here though

package.json
severinbeauvais: No problem, but even though you've pre-set the app version here, you're going to need to update/merge this line again anyway.

Resolving this comment.

src/utils/feature-flag-utils.ts
severinbeauvais: Oh, then mock the FF getter.

JazzarKarim (approved):

src/utils/feature-flag-utils.ts
seeker25: I mocked it in setup, there were a few groups of unit tests that relied on this

src/utils/feature-flag-utils.ts
```diff
@@ -7,14 +7,15 @@ declare const window: any
  * Default flag values when LD is not available.
  * Uses "business-edit" project (per LD client id in config).
  */
-const defaultFlagSet: LDFlagSet = {
+export const defaultFlagSet: LDFlagSet = {
```
seeker25, line 10: Exported this, so I could use it in the `setup`

tests/unit/setup.ts
```diff
@@ -28,3 +29,13 @@ Vue.use(TiptapVuetifyPlugin, {
   // optional, default to 'md' (default vuetify icons before v2.0.0)
   iconsGroup: 'mdi'
 })
+
+// Mock feature flags for unit tests.
+jest.spyOn(util, 'GetFeatureFlag').mockImplementation(
+  (name) => {
+    if (name === 'supported-alteration-change-business-types') {
```
seeker25, line 36: If this is too global, let me know, I can copy and paste it into the 3-4 different sections or call a function instead

seeker25:
Built locally: 
image

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -286,12 +286,12 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
   isEditingType = false
   dropdown: boolean = null
-  supportedEntityTypes = null
+  supportedEntityTypes: Array<string> = []
 
   /** Called when component is mounted. */
   mounted (): void {
     this.initializeEntityType()
-    this.supportedEntityTypes = GetFeatureFlag('supported-alteration-change-business-types') || []
+    this.supportedEntityTypes = GetFeatureFlag('supported-alteration-change-business-types')
```
severinbeauvais, line 295: I would still fall back to `[]` in case LD returns null/undefined.

tests/unit/setup.ts
severinbeauvais: This is too global -- other features may want different feature flags.

Nice code, though :)

src/components/common/YourCompany/ChangeBusinessType.vue
seeker25: Hahah I had a feeling - ok adding it to the list. We're going to have to make some more changes for the payment codes I believe

tests/unit/setup.ts
seeker25: adding it to the list for next PR

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: It's the conditional chaining on line 325 that you don't need if this variable is always an array (not falsy).

PR #523, opened 2023-08-10
Add in feature flag check that works with no entries for business typ…

seeker25:
…e changes.

### DIFF
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -376,6 +376,10 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
     if (this.getEntitySnapshot?.businessInfo?.legalType === CorpTypeCd.BC_CCC) {
       return false
     }
+    // If our feature flag is off, don't allow editing of the business type.
+    if (this.supportedEntityTypes == null || this.supportedEntityTypes?.length === 0) {
+      return false
+    }
     return this.supportedEntityTypes?.includes(this.getEntitySnapshot?.businessInfo?.legalType)
   }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-edit-ui",
-  "version": "4.5.6",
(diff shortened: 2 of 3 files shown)


src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -376,6 +376,10 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
     if (this.getEntitySnapshot?.businessInfo?.legalType === CorpTypeCd.BC_CCC) {
       return false
     }
+    // If our feature flag is off, don't allow editing of the business type.
+    if (this.supportedEntityTypes == null || this.supportedEntityTypes?.length === 0) {
+      return false
+    }
```
severinbeauvais, line 382: Should look at new FF instead.

severinbeauvais: I see what you're doing here... sec

severinbeauvais: You only load the flag when the component is mounted. Normally we'd check the flag in real-time (same call).

seeker25: can change it later in main

severinbeauvais: This flag isn't on or off. It's a list of entity types that support changing types.

So this logic isn't quite right. You need to check the current entity type.

seeker25: If it's empty, it wont show, because there are no supported types.

seeker25: It's inclusive:

image

If that's empty, it shouldn't allow any of the types

seeker25: We don't even need this PR - I think we can just turn off the feature flag

severinbeauvais: I think you're right. We should clean up this code though, so we can allow some entity type edits.

PR #588, opened 2024-09-12
16500 - Fixed Restoration People Section Bug

ArwenQin:
*Issue #:* /bcgov/entity#16500

*Description of changes:*
- Conversion to Full Restoration or Limited Restoration Extension page, when we edited the Applicant Person's information, click "Save and Resume", then go back - the "Undo" function doesn't work anymore
- Fixed this bug by removing the re-assign officer ID.
- Fixed the bug that removed and added person are not updated, previously in ticket
 https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/23277
(verify this ticket)
- Fixed the bug when the person is a business/corporation, the file and pay fails (fixed the party schema)

### DIFF
--- a/src/components/common/PeopleAndRoles/PeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/PeopleAndRoles.vue
@@ -730,24 +730,14 @@ export default class PeopleAndRoles extends Mixins(CommonMixin, DateMixin, OrgPe
         if (deleted) delete deleted.actions
       }
     } else {
-      // get ID of original officer to undo
-      const id = this.originalParties[0]?.officer?.id
-
-      let thisPerson
-      if (isNaN(+id)) {
-        // to check assigned UUID
-        thisPerson = cloneDeep(this.originalParties.find(x => x.officer.id === id))
-      } else {
-        // get a copy of original person from original IA
-        thisPerson = cloneDeep(this.originalParties.find(x => +x.officer.id === +id))
-      }
+      // copy the original person
+      let thisPerson = cloneDeep(this.originalParties[0])
 
       // safety check
-      if (!thisPerson) throw new Error(`Failed to find original person with id = ${id}`)
+      if (!thisPerson) throw new Error(`Failed to find original person`)
 
       // splice in the original person
       tempList.splice(index, 1, thisPerson)
-      console.log(tempList)
     }
 
     // set updated list
--- a/src/components/common/PeopleAndRoles/PeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/PeopleAndRoles.vue
@@ -730,8 +730,8 @@ export default class PeopleAndRoles extends Mixins(CommonMixin, DateMixin, OrgPe
         if (deleted) delete deleted.actions
       }
     } else {
-      // get ID of edited person to undo
-      const id = person?.officer?.id
+      // get ID of original officer to undo
+      const id = this.originalParties[0]?.officer?.id
 
       let thisPerson
       if (isNaN(+id)) {
@@ -747,6 +747,7 @@ export default class PeopleAndRoles extends Mixins(CommonMixin, DateMixin, OrgPe
 
       // splice in the original person
       tempList.splice(index, 1, thisPerson)
+      console.log(tempList)
     }
 
     // set updated list
--- a/src/components/common/PeopleAndRoles/PeopleAndRoles.vue
+++ b/src/components/common/PeopleAndRoles/PeopleAndRoles.vue
@@ -730,11 +730,20 @@ export default class PeopleAndRoles extends Mixins(CommonMixin, DateMixin, OrgPe
         if (deleted) delete deleted.actions
       }
     } else {
-      // copy the original person
-      let thisPerson = cloneDeep(this.originalParties[0])
+      // get ID of edited person to undo
+      const id = person?.officer?.id
+
+      let thisPerson
+      if (isNaN(+id)) {
+        // to check assigned UUID
+        thisPerson = cloneDeep(this.originalParties.find(x => x.officer.id === id))
+      } else {
+        // get a copy of original person from original IA
+        thisPerson = cloneDeep(this.originalParties.find(x => +x.officer.id === +id))
+      }
 
       // safety check
-      if (!thisPerson) throw new Error(`Failed to find original person`)
+      if (!thisPerson) throw new Error(`Failed to find original person with id = ${id}`)
 
       // splice in the original person
       tempList.splice(index, 1, thisPerson)
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -389,12 +389,24 @@ export default class FilingTemplateMixin extends DateMixin {
           identifier: this.getBusinessId,
           legalType: this.getEntityType
         },
-        parties: this.getOrgPeople,
+        parties: null, // applied below
         offices: this.getOfficeAddresses,
         contactPoint: this.getContactPoint
       }
     }
+    // Apply parties to filing
+    {
+      // make a copy so we don't change original array
+      let parties = cloneDeep(this.getOrgPeople)
 
+      // prepare parties
+      parties = isDraft ? parties : this.prepareParties(parties)
+
+      // fix schema issues
+      parties = this.fixPartySchemaIssues(parties)
+
+      filing.restoration.parties = parties
+    }
     // Set relationships object for a full restoration only
     if (this.isLimitedRestorationToFull) {
       filing.restoration.relationships = this.getRestoration.relationships
@@ -1506,7 +1518,12 @@ export default class FilingTemplateMixin extends DateMixin {
         if (!party.mailingAddress.streetAddressAdditional) delete party.mailingAddress.streetAddressAdditional
         if (!party.mailingAddress.deliveryInstructions) delete party.mailingAddress.deliveryInstructions
       }
-
+      // If the partyType is organization, remove firstName, middleName and lastName
+      if (party.officer.partyType === PartyTypes.ORGANIZATION) {
+        delete party.officer.firstName
+        delete party.officer.lastName
+        delete party.officer.middleName
+      }
       // NB: at this time, do not convert party from "middleName" to "middleInitial"
 
       return party
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -1518,12 +1518,7 @@ export default class FilingTemplateMixin extends DateMixin {
         if (!party.mailingAddress.streetAddressAdditional) delete party.mailingAddress.streetAddressAdditional
         if (!party.mailingAddress.deliveryInstructions) delete party.mailingAddress.deliveryInstructions
       }
-      // If the partyType is organization, remove firstName, middleName and lastName
-      if (party.officer.partyType === PartyTypes.ORGANIZATION) {
-        delete party.officer.firstName
-        delete party.officer.lastName
-        delete party.officer.middleName
-      }
+
       // NB: at this time, do not convert party from "middleName" to "middleInitial"
 
       return party
--- a/src/views/LimitedRestorationExtension.vue
+++ b/src/views/LimitedRestorationExtension.vue
@@ -112,7 +112,6 @@
 <script lang="ts">
 import { Component, Emit, Mixins, Prop, Watch } from 'vue-property-decorator'
 import { Action, Getter } from 'pinia-class'
-import { v4 as uuidv4 } from 'uuid'
 import { cloneDeep } from 'lodash'
 import { GetFeatureFlag } from '@/utils/'
 import RestorationSummary from '@/components/Restoration/RestorationSummary.vue'
@@ -337,7 +336,6 @@ export default class LimitedRestorationExtension extends Mixins(
 
     // make a copy of the original object
     const copy = cloneDeep(applicant)
-
     return copy
   }
 
--- a/src/views/LimitedRestorationExtension.vue
+++ b/src/views/LimitedRestorationExtension.vue
@@ -335,9 +335,8 @@ export default class LimitedRestorationExtension extends Mixins(
       throw new Error(`Applicant not found for ${this.getBusinessId}`)
     }
 
-    // make a copy of the original object and assign a new id (for UI use only)
+    // make a copy of the original object
     const copy = cloneDeep(applicant)
-    copy.officer.id = uuidv4()
 
     return copy
   }
--- a/src/views/LimitedRestorationToFull.vue
+++ b/src/views/LimitedRestorationToFull.vue
@@ -364,9 +364,8 @@ export default class LimitedRestorationToFull extends Mixins(
       throw new Error(`Applicant not found for ${this.getBusinessId}`)
     }
 
-    // make a copy of the original object and assign a new id (for UI use only)
+    // make a copy of the original object
     const copy = cloneDeep(applicant)
-    copy.officer.id = uuidv4()
 
     return copy
   }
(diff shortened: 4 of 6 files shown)


src/components/common/PeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -730,20 +730,11 @@ export default class PeopleAndRoles extends Mixins(CommonMixin, DateMixin, OrgPe
...
-      let thisPerson
-      if (isNaN(+id)) {
-        // to check assigned UUID
-        thisPerson = cloneDeep(this.originalParties.find(x => x.officer.id === id))
-      } else {
-        // get a copy of original person from original IA
-        thisPerson = cloneDeep(this.originalParties.find(x => +x.officer.id === +id))
-      }
+      // copy the original person
+      let thisPerson = cloneDeep(this.originalParties[0])
```
ArwenQin, line 735: The previous codes won't work because this.originalParties[0]?.officer?.id (the id) actually would change, every time when we save and resume. So we can copy the original person directly and splice it back.

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-588-3ztuvlhp.web.app

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-588-3ztuvlhp.web.app

ArwenQin:
Try: https://business-edit-dev--pr-588-3ztuvlhp.web.app/BC0883552/limitedRestorationExtension?restoration-id=150942&accountid=3040

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: When we save and resume a draft, the JSON should be untouched by the Legal API. Do you have an example of the saved JSON being different from the resumed JSON?

severinbeauvais: I'm not sure values the old code above is comparing, but if I look in a sample filing, I can see the Removed party with its ID that should be unchanged.

image

ArwenQin: I tried with this company:
https://dev.business.bcregistry.gov.bc.ca/BC0883552/?accountid=3040
This is the officer.id when we start a new extension application, with it never being save and resumed:
image
After we save and resume it, it changes:
image
I used console log to see the same changes

ArwenQin: the id only changes when the application is never save and resumed, and after it gets saved and resumed; the later save and resumes won't change it again

ArwenQin: This is where I try console.log: 
image

Open a new application, never save and resumed:
image

After save and resumed:
image

severinbeauvais: Resuming should give you the exact same data you saved. The Legal API does not look at or change anything.

Possibly, above, the id is refreshed by the UI on save? Or the wrong object is being saved?

I can look at it on Monday if you want.

severinbeauvais: When I save your existing Limited Restoration Extension, the data is:

image

On resume, I see the data as:

image

The party officer IDs are the same. If you're seeing something different then something else is going on.

(PS Make sure you look at the payload / preview for the LRE filing (150953), not the "state filing" (150917) that is also fetched.)

severinbeauvais: PS I deleted and re-created the Limited Restoration Extension filing, so it's a new filing id now.

ArwenQin: Thanks Sev. I figured out this is because the LimitedRestorationExtension Vue and LimitedRestorationToFull Vue did a re-assign of officer ID. Please see below.
I removed that re-assign so this bug is fixed.

src/views/LimitedRestorationExtension.vue
```diff
@@ -335,10 +334,8 @@ export default class LimitedRestorationExtension extends Mixins(
       throw new Error(`Applicant not found for ${this.getBusinessId}`)
     }
 
-    // make a copy of the original object and assign a new id (for UI use only)
+    // make a copy of the original object
     const copy = cloneDeep(applicant)
-    copy.officer.id = uuidv4()
-
```
ArwenQin, line 341: Here, the code used to assign a new ID for thee officer. I don't now what is it for, but I removed this line to fix this bug.

src/views/LimitedRestorationToFull.vue
```diff
@@ -364,10 +363,8 @@ export default class LimitedRestorationToFull extends Mixins(
       throw new Error(`Applicant not found for ${this.getBusinessId}`)
     }
 
-    // make a copy of the original object and assign a new id (for UI use only)
+    // make a copy of the original object
     const copy = cloneDeep(applicant)
-    copy.officer.id = uuidv4()
-
     return copy
```
ArwenQin, line 368: Same assigning new ID here, removed

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-588-3ztuvlhp.web.app

ArwenQin:
Try: https://business-edit-dev--pr-588-3ztuvlhp.web.app/BC0883552/limitedRestorationExtension?restoration-id=150995&accountid=3040

src/mixins/filing-template-mixin.ts
```diff
@@ -389,12 +389,24 @@ export default class FilingTemplateMixin extends DateMixin {
...
+    // Apply parties to filing
+    {
+      // make a copy so we don't change original array
+      let parties = cloneDeep(this.getOrgPeople)
+
+      // prepare parties
+      parties = isDraft ? parties : this.prepareParties(parties)
+
+      // fix schema issues
+      parties = this.fixPartySchemaIssues(parties)
```
ArwenQin, line 407: This also fixed the bug that New added person and removed person are not updated - ticket
https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/23277
This ticket can be verified and done now

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-588-3ztuvlhp.web.app

src/mixins/filing-template-mixin.ts
```diff
@@ -1506,7 +1518,12 @@ export default class FilingTemplateMixin extends DateMixin {
...
         if (!party.mailingAddress.deliveryInstructions) delete party.mailingAddress.deliveryInstructions
       }
-
+      // If the partyType is organization, remove firstName, middleName and lastName
+      if (party.officer.partyType === PartyTypes.ORGANIZATION) {
+        delete party.officer.firstName
+        delete party.officer.lastName
+        delete party.officer.middleName
+      }
       // NB: at this time, do not convert party from "middleName" to "middleInitial"
```
ArwenQin, line 1528: I saw this note, not sure whether it matters?
The UI has "middleName", while the schema has "middleInitial".
In addition, in the person schema, it doesn't have taxID and Identifier, while we have them in the filing JSON

ArwenQin:
- Test the undo editing person info
https://business-edit-dev--pr-588-3ztuvlhp.web.app/BC0883552/limitedRestorationExtension?restoration-id=150996&accountid=3040

- Test the remove and add person
previously in ticket
https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/23277
https://business-edit-dev--pr-588-3ztuvlhp.web.app/BC0883554/limitedRestorationExtension?restoration-id=151060&accountid=3040

- Test when the person is a business/corporation, the file and pay goes through now:
https://business-edit-dev--pr-588-3ztuvlhp.web.app/BC1218838/limitedRestorationExtension?restoration-id=151054&accountid=3040

src/components/common/PeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Let's close this conversation and I will open a new conversation in your new code...

src/mixins/filing-template-mixin.ts
severinbeauvais: I believe there is still a mismatch between what the UI calls "middle name" and what the BE calls it. In fact, I think one of the API endpoints returns one property and another endpoint calls it something else. In some places in the UI, we handle this...

That looks like a comment I wrote but I don't remember why. (Note to self: say why when you write something like this.)

I'm guessing this conversion is not needed because the API accepts middleName.

The schema doesn't prevent extra properties.

src/views/LimitedRestorationExtension.vue
severinbeauvais: The reason the ID was initially assigned is for the UI to keep officers separate. I think that removing this probably breaks something. Please test well around adding / changing / deleting / saving / resuming.

ArwenQin: I did more tests around adding / changing / deleting / saving / resuming, all seems good.
However, I found in the filing-template-mixin, this method will remove the parties "id" property after filing
image
I added a comment below too. This method is used by other filing types too.
I did tests, although removed "id" doesn't seem to have impact, do you want to keep the "id" for officer? I can create a new method in the filing-template-mixin for restoration only.
(The process is, open an extension application, loading the original officer, who does have an "id" 
- adding a new officer, which has an id assigned; or don't change the officer at all, just file an extension 
- but after filing, the original/new officer's id will be removed, because we called the above method
- after that, open another new extension application, the officer's "id" is no longer here)

src/mixins/filing-template-mixin.ts
```diff
@@ -389,12 +389,24 @@ export default class FilingTemplateMixin extends DateMixin {
...
         contactPoint: this.getContactPoint
       }
     }
+    // Apply parties to filing
+    {
+      // make a copy so we don't change original array
+      let parties = cloneDeep(this.getOrgPeople)
+
+      // prepare parties
+      parties = isDraft ? parties : this.prepareParties(parties)
```
ArwenQin, line 403: image
Please see line 1544, where it removes the officer.id

src/views/LimitedRestorationExtension.vue
ArwenQin: Alternatively, if we don't want to remove the copy.officer.id = uuidv4(), how about my previous solution
image

src/mixins/filing-template-mixin.ts
severinbeauvais: OK, so this shows us that `id` is only used for UI purposes (when party is added)?

src/mixins/filing-template-mixin.ts
```diff
@@ -389,12 +389,24 @@ export default class FilingTemplateMixin extends DateMixin {
...
+
+      // fix schema issues
+      parties = this.fixPartySchemaIssues(parties)
 
+      filing.restoration.parties = parties
+    }
```
severinbeauvais, line 409: The above code is the same as for Corrections and Change Of Registration filings, so I think it's good 👍

src/mixins/filing-template-mixin.ts
```diff
@@ -1506,7 +1518,12 @@ export default class FilingTemplateMixin extends DateMixin {
         if (!party.mailingAddress.streetAddressAdditional) delete party.mailingAddress.streetAddressAdditional
         if (!party.mailingAddress.deliveryInstructions) delete party.mailingAddress.deliveryInstructions
       }
-
+      // If the partyType is organization, remove firstName, middleName and lastName
+      if (party.officer.partyType === PartyTypes.ORGANIZATION) {
+        delete party.officer.firstName
+        delete party.officer.lastName
+        delete party.officer.middleName
+      }
```
severinbeauvais, line 1526: Why do we need this? I think the BE ignores these properties when the type is organization...?

If this is needed, say why in your comment (line 1521).

src/mixins/filing-template-mixin.ts
severinbeauvais: I think you should just leave this comment here for now.

Please separate the logical code blocks with a blank line (before 1521 and after 1526).

src/mixins/filing-template-mixin.ts
severinbeauvais: Also, please test correction and change filings, since they, too, use this method. (And mention it in Test Notes in your ticket in case Riyaz wants to test it, too.)

src/views/LimitedRestorationExtension.vue
severinbeauvais: I'm wary of your previous solution because, how do we know it's the first (index 0) party you need to undo? What if there are multiple persons in the list?

I think your above solution in filing-template-mixin is OK: only call "prepareParties" if this isn't a draft.

src/mixins/filing-template-mixin.ts
severinbeauvais: If your current solution passes all the tests then you can go and add a comment in the above ticket and resolve it.

src/mixins/filing-template-mixin.ts
ArwenQin: Yes, id is only for UI.

severinbeauvais: OK, I think this comment is resolved.

src/mixins/filing-template-mixin.ts
severinbeauvais: > I saw this note, not sure whether it matters? The UI has "middleName", while the schema has "middleInitial".

I filed a Limited Restoration Extension with a new applicant with a middle name and the output looks correct, so I assume the BE code is working correctly as is. No need to change anything.

image

severinbeauvais: Resolving this conversation.

src/mixins/filing-template-mixin.ts
ArwenQin: Yes agree, this is not needed. I will remove it

src/views/LimitedRestorationExtension.vue
ArwenQin: Is it ok to keep the current solution here?

severinbeauvais: Yes, this change is OK. I'm still not sure about removing the id generation in the other files.

severinbeauvais: I was thinking about your change in filing-template-mixin, which I believe is OK.

I'm still not sure about removing the id generation here. On the one hand, it seems to work, but on the other hand, I think it must have been put here for a reason. Since other filings using this common component, they need to well tested before I am OK with this change.

severinbeauvais: I'm confused again. I was looking at the screenshot above and thinking that was your current change. It's not. So you didn't change any common code.

However, my concern remains: for restorations, you removed the id assignment. Is it the same code in corrections and why does it work there? Maybe corrections needs more testing to find the same bug?

ArwenQin: I searched in the code base, it seems corrections didn't use this code (copy.officer.id = uuidv4()).

severinbeauvais: I think you're right about all of this. Let me test a couple of things...

ArwenQin: Sure, I just updated the filing-template-mixin.ts per your notes below (removed the unused lines)

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-588-3ztuvlhp.web.app

src/mixins/filing-template-mixin.ts
severinbeauvais: Fixed. Resolving.

src/views/LimitedRestorationExtension.vue
severinbeauvais: I've finished misc testing, and it looks good! Resolving this conversation.

severinbeauvais (approved):
Looks good. Thanks for all the extra testing.

After merging this PR, please test + comment + close ticket 23277.

PR #596, opened 2024-11-06
24084 - Alteration Name/Type Change Logic

ArwenQin:
*Issue #:* /bcgov/entity#24084
*Description of changes:*
- Don't show nameRequestRequiredError when the type change is by Changed Name

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-edit-ui",
-  "version": "4.11.3",
+  "version": "4.11.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "business-edit-ui",
-      "version": "4.11.3",
+      "version": "4.11.4",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
         "@bcrs-shared-components/action-chip": "1.1.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "4.11.3",
+  "version": "4.11.4",
   "private": true,
   "appName": "Edit UI",
   "sbcName": "SBC Common Components",
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -388,22 +388,23 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   }
 
   get nameRequestRequiredError (): boolean {
-    if (this.isNumberedCompany) {
+    // Don't show the Error when the type is changed by name
+    if (this.isNumberedCompany || this.isEntityTypeChangedByName) {
       return false
     }
     // Named companies to CC/CCC or ULC/CUL require a name request.
-    if (!this.isEntityTypeChangedByName && (this.isCommunityContribution || this.isUnlimitedLiability)) {
+    if (this.isCommunityContribution || this.isUnlimitedLiability) {
       return true
     }
     // Named ULC to BC Limited (or Benefit Company) requires a name request.
-    if (!this.isEntityTypeChangedByName && ((this.getOriginalLegalType === CorpTypeCd.BC_ULC_COMPANY &&
-      (this.isBcLimited || this.isBenefitCompany)))
+    if (this.getOriginalLegalType === CorpTypeCd.BC_ULC_COMPANY &&
+      (this.isBcLimited || this.isBenefitCompany)
     ) {
       return true
     }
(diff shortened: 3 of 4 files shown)


ArwenQin:
After update:
The nameRequestRequiredError is not showing:
image


We can check the box and click done:
image

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-596-b5vobzl5.web.app

ArwenQin:
Test with: https://dev.business.bcregistry.gov.bc.ca/BC0871787

ArwenQin:
By the current change, I fixed Named BC -> Named ULC, by entering the NR.

However, Named BC -> Numbered ULC, by changing to numbered, will not work. It only works by filing 2 separate changes:
First step, change the Named BC to a numbered BC, file it;
Second step, change the numbered BC to a numbered ULC, and file it.

So a Named BC will change to "xxx BC Ltd" in the first step, just to change a named to a numbered, but can't change its type at the same type.

In summary:
numbered BC -> numbered ULC   works, tested
numbered BC -> named ULC        works, need to enter the NR
named BC -> numbered ULC        needs 2 filings
named BC -> named ULC             works, need to enter the NR


What do you think?

severinbeauvais:
> What do you think?

Can you figure out why it doesn't work, or what's needed to fix it?

I don't like needing 2 filings as that's twice the fees, but named BC → numbered ULC is possibly much less likely than named → named, so maybe it's OK. Or maybe it's OK for now and we can fix it further later.

Please check with Mihai and Olga.

ArwenQin:
The reason is the current functionality only allows change from a named to a numbered by appending "B.C. Ltd." / or "UNLIMITED LIABILITY COMPANY", just same with the current legal type. It's not linking to the type change function. 
Type changes, like a limited company to an ULC, doesn’t automatically trigger the name change to update. 
To resolve this, we need to add logic that links these functions: selecting a change to ULC, then changing from a named company to a numbered one, should trigger the name to update to include "UNLIMITED LIABILITY COMPANY."

I will check with Olga and Mihai

severinbeauvais:
> The reason is the current functionality only allows change from a named to a numbered by appending "B.C. Ltd." / or "UNLIMITED LIABILITY COMPANY", just same with the current legal type. It's not linking to the type change function. Type changes, like a limited company to an ULC, doesn’t automatically trigger the name change to update. To resolve this, we need to add logic that links these functions: selecting a change to ULC, then changing from a named company to a numbered one, should trigger the name to update to include "UNLIMITED LIABILITY COMPANY."

I think you are on the right track: the name change needs to be linked to the type change. Question: should the user change name before type, or type before name, or it shouldn't matter?

ArwenQin:
I am thinking about:
1. 
- If the user changes the name before changing the type, the system updates the name to include "B.C. Ltd."
- When the user then changes the legal type, the name should update accordingly to reflect the new designation, don't show the error message, the user can click "Done".
2.
- If the user tries to change the type before updating the name, show the error message, and they can't click "Done."
- Then the user needs to change the name, it should automatically adjust to "UNLIMITED LIABILITY COMPANY" to match the type change.

severinbeauvais:
> I am thinking about:

How does it work when you use a NR -- can you use a NR first and then change type, or does it have to be the other way around?

I think option 2 makes the most sense. We have to make sure there is a note that says the name (/number) must be changed before they can click Done.

ArwenQin:
When use a NR:
1.
- If the user enters the NR directly (without choosing legal type change), the legal type change is automatically updated, and can't be edited.
2. 
- If the user select legal type change before entering the NR, the error message shows;
- The user can enter the NR now, and after it, the legal type change is automatically updated, and can't be edited.


For NR to numbered, I think both scenario 1 and 2 can happen. For scenario 2, when the error message shows, we can add a note that says the name (/number) must be changed before they can click Done.

severinbeauvais:
Arwen, can you please summarize all possibly scenarios/flows along with what works and what doesn't? It will help me and when explaining to Mihai/Olga.

My goals are: clarity, consistency and correctness :)

ArwenQin:
Olga said
For now let's just fix the problem in the ticket description.
For the scenario you described in the comment, please create a separate ticket and link to the Go-live clean-up epic.

severinbeauvais:
> Olga said For now let's just fix the problem in the ticket description. For the scenario you described in the comment, please create a separate ticket and link to the Go-live clean-up epic.

OK, let's go with this. Please copy-paste your awesome summary from above into the ticket. Thanks!

ArwenQin:
@severinbeauvais shall we merge this ticket now?

ArwenQin:
@severinbeauvais shall we merge this PR now?

severinbeauvais (approved):
LGTM

PR #598, opened 2024-11-13
22639 - Update Certify Section Wording

ArwenQin:
*Issue #:* /bcgov/entity#22639

*Description of changes:*
- This PR is for Edit UI (Create UI and Filing UI are in progress)
- Updated the wording in Certify Section

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@bcrs-shared-components/base-address": "2.0.3",
         "@bcrs-shared-components/breadcrumb": "2.1.5",
         "@bcrs-shared-components/business-lookup": "1.3.4",
-        "@bcrs-shared-components/certify": "2.1.5",
+        "@bcrs-shared-components/certify": "^2.1.55",
         "@bcrs-shared-components/completing-party": "2.1.30",
         "@bcrs-shared-components/confirm-dialog": "1.2.1",
         "@bcrs-shared-components/contact-info": "1.2.5",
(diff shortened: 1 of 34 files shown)


package-lock.json
```diff
@@ -347,12 +347,30 @@
       }
     },
     "node_modules/@bcrs-shared-components/certify": {
-      "version": "2.1.5",
-      "resolved": "https://registry.npmjs.org/@bcrs-shared-components/certify/-/certify-2.1.5.tgz",
-      "integrity": "sha512-+/6MRAHt1xRTcIauIg6lWDII9sLovl0rr2zAsV3+8zvm9+MVujNqC7mlG0CHI2Czl/tqT+CxpAhm/7m6pOk0+g==",
+      "version": "2.1.55",
+      "resolved": "https://registry.npmjs.org/@bcrs-shared-components/certify/-/certify-2.1.55.tgz",
```
ArwenQin, line 351: Requirement: Change ‘is’ to ‘are’
The Certify.vue in bcrs is already updated:
image
However, the updated @bcrs-shared-components/certify version is 2.1.55, while edit UI is depending on 2.1.5
image

package-lock.json
```diff
@@ -347,12 +347,30 @@
...
+    },
+    "node_modules/@bcrs-shared-components/certify/node_modules/@bcrs-shared-components/interfaces": {
+      "version": "1.1.20",
+      "resolved": "https://registry.npmjs.org/@bcrs-shared-components/interfaces/-/interfaces-1.1.20.tgz",
+      "integrity": "sha512-KzxAHYyLANaapMXLuVISFnuNsq2vT56TzK79z38NC3UsAXYalKZDXP5I0zGyKZ8FJIIZi0ZqJLewyHB2ef1dJQ==",
+      "dependencies": {
+        "@bcrs-shared-components/corp-type-module": "^1.0.16",
+        "@bcrs-shared-components/enums": "^1.1.15",
+        "vue": "^2.7.14"
       }
```
ArwenQin, line 374: I ran npm install @bcrs-shared-components/certify, it added those 2 dependencies, is it correct?

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-598-f03y4ktw.web.app

package-lock.json
severinbeauvais: Yes!

This is because the Certify component depends on Vue and Interfaces:
https://github.com/bcgov/bcrs-shared-components/blob/main/src/components/certify/package.json

And Interfaces depends on Corp Type Module and Enums:
https://github.com/bcgov/bcrs-shared-components/blob/main/src/interfaces/package.json

It's listing Vue as "2.7.14 or greater". Hopefully that \resolves to 2.7.16 (same as Edit UI) instead of importing 2 separate versions of Vue.

package.json
```diff
@@ -23,7 +23,7 @@
     "@bcrs-shared-components/base-address": "2.0.3",
     "@bcrs-shared-components/breadcrumb": "2.1.5",
     "@bcrs-shared-components/business-lookup": "1.3.4",
-    "@bcrs-shared-components/certify": "2.1.5",
+    "@bcrs-shared-components/certify": "^2.1.55",
```
severinbeauvais, line 26: Remove the `^` so that we import that exact version. A future version might have some incompatible changes.

package-lock.json
severinbeauvais: There have been a lot  of changes since this shared component was first imported! Please test it well in case there were some incompatible changes.

package.json
ArwenQin: Thanks, done

package-lock.json
ArwenQin: Just tested it, looks good!

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-598-f03y4ktw.web.app

package-lock.json
severinbeauvais: Anyway, I think this is OK.

src/resources/Alteration/BC.ts
```diff
@@ -51,5 +51,6 @@ export const AlterationResourceBc: ResourceIF = {
   },
   certifyClause: 'Note: It is an offence to make a false or misleading statement in respect ' +
     'of a material fact in a record submitted to the Corporate Registry for filing. ' +
-    'See section 427 of the Business Corporations Act.'
+    'See section 427 of the Business Corporations Act.',
+  certifyText: 'business'
```
severinbeauvais, line 55: Does every resource file now provide `certifyText`? If so then please change it to non-optional in the interface (resource-interface.ts).

severinbeauvais: I just checked. SP and GP don't use it, so it needs to remain optional. No change needed. Resolving.

severinbeauvais (approved):
Looks good.

I looked for a better way to do this (instead of updating 32 resource files), such as displaying "business" by default, but that would break CertifySection.vue::readableEntityType(), which needs to handle its current fallback for SPs and GPs. So, let's leave it as-is.

ketaki-deodhar (approved):
I tested a few filings but not all. Looks good in my tests

PR #599, opened 2024-11-19
23604 update logic for changing business name

meawong:
*Issue #:* /bcgov/entity#23604

*Description of changes:*
- update logic for changing business name

### DIFF
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -433,7 +433,7 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   submitTypeChange () {
     this.setEntityType(this.selectedEntityType)
     this.isEditingType = false
-    if (this.shouldUpdateName()) {
+    if (this.isNumberedCompany) {
       const originalName = this.getOriginalLegalName
       const updatedName = this.getUpdatedName(originalName)
       this.setNameRequest({
@@ -445,10 +445,6 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
         this.setNameChangedByType(true)
       }
     }
-  }
-
-  shouldUpdateName (): boolean {
-    return this.isNumberedCompany
   }
 
   getUpdatedName (originalName: string): string {
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -436,19 +436,19 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
     if (this.shouldUpdateName()) {
       const originalName = this.getOriginalLegalName
       const updatedName = this.getUpdatedName(originalName)
+      this.setNameRequest({
+        legalType: this.selectedEntityType,
+        nrNum: this.getOriginalNrNumber
+      } as any)
+      this.setNameRequestLegalName(updatedName)
       if (originalName !== updatedName) {
-        this.setNameRequest({
-          legalType: this.selectedEntityType,
-          nrNum: this.getOriginalNrNumber
-        } as any)
-        this.setNameRequestLegalName(updatedName)
         this.setNameChangedByType(true)
       }
     }
   }
 
   shouldUpdateName (): boolean {
-    return this.isNumberedCompany && !this.hasBusinessNameChanged
+    return this.isNumberedCompany
   }
 
   getUpdatedName (originalName: string): string {
--- a/src/components/common/YourCompany/ChangeBusinessType.vue
+++ b/src/components/common/YourCompany/ChangeBusinessType.vue
@@ -306,6 +306,7 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   @Getter(useStore) getEditedLabel!: string
   @Getter(useStore) getEntityType!: CorpTypeCd
   @Getter(useStore) getNameRequestLegalName!: string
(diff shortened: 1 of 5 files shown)


bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-599-h6gsnubn.web.app

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-599-h6gsnubn.web.app

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -433,24 +433,20 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   submitTypeChange () {
     this.setEntityType(this.selectedEntityType)
     this.isEditingType = false
-    if (this.shouldUpdateName()) {
```
meawong, line 436: We would only update the name and set the NR if the business name had **NOT** changed. This prevents the case when a user attempts to change the business type (and therefore the name) more than once, as the business name **HAS** changed.

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -433,24 +433,20 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
...
-        } as any)
-        this.setNameRequestLegalName(updatedName)
         this.setNameChangedByType(true)
       }
     }
   }
 
-  shouldUpdateName (): boolean {
-    return this.isNumberedCompany && !this.hasBusinessNameChanged
-  }
```
meawong, line 452:
```
/** Whether business name has changed. */
    hasBusinessNameChanged (): boolean {
      const currentLegalName = this.getNameRequestLegalName // may be empty
      const originalLegalName = this.getOriginalLegalName // from original business record
      return (currentLegalName !== originalLegalName)
}
    
```

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-599-h6gsnubn.web.app

src/views/Alteration.vue
```diff
@@ -214,6 +214,7 @@ export default class Alteration extends Mixins(CommonMixin, FeeMixin, FilingTemp
     }
   }
 
+  @Watch('hasBusinessNameChanged')
```
severinbeauvais, line 217: I don't know if this change is the reason why, but I found that I can Undo the Company Name and the Business Type remains changed.

image

Should this make the filing invalid?

Should the Business Type also be undone?

severinbeauvais (commented):
Good analysis and I agree with the solution, but please test a bit more (including saving and resuming) to make sure all edge cases are OK, too. Thanks!

src/components/common/YourCompany/ChangeBusinessType.vue
severinbeauvais: Does this logic work on a named company? With and without a new NR?

src/views/Alteration.vue
severinbeauvais: Oh, wait.

When I saved and resumed, the resume was in the old Edit UI. Let me test this again.

severinbeauvais (approved):

src/views/Alteration.vue
meawong: I just checked this on dev and this issue already exists but I can look into fixing this. Should I update the scope on this ticket?

severinbeauvais: Please test this thoroughly. I believe it's working well.

ArwenQin:
I did some tests, seems good!

ArwenQin (approved):

src/components/common/YourCompany/ChangeBusinessType.vue
meawong: Since these changes still retain the check for a numbered business, as before, named companies should not be reaching this block of code below. There should be separate logic that handles named businesses (I can verify where this is if you like).

meawong: > I did a little test on a named BC Limited, and could not change the type to ULC without a NR.

I thought this was the expected behaviour for named businesses? Since altering the business type for a named business would require a name change and this requires an NR.

severinbeauvais: Yes, but what if wanted to change my named BC to a numbered ULC? Anyway, we discussed this previously and decided it was low priority right now.

src/components/common/YourCompany/ChangeBusinessType.vue
```diff
@@ -433,24 +433,20 @@ export default class ChangeBusinessType extends Mixins(CommonMixin) {
   submitTypeChange () {
     this.setEntityType(this.selectedEntityType)
     this.isEditingType = false
-    if (this.shouldUpdateName()) {
+    if (this.isNumberedCompany) {
       const originalName = this.getOriginalLegalName
       const updatedName = this.getUpdatedName(originalName)
+      this.setNameRequest({
+        legalType: this.selectedEntityType,
+        nrNum: this.getOriginalNrNumber
```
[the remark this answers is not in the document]
meawong: The original NR is an empty string for a numbered company.

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-599-h6gsnubn.web.app

severinbeauvais (approved):
I did some quick tests and it LGTM!

Please list out, in the ticket, the tests you did. Great job! Thanks!

PR #627, opened 2025-04-23
26693 invalid shares prompt

TVWerdal:
Issue #: /bcgov/entity/issues/26693

Update so that name changes are no longer regarded as share structure changes.


bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-627-faiqkemb.web.app

src/store/store.ts
```diff
@@ -1109,11 +1109,30 @@ export const useStore = defineStore('store', {
...
 
-      return !IsSame(originalShareClasses, currentShareClasses)
+      // If share structures exist, check for changes.
+      // We only want to know if share structure has changed - this does not include name changes.
+      // Name changes also adds a record for an Edit action, which causes a mismatch
+      const omittedValues = ['name', 'action']
+      let change = false
+      for (const [k] of Object.entries(currentShareClasses)) {
+        // If we have currentShareClasses but no OriginalShareClasses, this is net-new.
+        if ((originalShareClasses === undefined || isEmpty(originalShareClasses))) {
```
JazzarKarim, line 1122: You can probably do `!originalShareClasses`.

src/store/store.ts
```diff
@@ -1109,11 +1109,30 @@ export const useStore = defineStore('store', {
...
+          return true
+        }
+        change = !IsSame(
```
JazzarKarim, line 1125: I'm thinking you can also make things much cleaner by getting rid of this variable. It'll be something like:
```typescript
        if (!IsSame(
          currentShareClasses[k as keyof ShareClassIF],
          originalShareClasses[k as keyof ShareClassIF],
          omittedValues
        )) {
          return true
        }
return false
```

src/store/store.ts
severinbeauvais: I see an extra set of parentheses in there, too.

src/store/store.ts
severinbeauvais: You could also try
```
return !IsSame( ... )
```

severinbeauvais:
Can you please list all the tests you've done, here and/or in the ticket?

src/store/store.ts
TVWerdal: Oh you are right @severinbeauvais  - I originally had more checks there that were unnecessary and I removed them. Surprised the linter didn't pick those up.

src/store/store.ts
TVWerdal: Originally the check was done as 'return !IsSame( ... )'  - but the problem is that the omit doesn't seem to work when checking the overall objects of currentShareClasses and originalShareClasses so you have to walk through all share items inside those classes to get a proper check. So if there was only one share class you could return off of the !IsSame, but if there are more you would not check them. So for instance, say the first was false, you would miss the second which was true. does that make sense?

severinbeauvais: OK, I see what you're doing: keep processing until you find a change.

Would like 1135 always return False then?

severinbeauvais: If you used a lambda then you could probably suimply this even more :)

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-627-faiqkemb.web.app

TVWerdal:
> Can you please list all the tests you've done, here and/or in the ticket?

My tests included checking if the message about required resolutions showed in the following circumstances  - 
image

Single share ex - 
[Before changes - dev](https://dev.edit.business.bcregistry.gov.bc.ca/C9900895/alteration?accountid=3040&businessid=C9900895) 
[With changes - gc](https://business-edit-dev--pr-627-faiqkemb.web.app/C9900895/alteration?accountid=3040&businessid=C9900895)

 - Change name of share on Dev (fails):  EXPECTED: Does not show  RESULT: shows
 - Change name of share on local (passes): EXPECTED: Does not show RESULT: does not show
 - Change element of share structure on Dev (passes): EXPECTED: shows, RESULT: shows
 - Change element of share structure on Local (passes): EXPECTED: shows, RESULT: shows

Multiple Shares ex - 
[Before changes - dev](https://dev.edit.business.bcregistry.gov.bc.ca/BC0887813/alteration?accountid=3040&businessid=BC0887813)
[With changes - gc](https://business-edit-dev--pr-627-faiqkemb.web.app/BC0887813/alteration?accountid=3040&businessid=BC0887813)

 - Change name of share on Dev (fails):  EXPECTED: Does not show  RESULT: shows
 - Change name of share on local (passes): EXPECTED: Does not show RESULT: does not show
 - Change element of share structure on Dev (passes): EXPECTED: shows, RESULT: shows
 - Change element of share structure on Local (passes): EXPECTED: shows, RESULT: shows

src/store/store.ts
JazzarKarim: Looks good, thanks. Resolving.

src/store/store.ts
JazzarKarim: Thanks, LGTM. @severinbeauvais You can resolve this conversation if you're satisfied with the change Sev.

JazzarKarim (approved):

severinbeauvais:
> Multiple Shares ex
> 
> * Change name of share on Dev (fails):  EXPECTED: Does not show  RESULT: shows
> * Change name of share on local (passes): EXPECTED: Does not show RESULT: does not show
> * Change element of share structure on Dev (passes): EXPECTED: shows, RESULT: shows
> * Change element of share structure on Local (passes): EXPECTED: shows, RESULT: shows

If I change the name of a series then I need to add a resolution. Is this correct?

If I check "this share class has special..." then I don't need to add a resolution. Is this correct?

If I add a series then I don't need to add a resolution. Is this correct?

TVWerdal:
> If I change the name of a series then I need to add a resolution. Is this correct?

> If I check "this share class has special..." then I don't need to add a resolution. Is this correct?

> If I add a series then I don't need to add a resolution. Is this correct?

> Update: Oops, I didn't notice that a resolution date was automatically added (no red text).

From the ticket, my understanding was the opposite. Changing a name should not count as a change to share structure. Changing other information (.ie the share value, max etc) should require a resolution/court order.

severinbeauvais:
> > If I change the name of a series then I need to add a resolution. Is this correct?
> 
> From the ticket, my understanding was the opposite. Changing a name should not count as a change to share structure. Changing other information (.ie the share value, max etc) should require a resolution/court order.

Should changing the name of a series require a resolution? In preview URL, it does.

TVWerdal:
Looks like those nested shares (series) don't get checked with this code. Need some recursion or something here I suppose. That and sorting.

severinbeauvais:
Sorting? I don't think so. I believe users can "move up" or "move down" items manually.

PS Maybe nested for loops instead of recursion? 😆

TVWerdal:
I mean sort the arrays so we catch any users actions that move their shares. This makes sure we are checking the proper records against each other. Otherwise any move at all causes the message to pop up.

Ex. If the original record is ordered A,B,C - moving down item A makes it B,A,C. So this will not match the original object anymore when we look for a change. One problem is that what we are getting prior to compare seems to get flipped between Arrays and Array-like objects. So that needs to be accounted for as well.

severinbeauvais:
> I mean sort the arrays so we catch any users actions that move their shares. This makes sure we are checking the proper records against each other. Otherwise any move at all causes the message to pop up.

Makes sense!

TVWerdal:
Got some clarification in the ticket [here](https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/26693) I was incorrect with my original understanding. The issue was that you should only see the highlighted red text and be required to add date if both of the following are true: 

1. Special Rights or Restrictions flag is on any one share in the list and... 
2. There have been changes to a Share

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-627-dsg5r3pg.web.app

src/store/store.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable padded-blocks */
```
severinbeauvais, line 1: Do you still need this? If so, can you disable just a single line, or perhaps move this before the block and then re-enable it after the block?

src/utils/RemoveNullProps.ts
```diff
@@ -0,0 +1,15 @@
+/******************
+ * Null action properties can be assigned to the ShareClasses when cancelling edits
+ * This is fail safe to ensure null actions are not included in the comparison
```
severinbeauvais, line 3: Good comment but I think it belongs in the store code where you use it.

Here, this is a generic util, which doesn't know about share classes. Use a generic description instead.

src/utils/OrderShares.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  // Sort the top-level shareClasses by ID
+  const sorted = [...shareClasses].sort((a, b) => Number(a.id) - Number(b.id))
+
+  // Recursive call if we have series shares
+  sorted.forEach(share => {
+    if (share.series) {
+      share.series = OrderShares(share.series)
+    }
+  })
```
severinbeauvais, line 30: I'm generally OK with recursion.

How does OrderShares know when to stop recursing?

severinbeauvais: Also, did you consider a nested function in hasShareStructureChanged() ? Might be cleaner.

TVWerdal: It will stop recursing (1) if it does not find any share series or (2) if it is checking a series because they can't have their own series. Originally I had this and "RemoveNullProps" nested in hasShareStrucutreChanged, but there was a lot going on in there so I pulled these two out. Do you think it is better to have everything in the hasShareStructureChanged?

severinbeauvais: My basic rules of thumb:
- try to keep complexity down
- try to keep similar things together
- write code that makes sense to others

I can't decide for this code, so maybe just leave it for now until there's a compelling reason to do it differently.

severinbeauvais: > How does OrderShares know when to stop recursing?

I would maybe renaming shareClasses -> shareOrSeries. What do you think?

severinbeauvais (approved):
Approving for when you're happy with it.

Please ensure you call out in the ticket what to verify (for Riyaz).

src/utils/OrderShares.ts
TVWerdal: > My basic rules of thumb:
> 
> * try to keep complexity down
> * try to keep similar things together
> * write code that makes sense to others
> 
> I can't decide for this code, so maybe just leave it for now until there's a compelling reason to do it differently.

These are good rules and I was trying to do something similar. I went back and forth on pulling them out because it would make more sense if they were re-used elsewhere - but I doubt that will happen.

JazzarKarim (approved):
Looks good. Comments in the code are great.

PR #642, opened 2025-07-21
25302 Document Record Service Integration

flutistar:
*Issue #:* /bcgov/entity#25302

*Description of changes:*
- Replaced Minio with Document Record Service
- DocumentID component


src/components/SpecialResolution/UploadRulesOrMemorandum.vue
```diff
@@ -59,12 +63,16 @@ import { LegalServices } from '@/services/'
 })
 export default class UploadRulesOrMemorandum extends Vue {
   @Getter(useStore) getUserInfo!: any
+  @Getter(useStore) getBusinessId!: string
 
   @Prop({ default: false })
   readonly invalidSection: boolean
+  @Prop({ default: null })
```
severinbeauvais, line 70: Would it make sense to have a default of False instead?

src/components/common/FileUploadPdf.vue
```diff
@@ -33,6 +36,13 @@ export default class FileUploadPdf extends Vue {
...
+  /** Set `required` to true for the three variables below once the feature flag is removed:
+   * `documentClass`, `documentType`, `businessIdentifier`
+  */
+  @Prop({ default: null }) readonly documentClass!: string
+  @Prop({ default: null }) readonly documentType!: string
+  @Prop({ default: null }) readonly businessIdentifier!: string
+
   // Network service functions
   @Prop({ required: true })
   readonly getPresignedUrl!: (fileName: string) => Promise<PresignedUrlIF>
```
severinbeauvais, line 48: FYI, I'll be moving the network functions to Legal Services module. Shouldn't make a difference.

src/components/common/FileUploadPdf.vue
```diff
@@ -246,16 +261,31 @@ export default class FileUploadPdf extends Vue {
...
+        psu = await this.getPresignedUrl(file.name)
 
+        // NB: will throw on API error
+        res = await this.uploadToUrl(psu.preSignedUrl, file, psu.key, this.userId)
+      }
       // check if successful
-      if (res?.status === 200) {
-        return psu.key
+      if (res && [StatusCodes.OK, StatusCodes.CREATED].includes(res.status)) {
+        return this.enableDocumentRecords ? res.data.documentServiceId : psu.key
```
severinbeauvais, line 288: OK, cool, I like this integration of DRS and Minio.

I'm wondering how we're going to migrate / cut-over from one to the other, because this implementation doesn't support both at the same time. Thoughts?

src/mixins/filing-template-mixin.ts
```diff
@@ -689,7 +695,8 @@ export default class FilingTemplateMixin extends DateMixin {
         name: FilingTypes.CONVERSION,
         certifiedBy: this.getCertifyState.certifiedBy,
         date: this.getCurrentDate, // "absolute day" (YYYY-MM-DD in Pacific time)
-        folioNumber: '' // not applicable to SP/GP
+        folioNumber: '', // not applicable to SP/GP
+        documentIdState: this.getDocumentIdState // document id will serve as a barcode number
```
severinbeauvais, line 699: This (and the other above) is for the "build" (save) functions. Do you need to add something to the "parse" (load) functions for resuming a draft?

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-642-1p4yabrm.web.app

src/components/common/FileUploadPdf.vue
flutistar: - The document will be uploaded to either Minio or DRS, depending on the feature flag.
- For downloading documents, a regex is used to determine the source:
     ```DRS_ID_PATTERN = /^DS\d{10}$/```
  - Example Minio file key: `adb1e8a0-5dd5-4f71-8726-6845ee3185ff.pdf`
  - Example DRS service ID: `DS0000101295`

severinbeauvais: Thanks for explaining upload vs download. Now I get it :)

package.json
severinbeauvais, line 1: Remember to update the app version before merging.

src/mixins/filing-template-mixin.ts
severinbeauvais: Wait. I could 6 places where you save document id state but only 5 places where you restore it.

src/components/common/FileUploadPdf.vue
```diff
@@ -33,6 +36,13 @@ export default class FileUploadPdf extends Vue {
   @Prop({ default: null }) readonly pageSize!: PageSizes
   @Prop({ required: true }) readonly userId!: string
 
+  /** Set `required` to true for the three variables below once the feature flag is removed:
```
severinbeauvais, line 39: Why set required to True later? Maybe update comment, please.

src/components/common/FileUploadPdf.vue
```diff
@@ -41,6 +51,7 @@ export default class FileUploadPdf extends Vue {
 
   pdfjsLib: any
   /** Component key, used to force it to re-render. */
+  enableDocumentRecords = GetFeatureFlag('enable-document-records')
```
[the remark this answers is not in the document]
severinbeauvais: Closing. No change needed. This works fine.

src/views/SpecialResolution.vue
```diff
@@ -105,7 +105,7 @@
           <StaffPayment
```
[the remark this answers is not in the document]
flutistar: For a special resolution filing, I believe there are uploaded documents such as rules and memorandum files.
Since we create a DRS record in real time when documents are uploaded via the UI, we might not need to create another DRS record before submission.

severinbeauvais:
Please fix the GHA errors in the diffs.

package.json
severinbeauvais: Poke.

src/mixins/filing-template-mixin.ts
```diff
@@ -945,8 +945,8 @@ export default class FilingTemplateMixin extends DateMixin {
 
     // store document service state
     this.setDocumentIdState({
-      valid: filing.header.documentIdState.valid,
-      consumerDocumentId: filing.header.documentIdState.consumerDocumentId
+      valid: filing.header.documentIdState?.valid,
```
severinbeauvais, line 948: Do you need this to fix the unit tests, but it works fine when deployed?

If so, a better solution might be to update the unit tests so the subject object is declared.

(I am bit wary of optional chaining because it can mask errors. I'm not saying that's the case here but I try to evaluate all uses of it. Let me know what you think.)

tests/unit/LimitedRestorationExtension.spec.ts
```diff
@@ -28,22 +28,27 @@ const store = useStore() // eslint-disable-line @typescript-eslint/no-unused-var
...
   }
 }
 const stateFiling = {
   header: {
     name: 'restoration',
-    status: 'COMPLETED'
+    status: 'COMPLETED',
+    documentIdState: {
+      consumerDocumentId: "DS12345678",
+      valid: true
```
severinbeauvais, line 50: Small GHA complaint above.

severinbeauvais: There are actually 3 lint issues.

severinbeauvais (commented):
I think you're almost ready to merge. Please update the app version. Given the significant update this is, I suggest updating the minor number (ie, 4.14.0).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "4.13.3",
+  "version": "4.14.0",
```
severinbeauvais, line 3: Please run `npm i` and commit the lock file as well.

PR #651, opened 2025-09-16
30532-Initial update to add officer compoenent.

TVWerdal:
*Issue #:* /bcgov/entity/30532
*Issue #:* /bcgov/entity/30533

Update to add in officer component. 

*NOTE* Also includes changes from 30533 as both were completed in tandem.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-edit-ui",
-  "version": "4.13.10",
+  "version": "4.13.11",
```
severinbeauvais, line 3: You can update the minor version if you like.

src/enums/roleTypes.ts
```diff
@@ -1,9 +1,21 @@
...
+  CFO = 'Cfo',
+  PRESIDENT = 'President',
+  VICE_PRESIDENT = 'Vice President',
+  CHAIR = 'Chair',
+  TREASURER = 'Treasurer',
+  SECRETARY = 'Secretary',
+  ASSISTANT_SECRETARY = 'Assistant Secretary',
+  OTHER = 'Other',
   SUBSCRIBER = 'Subscriber'
 }
```
severinbeauvais, line 21: If this list gets used in other UIs, let's consider moving this to shared enums (or common layer for Nuxt UIs).

src/views/Alteration.vue
```diff
@@ -220,6 +227,11 @@ export default class Alteration extends Mixins(CommonMixin, FeeMixin, FilingTemp
     }
   }
 
+  /** For LD FF */
+  get showOfficers (): boolean {
+    return GetFeatureFlag('supported-change-of-officers-entities')
```
severinbeauvais, line 232: This FF doesn't return True or False. It returns a space-delimited list of entity types for which the feature is enabled. See other examples for how to use this.

src/components/common/PeopleAndRoles/CurrentOfficers.vue
```diff
@@ -0,0 +1,314 @@
...
+                    medium
+                    class="add-officer-icon"
+                    color="appDkBlue"
+                    v-text="officers.length > 0 ? 'mdi-pencil-outline' : 'mdi-account-plus-outline'"
+                  />
+                  <span
+                    v-if="officers.length > 0"
+                    class="officer-change-text"
+                  >
+                    Manage Officers
```
severinbeauvais, line 36: For some reason, this button doesn't clearly show as active when I mouse over it. Compare with the other Change buttons.

severinbeauvais: PS I'm actually looking at the enabled button (not the disabled one in this code).

src/components/common/PeopleAndRoles/CurrentOfficers.vue
```diff
@@ -0,0 +1,314 @@
...
+          v-else
+          class="no-officers-header"
+        >
+          There are currently no officers. You can add officers if you are required to do so.
+        </thead>
+        <tbody v-if="officers.length > 0">
+          <!-- List Content -->
+          <tr
+            v-for="(orgPerson, index) in officers"
+            :key="`officer.officer.id:${index}`"
```
severinbeauvais, line 103: This is probably overkill. I think you could use either the officer id or the index. Thoughts?

src/components/common/PeopleAndRoles/CurrentOfficers.vue
```diff
@@ -0,0 +1,314 @@
...
+                    class="officers-name"
+                    v-on="on"
+                  >{{ formatFullName(orgPerson.officer) }}</span>
+                </template>
+                <span class="officers-name">{{ formatFullName(orgPerson.officer) }}</span>
+              </v-tooltip>
+            </td>
+            <!-- Roles -->
+            <td class="px-0">
+              <span v-if="orgPerson.roles.length > 0">
```
severinbeauvais, line 125: You might not need this check. I think v-for doesn't iterate at all if array length is 0.

src/components/common/PeopleAndRoles/CurrentOfficers.vue
```diff
@@ -0,0 +1,314 @@
...
+                <v-col
+                  v-for="(role, key) in orgPerson.roles"
+                  :key="key"
+                  class="roles-detail"
+                >
+                  <span class="roles-detail">{{ role.roleType }}</span>
+                </v-col>
+              </span>
+            </td>
+            <!-- Delivery Address -->
```
severinbeauvais, line 135: For directors, we show Mailing Address first then Delivery Address. Please check with UX if this should be made consistent.

src/components/common/PeopleAndRoles/CurrentOfficers.vue
```diff
@@ -0,0 +1,314 @@
...
+    </v-card>
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Mixins, Prop } from 'vue-property-decorator'
+import { Getter } from 'pinia-class'
+import { OrgPersonIF } from '@/interfaces/'
+import { RoleClass, RoleTypes } from '@/enums/'
+import { CommonMixin, OrgPersonMixin } from '@/mixins/'
```
severinbeauvais, line 166: What do you need OrgPersonMixin for? (CurrentDirectors doesn't use it.)

src/components/common/PeopleAndRoles/CurrentOfficers.vue
```diff
@@ -0,0 +1,314 @@
...
+
+  /** Gather all officers with appropriate roles */
+  get officers (): OrgPersonIF[] {
+    return this.getOrgPeople
+      .filter(person =>
+        person.roles?.some(role =>
+          role?.roleClass === RoleClass.OFFICER
+        )
+      )
+      .filter(person => !this.wasRemoved(person))
```
severinbeauvais, line 199: You're saying the list of persons that the API provides includes removed items?

src/views/Alteration.vue
severinbeauvais, line 232: Are you going to add a check for an existing COO draft?

TVWerdal: Ah, I looked at a few in this repo that treated it like a true/false. I had assumed that would be the case here. Will revise.

bcregistry-sre:
Temporary Url for review: https://business-edit-dev--pr-651-s0ymm71n.web.app

src/enums/roleTypes.ts
severinbeauvais: Nothing to do at the moment. Resolving.

severinbeauvais (approved):
I have only the comment about Mailing Address and Delivery Address order, otherwise LGTM! (Please get more reviews.)

_Update: Jacqueline said to leave the address order as it matches Officer Change UI._

src/components/common/PeopleAndRoles/CurrentOfficers.vue
TVWerdal: From Jacqueline: The officers layout follows what shows on the officers filing screen to be consistent with that, unfortunately it is inconsistent with directors. I have a task to try and solve this problem in the future but for now it will be like this

severinbeauvais: OK.

